PR #66, opened 2026-06-12
feat(index): compile and apply real .gitignore in walker + watcher

### LINKED ISSUE #62: Watcher + walker: compile and apply real .gitignore rules (not a hardcoded skip list)
Discovery (`index/walker.rs::should_skip_name`) and the watcher use a **hardcoded** directory skip list (`.git`, `.rag-rat`, `target`, `node_modules`, `dist`, `build`, `coverage`). This misses repo-specific `.gitignore` entries — generated dirs, build outputs, vendored code in non-standard locations — so they can get indexed and (recursively) watched.

Compile the repo's `.gitignore` (plus nested gitignores) — e.g. via the `ignore` crate (ripgrep's matcher) — and apply it in **both**:
- the walker (file discovery for indexing), and
- the watcher (`event_is_relevant` path classification + which dirs to register with notify).

Keep the existing hardcoded skips as a fallback for non-git trees.

Context: this is the clean version of "monitor only what we should." It is **separate** from the read-event re-index loop already fixed in 36a2e1b (that loop was caused by firing on file *reads*, not by scope).

skakri:
Replaces the hardcoded skip-name list with real `.gitignore` compilation, applied consistently by both the file walker (discovery) and the watcher (`event_is_relevant`).

### Approach
- New shared `crates/rag-rat-core/src/index/ignore_rules.rs::IgnoreMatcher`, compiled once per index pass / once at watcher start and consulted by **both** walker and watcher, so a path one skips the other also skips (no drift).
- **Augment, not replace**, the hardcoded list: it survives as an unconditional floor (`FLOOR_DIRS`: `.git .rag-rat .omx .omc node_modules target dist build coverage`) that can't be whitelisted away — guarantees correct skips in a non-git tree and that `.rag-rat/` is never indexed (feedback-loop guard). Compiled `.gitignore` rules layer on top.
- Added the `ignore` crate (ripgrep's engine; MIT/Unlicense, license-gate friendly).

### Two non-obvious findings (recorded as a repo memory)
1. A single flat `GitignoreBuilder` flattens nested files' globs to the root and `**/`-prefixes them, so a nested rule would wrongly ignore repo-wide. Fix: a per-directory stack of `Gitignore` matchers, deepest decision winning (nested `!whitelist` overrides an outer ignore).
2. `Gitignore::matched` only tests the leaf — a file under an ignored *directory* slipped through. Fix: `matched_path_or_any_parents`.

### Tests
New tests for root + nested `.gitignore` scoping, whitelist negation, floor-without-gitignore, walker end-to-end, and watcher `event_is_relevant` (skips ignored paths, fires on real source mutations, never on reads). The existing watcher target-directory constraint is preserved. 179 tests pass, clippy clean.

Fixes #62

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1870,6 +1870,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "globset"
+version = "0.4.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "52dfc19153a48bde0cbd630453615c8151bce3a5adfac7a0aebfbf0a1e1f57e3"
+dependencies = [
+ "aho-corasick",
+ "bstr",
+ "log",
+ "regex-automata",
+ "regex-syntax",
+]
+
+[[package]]
 name = "half"
 version = "2.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2284,6 +2297,22 @@ checksum = "cb68373c0d6620ef8105e855e7745e18b0d00d3bdb07fb532e434244cdb9a714"
 dependencies = [
  "icu_normalizer",
  "icu_properties",
+]
+
+[[package]]
+name = "ignore"
+version = "0.4.26"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b915661dd01db3f05050265b2477bcc6527b3792388e2749b41623cc592be67d"
+dependencies = [
+ "crossbeam-deque",
+ "globset",
+ "log",
+ "memchr",
+ "regex-automata",
+ "same-file",
+ "walkdir",
+ "winapi-util",
 ]
 
 [[package]]
@@ -3091,6 +3120,7 @@ dependencies = [
  "fastembed",
  "gix",
  "iai-callgrind",
+ "ignore",
  "libc",
  "model2vec-rs",
  "notify",
--- a/crates/rag-rat-core/Cargo.toml
+++ b/crates/rag-rat-core/Cargo.toml
@@ -15,6 +15,7 @@ categories.workspace = true
 anyhow.workspace = true
 fastembed = { workspace = true, optional = true }
 gix.workspace = true
+ignore = "0.4.26"
 model2vec-rs = { version = "0.2.1", default-features = false, features = ["hf-hub", "fancy-regex"], optional = true }
 notify = "8.2.0"
 rayon.workspace = true
--- a/crates/rag-rat-core/src/config.rs
+++ b/crates/rag-rat-core/src/config.rs
@@ -166,6 +166,23 @@ impl FromStr for TargetKind {
 }
 
 impl Config {
+    /// The deduplicated set of target directories (relative to [`Config::root`]) across all
+    /// targets, in stable order. Used to scope `.gitignore` nested-discovery to the indexed trees
+    /// (see [`crate::index::ignore_rules::IgnoreMatcher::compile`]) instead of recursing the whole
+    /// root into unindexed siblings.
+    pub fn target_directories(&self) -> Vec<PathBuf> {
+        let mut seen = BTreeSet::new();
+        let mut dirs = Vec::new();
+        for target in &self.targets {
+            for dir in &target.directories {
+                if seen.insert(dir.clone()) {
+                    dirs.push(dir.clone());
+                }
+            }
+        }
+        dirs
+    }
+
     pub fn load(path: impl AsRef<Path>) -> Result<Self, ConfigError> {
         let path = path.as_ref();
         let text = fs::read_to_string(path)?;
--- a/crates/rag-rat-core/src/index/git_history.rs
+++ b/crates/rag-rat-core/src/index/git_history.rs
@@ -625,6 +625,15 @@ fn count_table(conn: &Connection, table: &str) -> anyhow::Result<u64> {
     Ok(u64::try_from(count).unwrap_or(0))
 }
 
+/// The enclosing Git worktree root for `root` (the directory `git rev-parse --show-toplevel`
+/// reports), or `None` when `root` is not inside a Git worktree (or `git` is unavailable). This is
+/// the single place the codebase shells `--show-toplevel`; reuse it rather than adding a parallel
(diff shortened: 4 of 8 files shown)


skakri:
All four P2 review findings addressed in 8fdc9df.

**1 (correctness) — floor check matched ancestors outside the repo root.** `IgnoreMatcher` now stores the repo `root` and `is_ignored` strips it *first*, running the floor-component and gitignore checks only on the path **relative to the repo root**. A repo living under a floor-named ancestor (`/tmp/build/repo`, a checkout below `…/node_modules/…`, `…/target/…`) no longer has the whole repo classified as ignored. A path outside the root is simply not governed (returns not-ignored). Test: `repo_root_under_floor_named_ancestor_still_indexes`.

**2 (correctness) — nested `.gitignore` could un-ignore files under an already-ignored parent.** Git parent-exclusion is now honored two ways: (a) `is_ignored` walks the relative path root→leaf and short-circuits at the first ignored ancestor *directory* (a deeper `!negation` can't resurrect a file under an excluded dir); (b) `collect_gitignores` no longer descends into — or reads a nested `.gitignore` in — a directory an already-collected outer rule ignores. So root `gen/` + `gen/.gitignore` `!keep.rs` keeps `gen/keep.rs` ignored. Tests: `nested_negation_under_ignored_parent_stays_ignored`, `flat_negation_unignores_non_floor_file`, `negation_unignores_whitelisted_dir_subtree`.

**3 (staleness) — matcher compiled once at watcher start.** `watcher_main`'s matcher is now recompiled whenever a `.gitignore` mutation event is observed, so the running watcher classifies subsequent events against current rules. (Each pass's discover walk already compiles its own fresh matcher, so the index itself was current; this fixes the *event-classification* staleness.)

**4 (staleness) — a `.gitignore` edit scheduled no pass.** `event_is_relevant` now treats a mutation to any `**/.gitignore` as relevant (via `is_gitignore_path`), so a rules edit fires a pass — which recompiles (finding 3) and re-discovers (drop newly-ignored, add newly-unignored) — even with `periodic_sweep_secs = 0`. Test: `gitignore_edit_is_relevant_and_recompile_reflects_new_rules`.

**Design note:** kept the shared per-directory `ScopedGitignore` stack rather than switching the walker to `ignore::WalkBuilder`. The walker and watcher share one matcher by design (no drift), and the watcher fundamentally needs a single-path predicate for deleted-file events, which `WalkBuilder` doesn't expose. The two bugs came from *how* the stack matched, not the stack itself; `is_ignored` now implements the same git ancestor-walk semantics `WalkBuilder` uses internally, made correct by construction.

Files: `crates/rag-rat-core/src/index/ignore_rules.rs`, `crates/rag-rat-core/src/watch.rs`. `cargo +nightly fmt`, `cargo clippy --all-targets -- -D warnings` clean, `cargo test -p rag-rat-core` green (183 passed).

skakri:
@codex review

skakri:
Round-3 redesign pushed as `0c3a753`: the `.gitignore` matcher is rebuilt on the `ignore` crate's native gitignore machinery (an ancestor-anchored stack of per-directory `ignore::gitignore::Gitignore` matchers), so the recurring git-semantics edge cases are now correct **by construction** rather than patched.

**Round-3 findings**

1. **`watch.rs` — root `.gitignore` not watched.** The watcher subscribed only to the target dirs (`crates`, `docs`), which sit *below* the worktree/`config.root`, so an edit to a `.gitignore` above them was never delivered → no recompile. Fixed: `watch.rs::gitignore_watch_dirs(root)` returns the ancestor chain from the Git worktree root down to `config.root`, and `watcher_main` now subscribes to each of those dirs **non-recursively** (in addition to the recursive target watches). A `.gitignore` mutation above the target dirs is delivered, fires a pass, and recompiles the shared matcher — including the `periodic_sweep_secs = 0` case. Covered by a real-`notify`-watcher delivery test plus a recompile/classification test for the subdirectory-`config.root` case.

2. **`ignore_rules.rs` — discovery recursed the whole config root.** Nested-`.gitignore` discovery is now scoped to the configured `target.directories` (`Config::target_directories()` → `compile(root, target_dirs)`); `collect_nested_gitignores` walks only `root.join(dir)` for each target, never a full-root sweep into unindexed siblings. Floor dirs and outer-ignored dirs are still pruned mid-scan. A test asserts a nested `.gitignore` in an out-of-target sibling is never compiled (no effect), while an in-target one still is.

3. **`mod.rs` — ancestor gitignores missed for subdirectory roots.** When `[index].root` is a subdir of a larger worktree, the worktree-root `.gitignore` rules Git applies under it were unseen. Fixed: the matcher resolves the worktree root via the new `git_history::worktree_root` (extracted from the existing `git_repo`, the single `git rev-parse --show-toplevel` call — no parallel git pattern) and seeds the stack with the ancestor `.gitignore` chain from the worktree root down to `config.root` before the in-tree ones. A `git init`-backed test confirms a worktree-root glob/dir rule governs files under a subdirectory `config.root`.

**Earlier rounds, now correct by construction**

- **Floor is repo-relative, not absolute-ancestor.** The floor check runs on the `config.root`-relative path, and the gitignore frame (`base`) is accepted only when `root.starts_with(base)`. A repo under `/tmp/build/repo` still indexes (test retained).
- **Parent-exclusion.** `is_ignored` walks ancestors root→leaf via `Gitignore::matched` (leaf-only — the caller does the ancestor walk, so no double `matched_path_or_any_parents`) and short-circuits at the first ignored *directory*; discovery won't descend into an outer-ignored dir to read a nested `.gitignore`. `gen/` + `gen/.gitignore !keep.rs` → `gen/keep.rs` stays ignored (test retained).
- **`.gitignore` edits fire + recompile.** Preserved: `event_is_relevant` short-circuits on `is_gitignore_path`, and `watcher_main` recompiles on a gitignore mutation.

**Design notes.** Walker and watcher still share **one** `IgnoreMatcher` predicate (`is_ignored(path, is_dir)`) — the watcher needs a single-path decision for delete events, so no `WalkBuilder`. `.rag-rat/` remains an unconditional floor entry (feedback-loop guard). `cargo +nightly fmt` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test -p rag-rat-core` green (189 tests).

skakri:
@codex review

Issue #61, opened 2026-06-11
SCIP oracle epic — live LSP hardening and additional backends remaining


skakri:
## Tracking — decomposed into sub-issues

This issue is now the **epic**; the design below is the reference. Slices, in dependency order:

- [x] #67 — Prerequisite: persist callee-identifier byte range on edges (additive, byte-identical; unblocks everything)
- [x] #68 — Phase 1: SCIP protobuf reader + heuristic precision/recall eval *(depends #67)*
- [x] #69 — Phase 2: `oracle run` for Rust — `Compiler` tier, `resolved-external`, `compare_graph_to_scip` *(depends #68)*
- [x] #70 — Phase 3: moniker anchors for memories + logical symbols *(depends #69)*
- [x] #71 — Phase 4: TypeScript + C/C++ (scip-clang) backends *(depends #69)*
- [x] #72 — Phase 5: Kotlin (scip-java / SemanticDB) backend *(depends #69)*
- [x] #73 — Benchmark: `resolved-external` / `oracle-upgradeable` taxonomy + heuristic precision *(depends #69)*
- [ ] #74 — Phase 6: live LSP oracle for the incremental path *(depends #69)*

**Enabled-by (separate epic, not part of #61):** #75 — protocol-aware cross-repo / cross-service edges *(depends on #70 monikers)*. The phase-1 eval bar (#68) now gates on **precision + recall against multi-language fixtures**.

---

## Summary

Add an optional, reconcile-style pass that consumes SCIP indexes (`rust-analyzer scip`, `scip-typescript`, `scip-clang`, `scip-java`/Kotlin) and uses them as a resolution oracle:

1. Upgrade heuristic edge resolutions to a new `Compiler` confidence tier, with provenance.
2. Split the unresolved bucket into "unresolvable" vs. "resolves to a named external dependency."
3. Feed `eval` a ground-truth precision metric for the tree-sitter resolver.
4. Add SCIP monikers as a semantic anchor binding for repo memories and logical symbols.

The pass is batch, diff-based, resumable, and opt-in — same architectural slot as `reconcile` (embeddings) and lazy blame. Indexing stays fast, network-free, and dependency-free; the only new compiled dependency is a SCIP protobuf reader (`prost`). Indexer binaries are external prerequisites, consistent with the size-budget policy.

## Motivation

[...]

PR #76, opened 2026-06-12
Persist callee-identifier byte range on graph edges (#67)

### LINKED ISSUE #67: SCIP phase-1 prerequisite: persist callee-identifier byte range on edges
Foundation for #61 (SCIP oracle). The only part that touches the existing edge pipeline; everything else in #61 is greenfield and depends on this.

## Problem
SCIP occurrences key on the **callee identifier token's** position, but for a `calls_name` edge the source span is `span_for_node(node)` where `node` is the entire `call_expression` (`edges/extract.rs::symbol_edge_with_context`) — the whole `foo(a, b)` / `obj.method(x)`, not the identifier. Edge rows store `source_start_line`/`source_end_line` + byte range but **no callee-identifier position**. The extractor already locates the identifier node (`call_target_name` / `identifiers_under(child_by_field_name("function"))`) — it just discards the position.

## Scope
- Add **additive nullable** edge columns `callee_start_byte` / `callee_end_byte` (byte range only; `(line, col)` in the document's `position_encoding` is derived at join time from checkout bytes, so no column needed).
- Capture the identifier node's byte range in the extractor instead of discarding it.
- Thread through `EdgeCandidate` → `CompactEdge`/`CompactSpan` (the memory-tuned full-rebuild accumulator — note the per-edge footprint cost) → both insert paths (rebuild + incremental).
- Migration via the existing `add_column_if_missing` pattern (cf. `apply_edge_source_target_spans`).

## Invariants
- MUST stay **additive and byte-identical** — the rebuild bench + the id-independent content-fingerprint tests guard this.
- Nullable / default so existing rows and non-call edges are unaffected.

Part of #61.

skakri:
Fixes #67. Byte-identical-guarded prerequisite for the SCIP oracle epic (#61).

For a `calls_name` (and `references_type` / `constructs` / `uses_macro` / `implements`) edge the source span is `span_for_node(node)` over the whole `call_expression`, but SCIP occurrences key on the **callee identifier token's** position. This records that token's byte range on each symbol-referencing edge so the SCIP-oracle join can match on it. `(line, col)` in the document's `position_encoding` is derived at join time from checkout bytes, so only the byte range is stored.

## Design → change map

1. **`EdgeCandidate.callee_span: Option<CalleeRange { start_byte, end_byte }>`** (`edges/mod.rs`) — set ONLY for symbol-referencing edges built via `symbol_edge` / `symbol_edge_with_context` (new trailing `callee_span` param). `None` for `contains` / `imports` / `exports` / file-level (`file_edge`) edges and for the impl-header string-split `implements` / `references_type` in `rust_impl_edges` (no clean identifier node).

2. **Extractor threading** (`edges/extract.rs` + `edges/helpers.rs`) — added node-returning twins of the existing name helpers: `call_target_node`, `first_identifier_node`, `last_identifier_node`, `identifier_nodes_under`, and `final_segment_node`. Each call site passes `node.map(CalleeRange::of_node)`. Covers **all** edge-building languages: `rust_edges`, `typescript_edges`, `kotlin_edges`, `c_like_edges`, `rust_impl_edges`. Multi-segment paths point at the **final** identifier segment (`obj.method` → `method`, `a::b::c` → `c`) via `final_segment_node` narrowing a `scoped_identifier`/`scoped_type_identifier` to its `name` field — matching how SCIP keys the occurrence. The scoped-receiver `references_type` (`Foo::bar()`) points at the **first** segment (`Foo`), since that's the type referenced.

3. **`CompactEdge`** (`edges/mod.rs`) — stores the range as **two bare `u32` fields with `u32::MAX` as the none-sentinel** (`CALLEE_NONE`), NOT an `Option`, mirroring the `OptSym` niche. `pack_callee` clamps + sentinels on intern; `callee_byte_columns()` hydrates back to `(NULL, NULL)` for the insert. **Footprint delta: +8 bytes/edge (~64 → ~72)** — at the kernel's ~11M candidates, ~88 MB of additional accumulator held only until the single resolve-and-insert pass. Safe because graph-parsed files are bounded by `MAX_GRAPH_PARSE_BYTES` = 512_000, so offsets fit `u32` and `u32::MAX` is a free sentinel.

4. **Schema** — additive **nullable** `callee_start_byte` / `callee_end_byte` INTEGER columns via the `add_column_if_missing` pattern. New migration `apply_edge_callee_byte_range`, registered as **V017** in `mod.rs` (`LATEST_SCHEMA_VERSION` 16 → 17) + `known_version` / `known_migration` / `migration_checksum_mismatch`. `baseline.rs` carries the columns for fresh DBs. The pre-squash convention (additive migrations on top of baseline) is preserved — no baseline edited in place.

5. **Insert paths** — both the full-rebuild insert (`resolve.rs::resolve_and_insert_edges`, via `CompactEdge::callee_byte_columns()`) and the incremental insert (`helpers.rs::insert_candidates`, from `candidate.callee_span`) write the two columns, NULL when the range is absent. All `INSERT INTO edges` use explicit column lists, so the added columns are positionally safe.

## Byte-identical verification

The columns are additive + nullable, so existing edge fields and their content fingerprints do not move. Confirmed by the full `cargo test -p rag-rat-core` suite passing unchanged (194 tests), including `server_derived_call_path_hash_is_stable_and_validates_through_edge_churn` — the cp1 sorted edge-fingerprint hash over the existing fields. Non-call / file-level edges keep NULL.

## Tests added (`schema_bootstrap_tests.rs`)

- `calls_name_edge_stores_callee_identifier_byte_range_not_whole_call` — Rust `foo(a, b)` → `text[start..end] == "foo"` (and asserts it does NOT span the whole call); `obj.method(x)` → `method`; `a::b::c()` → `c`; a `contains` edge has NULL callee range.
- `imports_edge_has_null_callee_byte_range` — file-level `imports` edge has NULL.
- `c_calls_name_edge_stores_callee_identifier_byte_range` — non-Rust coverage: C `helper(runtime)` → `helper`.
- Extended `migrate_adds_edge_name_columns_before_indexing_them` to assert the two new columns are added to a legacy edges table.

`cargo +nightly fmt`, `cargo clippy --all-targets -- -D warnings`, and `cargo test -p rag-rat-core` all clean/green.

### DIFF
--- a/crates/rag-rat-core/src/index/edges/extract.rs
+++ b/crates/rag-rat-core/src/index/edges/extract.rs
@@ -74,6 +74,7 @@ pub(crate) fn contains_edges(symbols: &[IndexedSymbol]) -> Vec<EdgeCandidate> {
                 evidence: Some(child.qualified_name.clone()),
                 receiver_hint: None,
                 source_span: child.span(),
+                callee_span: None,
                 edge_kind: EdgeKind::Contains,
                 confidence: EdgeConfidence::Exact,
             });
@@ -191,6 +192,7 @@ pub(crate) fn rust_edges(
                         target_qualified_name: target_qualified_name(node, text),
                         receiver_hint: scoped_receiver_name(node, text),
                     },
+                    call_target_node(node).map(CalleeRange::of_node),
                 ));
             }
             // A scoped call receiver is a type reference only when it names a type. By Rust
@@ -207,6 +209,12 @@ pub(crate) fn rust_edges(
                     receiver,
                     EdgeKind::ReferencesType,
                     EdgeConfidence::NameOnly,
+                    // The type is the receiver — the LEADING `::` segment (`Foo` in `Foo::bar()`)
+                    // — so anchor the range on the function path's first
+                    // identifier, not its tail.
+                    node.child_by_field_name("function")
+                        .and_then(first_identifier_node)
+                        .map(CalleeRange::of_node),
                 ));
             }
         },
@@ -220,6 +228,7 @@ pub(crate) fn rust_edges(
                     EdgeKind::UsesMacro,
                     EdgeConfidence::NameOnly,
                     EdgeContext::default(),
+                    first_identifier_node(node).map(CalleeRange::of_node),
                 ));
             },
         "impl_item" => rust_impl_edges(text, node, symbols, out),
@@ -231,6 +240,7 @@ pub(crate) fn rust_edges(
                     name,
                     EdgeKind::ReferencesType,
                     EdgeConfidence::NameOnly,
+                    last_identifier_node(node).map(final_segment_node).map(CalleeRange::of_node),
                 ));
             }
         },
@@ -262,6 +272,9 @@ pub(crate) fn rust_impl_edges(
             evidence: Some(edge_evidence(node, text)),
             receiver_hint: None,
             source_span: span_for_node(node),
+            // The trait/type names here come from string-splitting the impl header, not from a
+            // located identifier node, so there is no clean callee range to record (#67).
+            callee_span: None,
             edge_kind: EdgeKind::Implements,
             confidence: EdgeConfidence::NameOnly,
         });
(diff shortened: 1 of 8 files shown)


skakri:
@codex review

Issue #69, opened 2026-06-12
SCIP phase 2: `oracle run` for Rust (Compiler tier, resolved-external, compare_graph_to_scip)


skakri:
Depends on the SCIP reader + eval pass.

## Scope
- `oracle` manifest (tool id, version detection, invocation, prerequisites); `rag-rat oracle run [--tool rust-analyzer] [--scip <path>]` — invoke `rust-analyzer scip` or consume a pre-built `.scip`. Missing tool → `Blocked` + install hint, no error exit (same UX as a missing embedding model).
- Content-addressed staleness: key by `(file_sha, tool, tool_version)`. Unchanged files carry upgrades across commits with no re-run; a changed file's edges revert to heuristic display, status `oracle-stale`. `gc` prunes `edge_oracle`/`oracle_runs` with the per-commit rows.
- New `Compiler` confidence tier (above `Exact`) surfaced when a current oracle row exists; `resolution_reason = "scip:<tool>@<version>"`; `resolved-external(<package> <version>)` from the SCIP symbol's package component.
- `trace_callees`/`find_callers`/`impact_surface` prefer `Compiler` when current; `completeness_risk` becomes quantitative.
- New MCP tool `compare_graph_to_scip` (sibling of `compare_graph_to_text`): where tree-sitter and the compiler disagree.
- Worktree: oracle rows bind to commit-scoped index rows only; dirty/overlay files always show heuristic (never `Compiler`).

## Acceptance
- `rag-rat oracle run --tool rust-analyzer` on this repo: upgrades > 0, contradictions logged (never auto-resolved), report printed; re-run with no changes is a no-op.
- `cargo build` + rebuild wall time + byte-identical output unaffected.

Part of #61.

PR #81, opened 2026-06-12
feat(oracle): SCIP phase 1 — reader, occurrence→edge join, heuristic precision/recall eval

### LINKED ISSUE #68: SCIP phase 1: protobuf reader + heuristic precision/recall eval
Depends on the callee-identifier byte-range columns (#67).

## Scope
- SCIP protobuf reader — decide: `prost`/prost-build codegen from `scip.proto`, or the Apache-2.0 `scip` crate. Do NOT hand-copy proto definitions.
- New subsystem `crates/rag-rat-core/src/index/oracle/` mirroring the embeddings layout (mod/manifest/scip/join/run/status).
- Additive side tables: `oracle_runs(id, tool, tool_version, commit_sha, started_at, status, stats_json)` and `edge_oracle(edge_id, file_sha, tool, tool_version, resolved_symbol_id, scip_symbol, kind, computed_at)`. The heuristic resolution on the edge row is NEVER overwritten — both coexist so eval can diff.
- Occurrence→edge join: read `Document.position_encoding` per document (don't assume UTF-8/UTF-16); skip `local N`; require identifier-token containment (not line equality) using the callee byte range (#67) → (line,col) at join time; def inside corpus → map by def-range/symbol overlap; outside → `resolved-external(<package>)`.
- Phase 1 consumes a pre-built `.scip` (no indexer invocation yet), Rust first.

## Eval acceptance bar
- `eval --json` emits, against fixtures + a corpus: **precision** (oracle-confirmed fraction of `Exact`/`Syntactic`), **recall** (oracle edges the heuristic missed), NameOnly/Ambiguous recovery, oracle-upgradeable fraction of unresolved.
- Fixtures are **multi-language** (not Rust-only) and include **non-ASCII identifiers + multibyte comments** (position-encoding traps). Baseline-gated via `eval --update-baseline`.
- No MCP-visible product change yet.

Part of #61.

skakri:
Phase 1 of the SCIP oracle (#61): consume a compiler-produced SCIP index as a resolution oracle and turn it into a trustworthy heuristic-vs-compiler precision/recall signal. Greenfield except for the additive schema; no change to indexing behavior, query output, or the byte-identical rebuild.

## What it adds
- **`index/oracle/` subsystem** (mirrors `index/ai/`): `scip.rs` (protobuf reader), `join.rs` (occurrence→edge join), `run.rs` (the pass + metrics), `store.rs` (side-table I/O), `status.rs`, `mod.rs`. Consumes a **pre-built `.scip`** (Rust first); no indexer invocation — `oracle run` (CLI), the `Compiler` confidence tier, and the `compare_graph_to_scip` tool are #69.
- **Dependency:** the `scip` crate (Apache-2.0) + `protobuf` for message (de)serialization. No hand-copied proto.
- **Schema V018 (additive):** `oracle_runs` + `edge_oracle` side tables (STRICT). The heuristic `edges` row is **never overwritten** — the oracle verdict lives only in `edge_oracle`, so eval can diff the two; `edge_oracle` cascades off `edges` (`ON DELETE CASCADE`); staleness keyed by `(file_sha, tool, tool_version)`.
- **Eval metrics** in `eval --json` (emitted only when a `.scip` is supplied; baseline-gated): precision (oracle-confirmed fraction of Exact/Syntactic), call **recall**, NameOnly/Ambiguous recovery rate, and oracle-upgradeable fraction.

## Correctness properties (the point of phase 1 — the numbers must not lie)
- **Checkout-scoped by construction.** Every metric/status read of `edge_oracle` routes through one shared `EDGE_ORACLE_SCOPE_JOIN` predicate filtered to the run's `(commit_sha, worktree_id)`, matching the scoped candidate selection. A source-scan test fails CI if any query reads `edge_oracle` outside `store.rs`.
- **Numerator ⊆ denominator, bounded ≤ 1.0** for every rate, all over the same scoped population.
- **Call recall over the call population on both sides.** Covered and gap are both the `calls_name` occurrence set (occurrence-deduped): type-ref/macro confirmations don't inflate it, and `Term` field/const reads don't deflate it.
- **Contradiction vs. external.** An Exact/Syntactic edge with an in-corpus target that SCIP resolves externally is recorded as a **contradiction** (honestly lowers precision), not `resolved-external`.
- **Content-integrity gate.** Per document, disk bytes are SHA-checked against the candidate's `file_sha`; drifted candidates are skipped (no verdict) and surfaced as `skipped_drifted`, so eval warns instead of silently misreporting.
- **Real containment** for def→symbol mapping; per-document `position_encoding` handling (UTF-8/16/32, surrogate pairs, multi-line ranges, malformed-range rejection).
- **Authoritative reruns** — per `(tool, tool_version, checkout)`: clear the scope, then write, in one transaction.

## Tests / quality
~50 oracle tests, including the non-ASCII identifier under both UTF-8 and UTF-16, multi-worktree metric isolation, the content-drift skip, and the call-population recall cases. clippy `-D warnings` + fmt clean; full `rag-rat-core` suite green.

Fixes #68. Part of #61.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1150,7 +1150,7 @@ dependencies = [
  "gix-worktree-stream",
  "nonempty",
  "smallvec",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1177,7 +1177,7 @@ dependencies = [
  "gix-trace",
  "kstring",
  "smallvec",
- "thiserror",
+ "thiserror 2.0.18",
  "unicode-bom",
 ]
 
@@ -1240,7 +1240,7 @@ dependencies = [
  "gix-ref",
  "gix-sec",
  "smallvec",
- "thiserror",
+ "thiserror 2.0.18",
  "unicode-bom",
 ]
 
@@ -1254,7 +1254,7 @@ dependencies = [
  "bstr",
  "gix-path",
  "libc",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1290,7 +1290,7 @@ dependencies = [
  "gix-trace",
  "gix-traverse",
  "gix-worktree",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1310,7 +1310,7 @@ dependencies = [
  "gix-trace",
  "gix-utils",
  "gix-worktree",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1325,7 +1325,7 @@ dependencies = [
  "gix-path",
  "gix-ref",
  "gix-sec",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1353,7 +1353,7 @@ dependencies = [
  "once_cell",
  "parking_lot",
  "prodash",
- "thiserror",
+ "thiserror 2.0.18",
  "walkdir",
  "zlib-rs",
 ]
@@ -1376,7 +1376,7 @@ dependencies = [
  "gix-trace",
  "gix-utils",
  "smallvec",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1390,7 +1390,7 @@ dependencies = [
  "gix-features",
  "gix-path",
  "gix-utils",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1414,7 +1414,7 @@ dependencies = [
  "faster-hex",
  "gix-features",
  "sha1-checked",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1476,7 +1476,7 @@ dependencies = [
  "memmap2",
  "rustix",
  "smallvec",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1487,7 +1487,7 @@ checksum = "65c9dedd9e90b0d47624d2ed241d394e09294118364e87b9b7e5f1fe755f3c2c"
 dependencies = [
  "gix-tempfile",
  "gix-utils",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1506,7 +1506,7 @@ dependencies = [
  "gix-validate",
  "itoa",
  "smallvec",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1527,7 +1527,7 @@ dependencies = [
  "memmap2",
  "parking_lot",
  "tempfile",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1546,7 +1546,7 @@ dependencies = [
  "gix-path",
  "memmap2",
  "smallvec",
- "thiserror",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


skakri:
## Review fixes — all six findings addressed (head `919fdd7`)

Each finding mapped to its fix, with tests. `cargo +nightly fmt` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test -p rag-rat-core` green (248 lib tests, +7 new oracle tests).

| # | Sev | Fix |
|---|-----|-----|
| 1 | HIGH | **Recall now measures one population.** Covered side restricted to `calls_name` occurrences (the run tracks distinct `covered_call_occurrences` → `OracleReport.covered_calls`, occurrence-deduped); gap side tightened via `symbol_is_callable` requiring `).` (Method), so bare-`.` `Term` reads — struct fields, consts, statics, enum variants — no longer count as missed calls. `eval_metrics` takes `RecallCalls { covered, oracle_only }` (both from the run) and computes `recall = covered / (covered + oracle_only)`. Per the sub-point, both sides are occurrence-deduped so a duplicate edge over one occurrence can't skew recall. |
| 2 | MED | **Content integrity.** Per document, the disk bytes read for position conversion are hashed and compared to the candidate's `file_sha`; mismatched (drifted) candidates are skipped with no verdict and tallied in `OracleReport.skipped_drifted`, surfaced as `EvalReport.oracle_skipped_drifted` + an `eval` WARNING line. |
| 3 | LOW/MED | **`oracle_runs.worktree_id`** added directly in V018 (unshipped), written by `record_oracle_run`, and `last_run_meta` filters on `(commit_sha, worktree_id)` — a sibling checkout's run no longer surfaces as this checkout's last run. |
| 4 | LOW | **`classify_edge` returns the matched occurrence range**; the run marks exactly that occurrence covered (the weaker start-only/first-match `mark_covered` is gone), so covered-marking and the verdict can't disagree on overlaps. |
| 5 | LOW | **`indexed_paths_in_scope` excludes `kind='deleted'`** tombstones, so a deleted-but-still-in-`.scip` file's occurrences don't inflate the gap. |
| 6 | LOW | (a) Source-scan test asserts `FROM edge_oracle` appears only in `store.rs`, so a future unscoped query fails CI. (b) `name_only_recovery_rate` numerator scoped to `upgrade` verdicts on `NameOnly`/`Ambiguous` edges — an `Exact`-with-NULL writer can't recreate an over-1.0 rate. |

**What the recall number now means:** CALL recall — the fraction of in-corpus *calls* the oracle saw that a `calls_name` edge covered. Both numerator and denominator are occurrence-counted over the same call population (covered = distinct `calls_name`-edge occurrences; gap = uncovered callable occurrences whose def maps in-corpus). Confirmations on `references_type`/`uses_macro`/`implements`/`constructs` no longer inflate it, and field/const `Term` reads no longer deflate it. Precision is unchanged.

**New tests (7):** `recall_gap_excludes_field_const_term_reads`, `covered_side_ignores_references_type_confirmation`, `drifted_file_sha_is_skipped_not_verdicted`, `last_run_meta_is_scoped_to_active_worktree`, `deleted_file_occurrences_do_not_inflate_gap`, `raw_edge_oracle_queries_live_only_in_store`, `name_only_recovery_rate_excludes_exact_upgrades` — plus the existing metric tests updated to the `RecallCalls` surface.

**Coverage (`cargo llvm-cov -p rag-rat-core --lib --no-default-features`, region %):** before → after — `run.rs` 0→93.6, `join.rs` 0→96.4, `scip.rs` 0→95.6, `store.rs` 0→84.8, `status.rs` 0→92.7, `mod.rs` 0→100 (the "before" 0% is the pre-fix snapshot; the residual `store.rs` misses are pre-existing `?`-error arms, not the new branches, which are all exercised).

The Risk memory Fable recorded (`mem_19ebbfdab09_3e56a6077c06`) is marked obsolete/resolved; new Invariant memories capture the call-recall, content-integrity, and worktree-scope rules.

PR #125, opened 2026-06-14
fix(memory): validate path/dir bindings to non-indexed files against the filesystem (#98)

### LINKED ISSUE #98: memory doctor false-flags path bindings to non-indexed files as 'gone'
## Symptom
`rag-rat memory doctor` reports a path-bound memory as `gone` ("code appears deleted; … mark-obsolete") even though the file exists on disk. Concrete case: a memory anchored to `tools/bench.Containerfile` was flagged gone and recommended for obsoletion, but the file is present and the note is still valid. Acting on the recommendation would **delete correct guidance**.

## Root cause
`validate_path_binding` (crates/rag-rat-core/src/query/memory/validate.rs:338-347) resolves the path against the **indexed** `files` table:
```sql
SELECT sha256 FROM files WHERE path = ?1 ORDER BY id DESC LIMIT 1
```
No row → `gone`. But `files` only contains files in the configured indexed language set (here rust + markdown — see `rag-rat.toml`; `index_status.file_count_by_language` confirms only those). Any memory bound to a path outside that set — Containerfiles, shell scripts, `.yml` workflows, `.toml` configs — has **no `files` row by construction**, so it's indistinguishable from a genuinely deleted file. `validate_dir_binding` has the analogous issue (a dir with only non-indexed files looks empty via `dir_has_files`).

This is a false `gone`: the binding target is alive, it's just not an indexed *source* file. It also explains a standing `memory doctor` noise floor (these never clear) and makes the AGENTS.md "re-anchor anything flagged gone after a refactor" guidance dangerous for such anchors.

## Suggested fix
For a path/dir binding whose target isn't in `files`, fall back to a **filesystem existence check** against the index `source_root` before declaring `gone`:
- path binding: if `<source_root>/<path>` exists on disk → `current` (it's an area anchor to a real-but-unindexed file), else `gone`.
- dir binding: if the dir exists on disk → `current`, else `gone`.

A bare path binding is already treated as an area anchor ("current while the file is indexed, never content-stale", validate.rs:348-356) — this just extends "indexed" to "present in the repo" for non-indexed file types. Spanned `path:start-end` bindings to non-indexed files can stay `unverified` (no chunk to hash) rather than `gone`.

## Repro
1. `memory_create` with `bind: { path: "tools/bench.Containerfile" }` (or any non-.rs/.md path).
2. `rag-rat memory doctor` → reports it `gone` though the file exists.

Severity: P2 — risks silent loss of valid memories if the mark-obsolete recommendation is followed; also a persistent false-positive in doctor output.

skakri:
## Problem

`memory doctor` false-flags path/dir-bound memories as `gone` when the target is a **non-indexed file type** (Containerfile, shell script, `.yml` workflow, `.toml` config) — even though the file is present in the repo.

Root cause: `validate_path_binding` / `validate_dir_binding` resolve the target only against the indexed `files` table, which by construction holds only files in the configured indexed language set. A path outside that set has **no `files` row**, so it was indistinguishable from a deleted file. Acting on the resulting mark-obsolete recommendation would **delete valid guidance**, and these anchors formed a standing doctor noise floor that never cleared.

## Fix

Before declaring `gone`, fall back to a filesystem existence check against the index `source_root` (read from `index_meta`):

- **path binding**, no `files` row:
  - present on disk + bare (no line span) → `current` (area anchor to a real-but-unindexed file)
  - present on disk + spanned (`path:start-end`) → `unverified` (no chunk to hash, but alive — never `gone`)
  - absent → `gone`
- **dir binding**: no indexed files but the directory exists on disk → `current`; else `gone`

A connection without `source_root` meta skips the fallback (pre-fix `gone` behavior), so behavior is unchanged for raw/test connections that don't record it.

Note: `memory doctor` reads the **persisted** `anchor_status` and does not re-validate (deliberate read-only design), so a false `gone` clears on the next validate pass (watcher / incremental gc tail / open paths).

## Tests (TDD)

Added to `index/oracle/tests.rs`, written failing-first:

- `path_binding_to_unindexed_file_present_on_disk_is_current`
- `path_binding_to_missing_file_is_gone`
- `spanned_path_binding_to_unindexed_file_is_unverified`
- `dir_binding_to_unindexed_dir_present_on_disk_is_current`
- `dir_binding_to_missing_dir_is_gone`

`cargo test -p rag-rat-core` (344 passed), `cargo +nightly fmt`, and `cargo clippy --all-targets` all clean.

Fixes #98

### DIFF
--- a/crates/rag-rat-core/src/index/oracle/tests.rs
+++ b/crates/rag-rat-core/src/index/oracle/tests.rs
@@ -3012,7 +3012,7 @@ fn memory_survives_file_move_via_moniker_relocation() {
     )])]);
     run_oracle(&h.conn, TOOL, VERSION, COMMIT, WORKTREE, &bytes, h.root(), None, None).unwrap();
 
-    let report = validate_memories(&h.conn).unwrap();
+    let report = validate_memories(&h.conn, None).unwrap();
     assert!(report.relocated >= 1, "expected a relocation, got {report:?}");
 
     let memory = memory_by_id(&h.conn, &memory_id).unwrap().unwrap();
@@ -3060,7 +3060,7 @@ fn cross_version_moniker_match_requires_kind_corroboration() {
         run_oracle(&h.conn, TOOL, "v-newer", COMMIT, WORKTREE, &bytes, h.root(), None, None)
             .unwrap();
 
-        validate_memories(&h.conn).unwrap();
+        validate_memories(&h.conn, None).unwrap();
         let memory = memory_by_id(&h.conn, &memory_id).unwrap().unwrap();
         let symbol_binding =
             memory.bindings.iter().find(|b| b.binding_kind == "symbol").expect("symbol binding");
@@ -3096,7 +3096,7 @@ fn path_binding_status_after_validate(h: &Harness, path: &str) -> String {
         bind: RepoMemoryBindTarget { path: Some(path.to_string()), ..Default::default() },
     })
     .unwrap();
-    validate_memories(&h.conn).unwrap();
+    validate_memories(&h.conn, None).unwrap();
     let memory = memory_by_id(&h.conn, &created.memory.memory_id).unwrap().unwrap();
     memory
         .bindings
@@ -3162,7 +3162,7 @@ fn spanned_path_binding_to_unindexed_file_is_unverified() {
         },
     })
     .unwrap();
-    validate_memories(&h.conn).unwrap();
+    validate_memories(&h.conn, None).unwrap();
     let memory = memory_by_id(&h.conn, &created.memory.memory_id).unwrap().unwrap();
     let pb = memory.bindings.iter().find(|b| b.binding_kind == "path").expect("path binding");
     assert_eq!(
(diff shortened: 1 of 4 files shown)


skakri:
Thanks @chatgpt-codex-connector — all three addressed in 9145aa6:

1. **Repo-relative paths** — added an `is_repo_relative` guard: a binding that is absolute or contains `..`/root components is treated as not-on-disk (so it can't validate against an out-of-repo file). Test `path_binding_escaping_source_root_is_gone`.
2. **Active worktree root** — `validate_memories` now takes the caller's `storage.source_root()` (the active checkout, set by `open_config`) and only falls back to the persisted `index_meta.source_root` when unknown. Test `validate_prefers_active_root_over_persisted_meta`.
3. **Require a file** — path bindings now use `is_file()` (not `exists()`); a directory occupying a deleted file's name is `gone`. Dir existence stays on the `dir` path. Test `path_binding_to_a_dir_replacing_the_file_is_gone`.

skakri:
@codex review

PR #131, opened 2026-06-14
fix(mcp): serialize logical_symbol_id as a string across serde boundaries (#130)

### LINKED ISSUE #130: MCP: large i64 ids (logical_symbol_id) lose precision as JSON numbers — accept as strings
## Symptom
Passing a `logical_symbol_id` to an MCP tool (e.g. `memory_rebind`, `memory_for_symbol`, `symbol_lookup`) fails with `-32603 logical_symbol_id <wrong-number> not found`. The id is silently rounded in transit, e.g. `2574604874062343519` → `2574604874062343700`.

## Root cause
`logical_symbol_id` is a content-derived 64-bit hash (~2.5e18), far above JSON/JS safe-integer range 2^53 (~9e15). MCP clients parse JSON numbers as f64, so the value can't round-trip losslessly. The arg structs in `crates/rag-rat-mcp/src/tools/mod.rs` type it as `Option<i64>` and accept it as a JSON number.

## Impact
Any MCP tool field carrying a large i64 id is effectively unusable from MCP clients. The CLI is unaffected (it takes the id as a base-10 string). `symbol_id` (small AUTOINCREMENT rowid) is JSON-safe but churns on reindex, so it's not a durable substitute.

## Fix
Accept large ids as **strings** at the MCP arg boundary — a serde `deserialize_with` that takes string-or-number and parses to i64 — for `logical_symbol_id` (and audit other i64 id fields: `start_logical_symbol_id`, `end_logical_symbol_id`, edge ids if hash-derived). Keep emitting them as strings in tool output too, so a round-trip (read an id from one tool, pass to another) is lossless.

## Workaround
Bind/anchor via a JSON-safe target (bare `path`, `commit_hash`, `github`) or use the CLI `rag-rat memory rebind` for logical_symbol_id.

Found while fixing #128 (memory_rebind routing). Severity: P2.

skakri:
## Problem

`logical_symbol_id` is a content-derived 64-bit hash (~2.5e18), far above JSON's 2^53 safe-integer range. As a JSON **number** it can't round-trip: a client that parses numbers as f64 (every JS-based MCP client) silently rounds it — `2574604874062343519` → `2574604874062343700` — so the id is corrupted before the server ever sees it. Every MCP tool taking `logical_symbol_id` (`memory_rebind`, `memory_for_symbol`, `symbol_lookup`, …) was effectively unusable; only the CLI (string arg) worked.

## Fix

These ids now cross **every** serde boundary (CLI `--json`, MCP JSON, MCP TOON) as a **string**.

New `crate::serde_big_id` helper:
- `big_id` (i64) / `big_id_opt` (Option<i64>) — serialize → string; deserialize → accept **string or number** (back-compat for in-range values and older clients).

Applied to:
- **Output** (serialize → string): `SymbolHit`, `LogicalSymbolHit`, `GraphTraversalQuery`, `LogicalSymbol`, `CompareGraphTextQuery`, `RepoMemoryBinding`, `RepoMemoryCallPath`.
- **Input** (deserialize string-or-number): `RepoMemoryBindTarget` + the MCP arg structs, with `#[schemars(with = "Option<String>")]` so the advertised input schema asks clients for a string. `#[serde(default)]` keeps the fields optional now that a custom deserializer is attached.

Small rowids (`symbol_id`/`chunk_id`/`edge_id`) stay numbers — they're within JSON's safe range.

## Tests
- `serde_big_id` unit tests: serialize-as-string, >2^53 lossless round-trip, deserialize from string/number/null.
- The in-process MCP memory test now reads `logical_symbol_id` from `symbol_lookup` output **as a string** and passes it back through the arg structs into `memory_create`/`memory_for_symbol`/`impact_surface` — proving the end-to-end round-trip (the fixture's id is a content-hash > 2^53). It would have failed (`as_i64` → None) before the fix.

Full workspace tests, clippy, and `cargo +nightly fmt` clean.

## Compatibility note
CLI `--json` and MCP output now emit these ids as JSON strings instead of numbers. Pre-1.0; correct per JSON's integer limits. Deserialize still accepts numbers, so existing in-range callers keep working.

Fixes #130

### DIFF
--- a/crates/rag-rat-core/src/lib.rs
+++ b/crates/rag-rat-core/src/lib.rs
@@ -7,6 +7,7 @@ pub mod locks;
 pub mod output;
 pub mod query;
 pub mod search;
+pub mod serde_big_id;
 pub mod storage;
 pub mod watch;
 
--- a/crates/rag-rat-core/src/query/graph/mod.rs
+++ b/crates/rag-rat-core/src/query/graph/mod.rs
@@ -50,6 +50,8 @@ pub struct GraphTraversalReport {
 pub struct GraphTraversalQuery {
     pub tool: String,
     pub symbol_id: Option<i64>,
+    // Content-derived 64-bit hash > 2^53 — emit as a string so JSON clients don't round it (#130).
+    #[serde(serialize_with = "crate::serde_big_id::big_id_opt::serialize")]
     pub logical_symbol_id: Option<i64>,
     pub symbol_path: String,
     pub resolution: String,
@@ -57,6 +59,7 @@ pub struct GraphTraversalQuery {
 
 #[derive(Debug, Clone, Serialize)]
 pub struct LogicalSymbol {
+    #[serde(serialize_with = "crate::serde_big_id::big_id::serialize")]
     pub logical_symbol_id: i64,
     pub qualified_name: String,
     pub variant_count: u64,
@@ -174,6 +177,8 @@ pub struct CompareGraphTextReport {
 #[derive(Debug, Serialize)]
 pub struct CompareGraphTextQuery {
     pub symbol_id: Option<i64>,
+    // Content-derived 64-bit hash > 2^53 — emit as a string so JSON clients don't round it (#130).
+    #[serde(serialize_with = "crate::serde_big_id::big_id_opt::serialize")]
     pub logical_symbol_id: Option<i64>,
     pub symbol_path: String,
     pub pattern: String,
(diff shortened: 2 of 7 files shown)


skakri:
Good catch @chatgpt-codex-connector — fixed in the latest commit. The numeric deserialize path is now bounded to ±(2^53-1) (`MAX_SAFE_INTEGER`); a `logical_symbol_id` arriving as a JSON number above that is rejected with a clear "pass it as a string" error rather than silently binding a possibly-rounded id. The string path is unbounded (exact). Added `unsafe_number_is_rejected_so_a_rounded_id_never_silently_binds` plus a boundary-value (`2^53-1`) back-compat assertion.

PR #136, opened 2026-06-14
feat: crates.io version check surfaced to agents + operators (opt-out via rag-rat.toml)

skakri:
Surfaces "a newer rag-rat is published" to the agent — and through it the operator — with the exact update command, so a stale install gets noticed instead of silently lagging.

## What it does
- **`core::version_check`** — the one HTTP call in core (`ureq`, rustls, 4s timeout, mandatory `User-Agent` since crates.io 403s without one). **Fail-open**: offline / non-200 / parse miss → no version info, never an error. Result is **cached** to `<index-dir>/version-check.json`.
- **Surfaces**:
  - SessionStart digest line: `rag-rat 0.5.0 (latest on crates.io)`, or when behind `⚠ rag-rat update available: 0.5.0 → 0.6.0 — run \`cargo install rag-rat --force\``.
  - `index_status` MCP tool gains a structured `version` field: `{ current_version, latest_version, update_available, update_command, checked_at_ms }`.
- **Never blocks**: the digest and `index_status` only *read* the cache. The network refresh runs **out of band** — a detached thread on the long-lived `rag-rat mcp` server (once/day TTL), or synchronously via the new **`rag-rat version-check`** command.
- **Opt-out**: `[version_check] enabled = false` in `rag-rat.toml` (default opted in) → zero network calls. Documented in `docs/config.md`.

## Design notes
- HTTP (`ureq`) added at the **core edge** deliberately — per the request, it's reused by future GitHub/GitLab/Bitbucket REST integration. rustls (no system OpenSSL); gzip off (responses are tiny).
- Version compare is lenient numeric semver (`major.minor.patch`, pre-release/build stripped); if either side won't parse, no update is claimed.

## Verification
- Unit tests: version compare, status builder (update/up-to-date/unknown), `needs_refresh` TTL, cache round-trip, config opt-out parse.
- Live: `rag-rat version-check --json` fetched crates.io (`latest 0.5.0`); MCP `index_status` shows the `version` field; SessionStart digest renders the line.
- Full `cargo test` (core 363 + mcp/cli), `cargo clippy --all-targets`, `cargo +nightly fmt` clean.

The many one-line diffs in tests/`watch.rs`/`query_api.rs`/bench harness are just the new `version_check: Default::default()` field on `Config` literals.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3162,6 +3162,7 @@ dependencies = [
  "tree-sitter-kotlin-ng",
  "tree-sitter-rust",
  "tree-sitter-typescript",
+ "ureq 2.12.1",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,9 @@ serde_json = "1.0"
 sha2 = "0.11"
 thiserror = "2.0"
 tokio = { version = "1.48", features = ["rt-multi-thread", "macros", "io-std"] }
+# Minimal blocking HTTP for the crates.io version check (and future GitHub/GitLab/Bitbucket REST).
+# rustls (no system OpenSSL); gzip off — the responses are tiny. The only HTTP at the core edge.
+ureq = { version = "2", default-features = false, features = ["tls"] }
 # TOON (Token-Oriented Object Notation): the default render for CLI + MCP output. ~34% fewer tokens
 # than compact JSON on uniform-row payloads, neutral on nested ones — see rag_rat_core::output.
 # default-features = false drops its `cli` feature (ratatui/crossterm/arboard/syntect/tiktoken-rs/…)
--- a/crates/rag-rat-cli/src/claude_hook.rs
+++ b/crates/rag-rat-cli/src/claude_hook.rs
@@ -286,8 +286,12 @@ fn version_line(status: &rag_rat_core::version_check::VersionStatus) -> Option<S
             "\n⚠ rag-rat update available: {} → {} — run `{}`\n",
             status.current_version, latest, status.update_command
         ))
-    } else {
+    } else if latest == status.current_version {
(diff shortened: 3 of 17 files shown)


skakri:
Addressed in 18fadcf + the coverage commit:

**Periodic refresh (the P2):** `spawn_detached_version_refresh` is now a **poll loop**, not startup-only — it refreshes-if-stale, then sleeps a 6h cadence and re-checks, so a release that lands while the MCP server stays up across the daily TTL is picked up within the TTL (no restart / manual `version-check` needed). The actual crates.io call is still gated by `needs_refresh` (~once/TTL), fail-open and non-blocking; the thread dies with the process and a hot-upgrade re-exec re-spawns it.

**Coverage:** added tests for the previously-thin paths — `parse_latest_response` (valid/malformed/missing/non-string, network-free), `cached_status` enabled (cache present + absent), the pure `version_line` digest formatter (behind / up-to-date / unknown), and an MCP test that `index_status` surfaces the `version` field. The remaining uncovered lines are the I/O edges (the actual crates.io GET, the cache write, the poll thread, and the CLI command's print path).

skakri:
@codex review

skakri:
All three addressed:

- **Null shape (P2):** dropped `skip_serializing_if` on `latest_version`/`checked_at_ms` — they now serialize as `null` when unknown instead of vanishing, so the version object has a stable shape consumers can test `latest_version == null` against. Added a serialization test.
- **Fresh result before cache (P2):** `refresh()` now returns the fresh `CachedVersion`; `version-check` reports that directly and only falls back to `read_cache` when the network fetch itself failed — so a successful check still reports even if the cache write didn't land (read-only checkout / full disk).
- **Ahead-of-latest (P3):** `version_line` now three-ways — behind → update nag; equal → `(latest on crates.io)`; local build ahead of the published latest → `(ahead of crates.io latest X)` rather than claiming it's the latest. Added a test.

Issue #139, opened 2026-06-14
grep-augment 'Indexed hits' are duplicated (same chunk emitted 2-3x)


skakri:
## Problem
The grep-augment digest's **Indexed hits** block repeats the identical chunk multiple times. Observed many times this session, e.g.:

```
- crates/rag-rat-cli/src/claude_hook.rs:67-67 — const SEARCH_COMMANDS ...   (×3 identical)
- crates/rag-rat-mcp/src/tools/handlers.rs:413-464 — ...                    (×3 identical)
- crates/rag-rat-core/src/index/ai/store.rs:338-353 — ...                   (×3 identical)
- crates/rag-rat-core/src/query/grep_augment.rs duplicate memory lines too
```

The same `path:start-end` chunk (and sometimes the same bound memory) appears 2–3× in one digest. It wastes tokens and reads as broken.

## Likely site
The lexical/indexed-hits lane in `crates/rag-rat-core/src/query/grep_augment.rs::compose` — the indexed-hits rendering appears to lack a dedup by `(path, start_line, end_line)` (or chunk id). The memory lane already has insertion-order dedup (`seen` set, grep_augment.rs ~220); the indexed-hits list seems to miss the equivalent. Possibly the same chunk is returned once per matching lane/term and not collapsed.

## Fix
Dedup indexed hits by chunk identity (path + line span, or chunk_id) before rendering, preserving first-seen order — mirroring the memory lane's existing dedup.

## Repro
Run any rag-rat-augmented `grep` for a symbol that has a few near-identical matches; the digest's Indexed hits show repeats. (Reproduced continuously this session.) Found while reviewing augmentation quality.

Issue #109, opened 2026-06-14
Spike: code-specific reranker over hybrid (BM25+vector) candidates


skakri:
**Spike.** Add a second-stage reranker over the top-k hybrid (BM25+vector) candidates using a code-trained cross-encoder or ColBERT-style late-interaction model (CodeRankEmbed / CoRNStack reranker is small + open). Ship as an optional installed model like the existing embedders.

**Why it helps an agent:** dramatically better top-3 precision → the agent reads the *right* chunk first instead of the 7th. Fewer tool round-trips, less context burned. Research consensus: reranking moves code-retrieval quality more than ANN does.

**Scope:** rerank stage after `search/lexical.rs` candidate generation; optional model in the model manifest; measure lift on the commit-replay eval (#120).

Ref: `docs/plans/2026-06-14-agent-value-strategy.md` §4.2.

skakri:
Cross-link: the grep-augment lane is a concrete consumer of this reranker. The augmentation design (see #138) wants E5 dual-use — the same embedder powers a SetFit intent gate *and* this candidate reranker (with a confidence floor that can inject nothing). Justifying E5 across both is what makes adding the dependency worth it; see #138 for the full layering + distribution (base = rules-only, E5 opt-in via `models install`).

skakri:
## Resolved — the reranker spike is concluded (both families measured).

Built and A/B'd two rerankers over the hybrid candidates on the #120 commit-replay eval (200 cases, at-head). recall@3 was added as the top-3-precision metric — a within-top-10 reorder can't move recall@10, and mrr@10 is near-saturated.

- **Structure-aware graded-git rerank** (grade the binary git signal into a recency/churn magnitude + generated/test demotion, at the inner wide-pool site): **+0.008 recall@3 — within noise** (19 cases up / 12 down). Ships behind `[search] graded_git_rerank`, **off by default**.
- **Code cross-encoder** (bge-reranker-base over the top-50): **−0.075 recall@3, −0.147 mrr@10, +24s/query (18× slower)** — actively *hurts*. A general NL cross-encoder mis-ranks (commit message, code) by construction. Rejected, not shipped.

### Why reranking doesn't win here (ceiling analysis)
[...]

PR #140, opened 2026-06-14
fix(grep-augment): skip pipe-incidental greps + dedup indexed hits (#138, #139)

### LINKED ISSUE #138: grep-augment hook fires on incidental `grep` in pipelines (cargo/gh/git | grep), injecting irrelevant context
## Problem
The PreToolUse grep-augment hook triggers whenever a Bash command contains one of `SEARCH_COMMANDS` (`crates/rag-rat-cli/src/claude_hook.rs:67` — `["grep", "rg", "ag"]`), regardless of whether grep is the *primary* command or just a downstream **pipe filter** on another tool's output. In a real working session the majority of grep usages are filters, e.g.:

```
cargo clippy ... | grep -E "warning:|error"
cargo test ...   | grep -E "test result"
gh run view ...  | grep -iE "error|fatal"
git log ...      | grep -i "import.scope"
```

For these, the "search pattern" is a log-filter alternation (`error|warning`, `test result`), not a code search. The hook strips the regex and matches it against memories/symbols, surfacing **near-random, irrelevant context** — observed this session: a `cargo clippy | grep` fired the "Vector search is a brute-force flat scan" memory; a `git push` pipeline fired "Memory relocation requires a content-hash match." Pure token overhead on most Bash calls.

## Impact
Low signal-to-noise across a session: high value when it hits a genuine code search, but it fires far more often on incidental pipe-greps where it injects noise + costs tokens (and the model has to read past it).

## Fix ideas
- Only trigger when the search command is the **primary** command, not a pipe consumer: if the command contains `|` and the grep/rg/ag is downstream of another tool (`cargo`/`gh`/`git`/`cargo test`/etc.), skip augmentation.
- And/or suppress when the pattern looks like an **output filter** rather than a source search: a regex alternation of log-ish tokens (`error|warning|test result|FAILED|...`), or a pattern with no plain identifier after regex-stripping.
- Conservative bar: augment standalone `grep/rg/ag <pattern> <paths>` over source; skip `<cmd> | grep <filter>`.

Trigger logic: `extract_search` / `SEARCH_COMMANDS` in `crates/rag-rat-cli/src/claude_hook.rs`. Found while reviewing rag-rat augmentation quality across a long session.

skakri:
The two concrete augmentation-quality fixes from the session review.

### #138 — stop firing on incidental pipe-greps
The PreToolUse hook triggered whenever a Bash command *contained* `grep`/`rg`/`ag`, including when it merely filters another tool's output (`cargo test | grep "result"`, `gh run view | grep -i error`, `cargo clippy | grep -E warning`). It then matched the log-filter pattern against memories/symbols and injected near-random context — noise on the majority of Bash calls.

`split_top_level` now tags each segment with a **`piped`** flag (true iff preceded by a single `|`; `||`/`&&`/`;`/`&`/first-segment are not piped), and `parse_bash_search` **skips a search command that consumes a pipe**. A grep that is the pipeline *head* (`rg foo | head`) or a sequenced command (`cd x && rg …`) is a real search and still matches. Aligns with the existing "false negatives are fine, false positives are not" contract.

The prior test that asserted `git log | rg fix` → matches `fix` encoded exactly this bug — moved to the negatives.

### #139 — dedup "Indexed hits"
The lexical lane rendered the same chunk 2–3× when `search_lexical_only` returned it more than once (one row per matched FTS term), capped at `MAX_LEXICAL_HITS=3` → up to 3 identical lines. Extracted `lexical_lines_from_hits` and dedup by `(path, start, end)` after the relevance floor, preserving rank order.

## Tests
- `bash_parser_table`: + pipe-incidental negatives (`git log | rg fix`, `cargo test | grep result`, `gh run view | grep -i error`, `cargo clippy | grep -E warning`) and a pipeline-head positive (`rg foo | head`).
- `lexical_lines_dedup_chunks_and_apply_floor`: a chunk repeated 3× collapses to one line; a below-floor hit is dropped.
- Full `cargo test` (core query + cli claude_hook), clippy `--all-targets`, `cargo +nightly fmt` clean.

The deeper layers (symbol-resolution gate, SetFit-E5 intent tiebreaker, reranker) remain tracked in #138's design comment + #109/#112; this PR is Layer 1 + the dedup.

Fixes #138, #139

### DIFF
--- a/crates/rag-rat-cli/src/claude_hook.rs
+++ b/crates/rag-rat-cli/src/claude_hook.rs
@@ -102,7 +102,7 @@ pub fn parse_bash_search(command: &str) -> Option<(String, Option<String>)> {
         return None; // substitution: ambiguous
     }
     // Split into pipeline/sequence segments; examine each for a search command.
-    for segment in split_top_level(command) {
+    for (piped, segment) in split_top_level(command) {
         let tokens = shell_tokens(&segment)?;
         let mut tokens = tokens.as_slice();
         // Skip env-var prefixes (FOO=bar) before the command word.
@@ -115,6 +115,14 @@ pub fn parse_bash_search(command: &str) -> Option<(String, Option<String>)> {
             return None; // grep as an argument of these is ambiguous
         }
         if !SEARCH_COMMANDS.contains(&base) {
+            continue;
+        }
+        // A search command DOWNSTREAM of a pipe is filtering another tool's output
+        // (`cargo test | grep …`, `gh … | rg …`), not a code search — skip it so augmentation
+        // doesn't fire on incidental greps (#138). A grep that is the pipeline HEAD (`grep … |
+        // head`) or a sequenced command (`cd x && rg …`) is a real search and still
+        // matches.
+        if piped {
             continue;
         }
         let mut pattern: Option<String> = None;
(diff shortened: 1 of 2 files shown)


skakri:
Addressed the `|&` review + an adversarial pass over the implementation:

- **`|&` (P2, you + an adversarial review both caught it):** `|&` is bash shorthand for `2>&1 |` — the `&` arm was resetting `piped=false`, so `cargo test |& grep result` still augmented. The `|` arm now recognizes `|&` and keeps the next segment piped; `||` stays non-piped. Added `cargo test |& grep result` / `cargo clippy |& rg warning` to the negatives.
- **Symbol lane dedup (adversary found it live):** the `#139` duplicate-render defect also affected the *symbol* lane (`symbol::lookup` returns one row per overload/cfg-variant/re-export → N identical 'Known symbols' lines — visible in this session as `clusters.rs:104` ×3). Now deduped within the call by `(path, qualified_name)`; test `symbol_lane_dedups_duplicate_rows` (3 dup rows → one line).

The adversary otherwise found only acceptable-per-contract false negatives (e.g. `cd x | rg foo`) and unreachable cases (all-negative lexical scores, given the scoring is strictly positive). All tests green; clippy clean.

skakri:
@codex review

PR #82, opened 2026-06-12
SCIP phase 2: oracle run, Compiler tier, resolved-external, compare_graph_to_scip (#69)

skakri:
Fixes #69 (part of the #61 SCIP-oracle epic). Builds on #68 (the `index/oracle/` reader + join + eval + V018 `oracle_runs`/`edge_oracle` tables, merged on `main`). All of this is **read-side**: the heuristic `edges` row is never overwritten (the #68 side-table invariant holds), and every new `edge_oracle` read routes through `store.rs` (the #81 source-scan invariant stays green).

## Design point → change

**1. `oracle/manifest.rs` — the tool registry.** Per-tool `program` / `languages` / `install_hint`, `scip_command(root, output)`, and `probe()` (runs `<program> --version`). Registers `rust-analyzer scip`. A missing/unrunnable tool → `ToolAvailability::Blocked` with an install hint — **never an error** (same UX as a missing embedding model). The seam later language backends (#71/#72/#74) extend with one entry each.

**2. CLI `rag-rat oracle run [--tool <id>] [--scip <path>]` + `rag-rat oracle status`.** (clap derive in `cli.rs`; dispatch in `main.rs`; bodies in `commands.rs`.) Either invoke the indexer to produce a `.scip` (temp file) then run the phase-1 join, or consume a pre-built `--scip`. Records an `oracle_runs` row; prints the `OracleReport`. Missing tool → `Blocked` + hint, **exit 0**. A present tool that fails to produce SCIP → real error (exit 1) — verified on-device.

[...]

PR #142, opened 2026-06-14
feat: confidence-aware + SCIP-aware symbol importance ranking (#108)

### LINKED ISSUE #108: Spike: PageRank symbol importance ranking over the edge graph
**Spike / research.** Rank symbols by importance using weighted PageRank over the existing call/type/import edge graph (~226k edges already built), personalized toward the symbols in the agent's current query / changed files (Aider repo-map approach).

**Why it helps an agent:** tells it which symbols are load-bearing *before* it edits — the best "don't reinvent / don't break the spine" signal. Also the cheapest input to the risk score (#115).

**Scope:** weighted PageRank pass over the edge graph; expose rank per symbol + a "load-bearing symbols for this area" view (extends `repo_brief`). Personalization vector biased toward query/changed-file symbols.

**Cost:** low — the graph exists. Near-free leverage.

Ref: `docs/plans/2026-06-14-agent-value-strategy.md` §4.1.

skakri:
## Summary

Implements confidence-aware, SCIP-aware **symbol importance ranking** (#108) and threads the resulting signal through the read surface so agents see "what's load-bearing here" inline.

### What's in it
- **PageRank importance** (`query/pagerank.rs`): confidence-weighted, SCIP-aware ranking over the call graph. Edge weights respect resolver confidence and oracle (compiler-tier) corroboration.
- **Load-bearing analysis** (`query/load_bearing.rs`): scoped weighted fan-in signal — the `importance: { label, signal, score, bucket }` block now attached to symbols/chunks in query output.
- **Enrichment on the read path**: `impact_surface` and `semantic_search` carry importance; `impact_surface` CALLEE neighbors are resolved via `to_symbol` first (correct enrichment).
- **`important_symbols`** MCP tool + CLI: clamped result count, scope-consistent hydration, personalization.
  - **Personalize by name**: a bare `--personalize <name>` seeds ALL in-scope symbols matching the name (teleport set), never skip-on-ambiguity.
  - **Auto-seed from git diff** (MCP): the MCP path seeds from the working-tree diff; CLI stays global — intentional divergence.
- **Opt-in background auto-fresh oracle** (`index/oracle/auto_run.rs`) + a ranking nudge for the #108 UX.

### Notes
- Importance shows as `label: "local structural load"`, `signal: "scoped weighted fan-in"`, with a `bucket` (critical/high/low).
- Config + docs updated (`docs/config.md`).

Closes #108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,14 @@ language = "markdown"
 directories = ["."]
 include = ["**/*.md"]
 exclude = [".git/**", ".rag-rat/**", "target/**", "node_modules/**"]
+
+# Optional: background refresh of compiler-grade (SCIP) importance ranking. Off by default.
+# Needs a language tool on PATH (e.g. rust-analyzer); the throttled pass runs only in the MCP
+# server. `rag-rat init` asks whether to enable this.
+[oracle]
+auto_run = false
+# auto_run_quiet_period_secs = 900    # wait this long after the last index change before a run
+# auto_run_min_interval_secs = 21600  # never re-run more often than this
 ```
 
 Then run the pieces directly:
@@ -602,6 +610,10 @@ name = "android-kotlin"
 language = "kotlin"
 directories = ["android/src/main/java"]
 include = ["**/*.kt"]
+
+# Opt-in background compiler-grade ranking refresh (off by default; needs e.g. rust-analyzer).
+[oracle]
+auto_run = false
 ```
 
 ## Git Hooks
--- a/crates/rag-rat-cli/src/cli.rs
+++ b/crates/rag-rat-cli/src/cli.rs
@@ -54,6 +54,9 @@ pub(crate) enum Command {
     /// Ownership / co-change clusters.
     Clusters(ClustersArgs),
 
+    /// Rank the most load-bearing symbols by weighted PageRank over the edge graph.
+    ImportantSymbols(ImportantSymbolsArgs),
+
     /// Run the stdio MCP server.
     Mcp,
 
@@ -145,6 +148,13 @@ pub(crate) struct BriefArgs {
     /// Omit drive-by repo memories.
     #[arg(long)]
     pub no_memories: bool,
+}
+
+#[derive(Debug, Args)]
+pub(crate) struct ImportantSymbolsArgs {
+    /// Max load-bearing symbols to return.
+    #[arg(long)]
+    pub limit: Option<u32>,
 }
 
 #[derive(Debug, Args)]
--- a/crates/rag-rat-cli/src/cli.rs
+++ b/crates/rag-rat-cli/src/cli.rs
@@ -155,6 +155,10 @@ pub(crate) struct ImportantSymbolsArgs {
     /// Max load-bearing symbols to return.
     #[arg(long)]
     pub limit: Option<u32>,
+    /// Symbol ids to bias importance toward (the symbols you're working on) — comma-separated or
+    /// repeated. Empty = global importance.
+    #[arg(long, value_delimiter = ',')]
+    pub personalize: Vec<i64>,
 }
 
 #[derive(Debug, Args)]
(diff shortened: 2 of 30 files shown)


skakri:
Thanks — addressed the review in 9e8e4e2.

- **Report global mode when diff seeds never enter the graph** ✅ `pagerank::important_symbols` now returns `RankedImportance { symbols, effective_seed_count }`. `important_symbols` decides the mode by *effective* seeds (how many landed as graph nodes), so a resolved-but-isolated seed is labeled `Global` with a reason (`"diff symbols are not connected in the graph"` / `"named symbols are not connected in the graph"`) instead of `PersonalizedToChanges`. `SeedSource` gained `effective_seed_count` for transparency. Test: `seed_resolving_only_to_isolated_symbols_is_labeled_global`.

- **Serialize the oracle tier as the documented label** ✅ `OracleTier` now uses `#[serde(rename_all = "snake_case")]`, so it serializes as `"compiler"`, matching `as_str()` and the contract. Test: `oracle_tier_serializes_as_lowercase_compiler` (asserts both the enum and the `ImportanceEnrichment.oracle_tier` field on the wire).

- **Fall back when auto-diff runs outside Git** ✅ `diff_seed` treats any `git_changed_paths` error as an empty diff and falls through to global, instead of propagating the error. Test: `auto_seed_outside_a_git_worktree_falls_back_to_global`.

- **Let the watcher catch up before auto-run** ✅ The background oracle thread now sleeps one poll interval *before* its first decision, so the MCP watcher's initial pass indexes working-tree changes first — avoiding a first run against the stale pre-watcher index (which would record skipped/drifted docs and then min-interval-block a corrected run).

- **Avoid whole-oracle scans for small read enrichments** ⚠️ partially. The three load-bearing enrichers now early-return when there are no hits/neighbors, so empty results pay nothing. For non-empty results I kept the single fetch-once-and-reuse of the gated verdict map rather than per-candidate queries: the `any_run_in_scope` existence probe already makes the dominant no-oracle case ~free, and for repos *with* oracle data, one scan reused across all hits is cheaper than N per-candidate verdict queries. Happy to revisit a per-candidate path if profiling on a large oracle-backed repo shows it matters.

Also wired up the `[oracle]` config surfacing: `rag-rat init` now asks whether to enable `auto_run` (default no) and always renders the `[oracle]` section, and the README example configs show it.

skakri:
@codex review

skakri:
Round 2 addressed in a0cabdf (then merged up to date with main):

- **P1 — include oracle-upgraded NULL-target edges** ✅ The edge query required `to_symbol_id NOT NULL`, which dropped exactly the upgrade cases SCIP-aware ranking exists for (a `NameOnly`/unresolved heuristic edge whose only resolved target is the compiler verdict). Now keep `from`-non-null edges and let `EdgeOracleEffect::Retarget` supply the target; a NULL-target edge with no Retarget is skipped. Test: `oracle_upgrade_ranks_a_recovered_null_target_edge`.
- **Avoid enriched lookups while resolving seeds** ✅ `resolve_seed_selectors` uses the unenriched `lookup_candidates` (it discards everything but `symbol_id`), so no per-seed oracle scan + fan-in.
- **Skip oracle scans when search hits have no symbols** ✅ `enrich_search_hits_with_load_bearing` returns early when no hit has a `symbol_path`.
- **Rewrite ranking hints for MCP auto-run** ✅ `call_tool_for_config` now applies the `RANKING_HINT_AUTO_RUN` rewrite (mirroring the CLI) when `[oracle] auto_run` is on. Test: `mcp_rewrites_ranking_hint_when_auto_run_enabled`.

- **Use the oracle run's actual start time** → tracked as #145, not bundled here. It's a real (self-healing) P2 but the fix touches the oracle run-recording contract (`record_oracle_run` / `OracleRunInput` / `run_oracle` + callers, ~9 sites) and the #82/#83 TOCTOU machinery — a different subsystem from importance ranking, so it gets its own focused change rather than expanding this PR's blast radius.

Issue #77, opened 2026-06-12
Compress stored chunk text — dictionary-trained zstd at the blob level (~3.8 GB → <1 GB on the kernel)


skakri:
Surfaced from the ~10 GB kernel-index DB analysis — stored **chunk text dominates** (~3.8 GB of the ~10 GB, kept verbatim).

## Low-risk win: dictionary-trained zstd at the blob level
- Train a **shared dictionary per language** on a few thousand sampled chunks (the `zstd` crate handles dictionaries natively). Source compresses unusually well with a shared dict — **4–6× on C** is typical → ~3.8 GB to under 1 GB.
- **No semantics touched:** decompress on read; updates, deletes, and FTS re-tokenization all keep working unchanged. The dictionary is versioned (train once at index/build time; store the dict + version with the DB).
- This is the one-change, no-coupling option.

## Rejected for now (document as a future option, not phase 1)
Dropping stored text entirely and **materializing from the git object store** — clean rows are pinned by `commit_sha` and blobs are immutable, so it's the lazy-blame pattern applied to text. Architecturally pretty, but a real footgun:
- **External-content FTS5 requires reproducing the EXACT original text on update/delete** to keep the index consistent — so reconstruction becomes *correctness-critical*, not just a read path.
- `gc` pruning of dead commits acquires an **ordering dependency with FTS deletes**.

Compression gets most of the bytes with none of that coupling. **Recommendation: ship compression; leave materialization as a documented future option.**

skakri:
### Relation to #143 (MCP read-lock contention)

Surfaced again while fixing #143 ("database is locked" under concurrent MCP clients on the kernel index). Worth recording how the two relate so neither is mistaken for the other:

- **#143 (fixed in #144)** removes read-side lock contention directly — read tools no longer take the SQLite write lock on open (read-first `ensure_model_manifest` + read-only connections). It changes *who* takes the write lock.
- **#77 (this issue)** shrinks the DB (~3.8 GB stored chunk text → <1 GB on the kernel via dictionary-trained zstd). A smaller DB means faster WAL checkpoints and shorter write-lock **hold times**, so any remaining writer (the watcher, a heal, a reconcile) blocks others for less time and is less likely to blow the busy_timeout. It changes *how long* a writer holds the lock.

[...]

PR #144, opened 2026-06-14
fix(mcp): read tools open read-only so a writer can't lock them out (#143)

### LINKED ISSUE #143: MCP read tools take the SQLite write lock on open → intermittent "database is locked" under concurrent clients
## Symptom

A second concurrent MCP client (e.g. Codex while another agent's MCP server is also attached) intermittently gets **`database is locked` (SQLITE_BUSY)** from read-only tools on a large index (observed on the Linux-kernel index, ~38.8k files, ~1M chunks).

## Root cause

Every MCP tool call opens the DB via `IndexDatabase::open_config` → `IndexConnection::open` (a **read-write** connection), and `open_with_graph_check` unconditionally runs `ensure_model_manifest(...)` on open. That helper issues write DML on **every** open regardless of whether anything changed:

- `remove_legacy_models` → 3× `DELETE`
- `upsert_model` ×3 → `INSERT … ON CONFLICT DO NOTHING`
- `normalize_embedding_model_versions` → `UPDATE`

In rusqlite every DML statement takes the SQLite **write lock** for its implicit transaction, even when it matches zero rows. So **every read tool acquires the write lock on open.** WAL allows only one writer; with a live watcher and/or a second MCP instance also opening connections, these per-open writes serialize, and when one holder runs past the 5s `busy_timeout`, the next returns `SQLITE_BUSY` → "database is locked." It is intermittent because it only fires on overlap.

(`ensure_graph_index_current` is *not* the culprit in steady state — it early-returns when `graph_index_version` matches.)

## Fix (this issue)

**A. Make `ensure_model_manifest` read-first.** Guard the writes behind a cheap "is the manifest already current?" read so a steady-state open performs **zero** writes. After the first open establishes the manifest, every subsequent open is read-only. Protects all RW callers (CLI, watcher, write tools), not just the MCP.

**B. Route pure-read MCP tools through a real read-only connection.** Read tools open with `SQLITE_OPEN_READ_ONLY` (5s busy_timeout) and skip the on-open heal writes (`ensure_model_manifest` / `ensure_graph_index_current`). A read-only connection can never take the main write lock, so reads become structurally immune to any writer. If the index needs a heal/migrate (schema not `Compatible`, graph version stale, or manifest not current), the read path falls back once to the RW open so healing still happens; subsequent reads are lock-free again.

Read tools: everything except the writers `heal_index`, `memory_create`, `memory_rebind`, `memory_update`, `memory_mark_obsolete`, `memory_validate`.

## Complementary, not part of this issue: #77

#77 (zstd-compress stored chunk text, ~3.8 GB → <1 GB on the kernel) helps on a **different axis** — a smaller DB shortens checkpoint / write-lock hold times, so the busy_timeout is less likely to be blown. It does not change *who* takes the write lock. A+B remove the read-side contention directly; #77 shortens writer hold time. Track #77 separately.

skakri:
Fixes #143.

## Problem

A second concurrent MCP client (e.g. Codex while another agent's MCP server is also attached) intermittently gets **`database is locked`** from read-only tools on a large index (observed on the Linux-kernel index).

Every MCP tool call opened the DB read-write (`call_tool_for_config` → `IndexDatabase::open_config` → `IndexConnection::open`), and `open_with_graph_check` ran `ensure_model_manifest` on **every** open. That helper issues unconditional write DML (`remove_legacy_models` DELETEs, 3× `upsert_model` INSERT, `normalize_embedding_model_versions` UPDATE). In rusqlite every DML statement takes the SQLite **write lock** for its implicit transaction even when it changes zero rows — so **every read tool acquired the write lock on open**. With a live watcher and/or a second MCP instance, those per-open writes serialize; when one holder runs past the 5s `busy_timeout`, the next returns `SQLITE_BUSY`. Intermittent because it only fires on overlap.

(The RW path already had a 5s busy_timeout; the 100ms one is only the grep-augment fallback. `ensure_graph_index_current` is not a steady-state culprit — it early-returns when the graph version matches.)

## Fix (two independent layers)

**A — `ensure_model_manifest` is read-first.** A new `model_manifest_is_current` short-circuits the writes when the manifest already matches what they would produce, so a steady-state open performs **zero** writes. Protects all read-write callers (CLI, watcher, write tools), not just the MCP.

**B — pure-read MCP tools open read-only.** `try_open_config_read_only` opens a `SQLITE_OPEN_READ_ONLY` connection (5s busy_timeout) and skips the on-open heal writes. A read-only connection can never take the main write lock, so reads are structurally immune to any writer. It declines (`None` → read-write fallback that heals once) when a write is still owed: schema not `Compatible`, stale graph index, or a not-yet-current manifest. Tool classification is a writer **deny-list** (a new tool is read-only by default), guarded by a test.

## Tests
- `ensure_model_manifest_does_not_write_when_already_current` — proves the no-op by running it on a read-only connection.
- `read_only_open_serves_current_index_and_declines_when_heal_is_owed`.
- `read_only_classification_covers_every_tool_and_denies_writers`.

Full `rag-rat-core` + `rag-rat-mcp` suites pass; clippy clean.

## Not in this PR
#77 (zstd-compress stored chunk text) is complementary on a different axis — it shortens how long any writer holds the lock, but doesn't change who takes it. Tracked separately.

### DIFF
--- a/crates/rag-rat-core/src/index/ai/reconcile.rs
+++ b/crates/rag-rat-core/src/index/ai/reconcile.rs
@@ -1,6 +1,15 @@
 use super::*;
 
 pub(crate) fn ensure_model_manifest(conn: &Connection) -> anyhow::Result<()> {
+    // Read-first: skip the (write-locking) DML entirely when the manifest already matches what we
+    // would write. `ensure_model_manifest` runs on EVERY `IndexDatabase::open*`, so issuing
+    // unconditional INSERT/UPDATE/DELETE here made every open — including read-only MCP tools —
+    // take the SQLite write lock, serializing them against the watcher and other clients and
+    // surfacing "database is locked" under concurrency (#143). After the first open establishes
+    // the manifest, every later open is a handful of SELECTs and never touches the write lock.
+    if model_manifest_is_current(conn)? {
+        return Ok(());
+    }
     remove_legacy_models(conn)?;
     upsert_model(conn, HASH_MODEL_ID, "embedding", Some(HASH_EMBEDDING_DIM), "hash", false)?;
     upsert_model(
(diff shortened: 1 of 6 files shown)


skakri:
Good catch — fixed in 902c487.

You're right that `semantic_search` (FTS heal), `read_chunk` (`mark_file_deleted`/`heal_file`), and `git_blame_chunk` (`store_blame`) lazily write on cold paths. Rather than deny-list them — which would make the *hottest* tool (`semantic_search`) always take the write lock, defeating the point of the fix — the dispatcher now **detects a `SQLITE_READONLY` violation and transparently retries the whole call on a read-write connection** (which performs the heal):

- `core`: `storage::is_readonly_violation(&anyhow::Error)` walks the cause chain for `rusqlite::ErrorCode::ReadOnly`.
- `mcp`: `call_tool_for_config` runs the read tool read-only; on a read-only violation it falls through to the read-write open and re-runs the call. The warm path never writes, so it stays lock-free; the caller never sees `SQLITE_READONLY`.

So the deny-list now means "always writes" (memories, heal_index); tools that only *sometimes* write stay read-only and rely on the retry.

Tests added: `is_readonly_violation_flags_only_sqlite_readonly_errors` (storage) and `read_tool_lazy_write_retries_read_write_not_readonly_error` (deletes a chunk's source file so `read_chunk` takes the `mark_file_deleted` write path, and asserts the caller gets the domain "gone" error, never a read-only violation).

skakri:
@codex review

Issue #145, opened 2026-06-14
oracle_runs.started_at is completion time, not start time → auto-run staleness gate can wrongly block a compiler refresh


skakri:
## Problem

`oracle_runs.started_at` is stamped by `record_oracle_run` (oracle/store.rs) with `now_ms()` **at completion**, not when the run actually began. The background auto-fresh oracle's staleness gate (`auto_run_decision`) compares `last_run_started_at` against the index's `last_index_change_ms`:

- A background run starts, then a watcher reindex lands **during** the (multi-second) run.
- The run's content gates correctly skip the drifted documents (#82/#83), so it records no verdict for them.
- But `started_at` is written at completion, so `last_run_ms > last_index_change_ms` ⇒ the next tick returns `NotStale` ⇒ compiler-grade ranking for those edits **never refreshes** until another edit or a manual `oracle run`.

Flagged by Codex review on #142 (the importance-ranking PR). Self-healing (next edit unblocks it), hence not bundled into #142 — the fix touches the oracle run-recording contract, a different subsystem.

## Fix

Record `started_at` as the run's actual start — the moment the pre-spawn snapshot is captured (the same instant `pre_spawn_sha` is taken in `commands::oracle_run` and the background `run_oracle_tool_background`). Thread an explicit `started_at_ms` through `run_oracle` / `OracleRunInput` / `record_oracle_run` (clock injected, per the codebase convention), defaulting to `now_ms()` for paths with no controlled spawn moment (`--scip`, tests).

~9 call sites: `store::record_oracle_run`, `run::OracleRunInput` + `run::run`, `oracle::run_oracle` (+ `run_oracle_with_tool`), `IndexDatabase::run_oracle` (+ `run_oracle_from_scip`), and callers in `commands.rs` / `main.rs` (background) / tests.

## Test

A run whose `started_at` predates an index change must read as `Stale` on the next `auto_run_decision` tick (assert against `latest_run_started_at`), so the compiler refresh is not wrongly gated `NotStale`.

skakri:
Fixed in #146 (merged): oracle runs now record `started_at` as the run's real start (the pre-spawn snapshot, captured under the write lock), and the latest-run readers order by `id DESC` (completion order) so an overlapping pair is keyed to the run whose verdicts are actually live. The auto-run staleness gate no longer wedges at NotStale after a run overlaps a watcher reindex.

Issue #49, opened 2026-06-09
impact_surface: surface a caveat when a section is truncated at LIMIT (no silent caps)


skakri:
## Problem

The `impact_surface` symbol-report sections (`tests_touching_symbol_path`, `docs_mentioning_symbol_path`, `text_fallback_hits`, and the import/export + git/papertrail sections) silently `truncate(limit)` their results. When more files match than `limit`, the response gives no indication that it was capped — an agent reading it can't distinguish "these are all the matches" from "these are the first N of many."

This violates the repo's **"no silent caps — log/surface what was dropped"** convention.

## Context

Follow-up to #48. The one-row-per-file collapse landed in 8c923e4 makes hitting the cap far rarer (LIMIT now counts distinct files, not fanned-out symbol rows), so this is lower priority — but the silent truncation itself is still there.

## Fix sketch

Detect per-section overflow and surface it honestly, e.g.:
- Fetch `limit + 1` and, if the extra row comes back, drop it and set a `truncated: true` flag / push a caveat into `completeness_and_caveats.caveats` like `"tests_touching_symbol_path truncated at N files; more matches exist."`
- Or run a cheap `COUNT(DISTINCT files.path)` per section and report `shown` vs `total`.

Prefer the `limit + 1` probe — it avoids a second scan per section/needle and is enough to flag "there is more." Wire the flag/caveat through `ImpactSurfaceReport` so the MCP `completeness_risk` logic can also factor it in (same spirit as the degraded-coverage escalation in #47).

skakri:
Fixed in #146 (merged): the structured `impact_surface` report now lists every section capped at `limit` in `truncated_sections` plus a caveat — including the `repo_memories` section (per-lane pre-split count). The residual flat/query-only `Vec<ImpactItem>` shape (free-text query + allow_ambiguous fallback) has no completeness slot and is a separate compat-shape decision, tracked in #150.

PR #146, opened 2026-06-14
fix: oracle started_at = run start (#145); impact_surface flags truncated sections (#49)

skakri:
Two pre-release fixes from the open-issue triage.

## #145 — oracle `started_at` is the run's start, not completion

`oracle_runs.started_at` was stamped with `now_ms()` *inside* `record_oracle_run`, i.e. at completion. The background auto-fresh oracle's staleness gate compares the latest run's `started_at` against the index's last-change clock — so a run that overlapped a watcher reindex (its content gates correctly skip the drifted docs) recorded a completion time *after* that change, making the gate read `NotStale` and **never refresh compiler ranking** until the next edit or a manual run.

Fix: capture the real start (before the tool subprocess, at the pre-spawn snapshot) and thread it through new explicit-start entry points — `store::record_oracle_run_at`, `oracle::run_oracle_at`, `IndexDatabase::run_oracle_at` — used by `oracle run`, the background auto-run, and `run_oracle_with_tool`. The original `now_ms()`-stamping names remain for callers with no controlled spawn moment (pre-built `--scip`, tests), so the wide test surface is untouched.

## #49 — `impact_surface` no longer caps sections silently

Each `impact_surface` section was truncated at `limit` with no signal. The completeness block now reports every section that returned exactly `limit` rows in a new `truncated_sections: [..]` field and adds a caveat, so a truncated list never reads as complete coverage. (Graph envelope tools already had `summary.truncated`; this brings impact in line.)

## Tests
- `record_oracle_run_at_persists_the_passed_start_time` — recording with a past start time reads back that value, not completion.
- `impact_report_flags_a_section_truncated_at_limit` — 3 callers + limit=2 → `direct_semantic_callers` named in `truncated_sections` + a caveat; a generous limit flags nothing.

Full `rag-rat-core` (419) + `rag-rat-mcp` (20) + CLI suites green; clippy + nightly-fmt clean.

### DIFF
--- a/crates/rag-rat-cli/src/commands.rs
+++ b/crates/rag-rat-cli/src/commands.rs
@@ -296,13 +296,14 @@ fn oracle_run(config: &Config, args: &OracleRunArgs) -> anyhow::Result<()> {
     // including DURING the subprocess, which the post-exit `production_sha` snapshot cannot see —
     // is skipped, never mis-joined. A reindex slipping in between this lock release and the spawn
     // is detected by the same gate.
-    let pre_spawn_sha = with_oracle_write_lock(config, |db| db.oracle_pre_spawn_snapshot())?;
-    // Stamp `started_at` right AFTER the pre-spawn snapshot (which read the indexed state this run
-    // covers) and before the tool subprocess: later than the indexed_at it's based on (so a run
-    // that covers fresh state isn't falsely judged stale even if the write-lock wait was long),
-    // yet before any mid-run reindex (so a run that misses one IS judged stale). (#145 + #146
-    // review)
-    let started_at_ms = crate::now_epoch_ms();
+    // Stamp `started_at` INSIDE the same write-lock as the pre-spawn snapshot, so no watcher
+    // reindex can land between reading the indexed state and recording the start. Under the lock,
+    // started_at corresponds exactly to the indexed state this run covers: ≥ that indexed_at (so a
+    // run covering fresh state isn't falsely judged stale even after a long lock wait) yet before
+    // any mid-run reindex (so a run that misses one IS judged stale). (#145 + #146 review)
+    let (started_at_ms, pre_spawn_sha) = with_oracle_write_lock(config, |db| {
+        Ok((crate::now_epoch_ms(), db.oracle_pre_spawn_snapshot()?))
+    })?;
     let scip_output = config
         .database
         .parent()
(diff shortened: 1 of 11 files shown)


skakri:
All three addressed in b6c1bd2:

- **Include memory evidence in truncation reporting** ✅ `repo_memories` is now part of the truncation check — `memory_evidence_for_symbol_and_edges` caps each lane at `limit`, so `truncated_sections` (and the caveat) flag `repo_memories` when any active lane (`direct`/`path_crossed`/`call_path_crossed`) fills the cap. Test extended: 3 memories on `hub` + limit=2 now asserts `repo_memories` is reported.

- **Keep staleness tied to the last completed write** ✅ The "latest run" readers (`latest_run_started_at`, `latest_run_tool_version`, `status::last_run_meta`) now `ORDER BY id DESC` = insertion/completion order (the run row is written at the end of `run::run`, under the write lock), not `started_at DESC`. So when two runs overlap and finish in the opposite order they started, the staleness clock follows the run whose `edge_oracle` verdicts are actually live (highest id), and returns *its* start time.

- **Stamp the run after the pre-spawn snapshot lock** ✅ `started_at` is now captured right *after* the pre-spawn snapshot (the indexed state the run covers) in all three sites (`commands::oracle_run`, the background auto-run, `run_oracle_with_tool`). A long write-lock wait while the watcher indexes can no longer record a start earlier than the state the run covers (which would force a needless rerun); it's still stamped before the subprocess, so a mid-run reindex is still correctly treated as stale.

Full core (419) + mcp (20) + CLI suites green; clippy + nightly-fmt clean.

skakri:
@codex review

skakri:
Round 3 — addressed in 694edb6:

- **Stamp oracle start while the snapshot lock is held** ✅ `started_at` is now captured *inside* the same `with_oracle_write_lock` closure as the pre-spawn snapshot, in both `commands::oracle_run` and the background path: `let (started_at_ms, pre_spawn_sha) = with_oracle_write_lock(.., |db| Ok((now_epoch_ms(), db.oracle_pre_spawn_snapshot()?)))?`. No watcher reindex can interleave between reading the indexed state and stamping the start, so `started_at` is atomic with the state the run covers.

- **Count stale memory lanes when flagging truncation** ✅ `memory_evidence_for_symbol_and_edges` now returns `(evidence, truncated)`, where `truncated` is computed from each lane's **pre-split** row count (`>= limit`) — before `split_active_stale` moves rows to `stale`. The impact builder uses that flag instead of the active-lane-max heuristic, so an all-stale or mixed result that filled the cap is correctly flagged.

- **Add truncation reporting to query-only impact results** → tracked as **#150**, not in this PR. The flat `Vec<ImpactItem>` path (free-text `query` + the `allow_ambiguous` fallback) is the documented compatibility shape with no completeness slot; adding it means either changing that shape or a side-channel — a separate design decision rather than a drop-in. #49's structured/symbol-selected report (the primary path) is fully covered here; #150 carries the flat-shape gap with options.

The three round-2 comments (commit 4bb811b) were already fixed in b6c1bd2. Full core (419) + mcp (20) + CLI suites green; clippy + nightly-fmt clean.

Issue #149, opened 2026-06-14
Wire contract: drop ephemeral symbol_id, make logical_symbol_id an opaque `sym_<hex>` handle


skakri:
## Problem

Two symbol ids leak onto the MCP wire, and both are footguns:

- **`symbol_id`** is an `INTEGER PRIMARY KEY AUTOINCREMENT` rowid. A per-file reindex does `DELETE+INSERT` (`internals.rs` heal/rebuild), so the id is **reassigned on every reindex** — observed live: `is_readonly_violation` 875582 → 875938 after a one-line edit. An agent that caches a `symbol_id` from one call and passes it to the next (across an edit) gets a dangling handle. There is no safe way to use it across edits.
- **`logical_symbol_id`** is the *stable* handle (63-bit truncation of `SHA-256(language, path, name, qualified_name, kind, signature)`; `stable_id` in `index/mod.rs`). But it's a ~6×10¹⁸ number, serialized as a **decimal string** (`serde_big_id`, #130) so JSON clients don't round it past 2⁵³. A decimal string still *looks* like a number and invites `parseInt()` → silent rounding/corruption.

## Decision

One **opaque, stable, self-describing** symbol handle on the wire; the ephemeral rowid never leaves the engine. **Clean break — no transitional support.**

1. **Drop `symbol_id` from the MCP surface entirely** — both result output and `SymbolSelector` input. It stays internal-only: it remains the rowid + FK target (`edges.from/to_symbol_id` SET NULL, `symbol_facts.symbol_id` / `logical_symbol_members.symbol_id` CASCADE, `repo_memory_bindings.symbol_id`). **No schema change** — this is purely the query/MCP serialization + selector boundary.
2. **Make `logical_symbol_id` an opaque token `sym_<hex>`** (lowercase hex of the positive 63-bit id, prefixed). Lossless like the decimal string, but unmistakably *not* a number, so no client rounds or does math on it. Encode/decode lives in `serde_big_id` + the `SymbolSelector` parser. (Keep the field name `logical_symbol_id`, token-valued, to minimize churn — or rename to `handle`; implementer's call.)
3. **No `symbol_id` input** is accepted anymore; agents pass the `sym_<hex>` handle, or `symbol_path` (human-readable secondary). Internal callers/tests keep using the raw `symbol_id` directly.

## Surface that moves together (one coordinated change)

[...]

PR #151, opened 2026-06-14
feat: lazy-heal symbol_lookup + flag dirty result files (#147, #148)

### LINKED ISSUE #147: symbol_lookup/impact return stale line numbers until the watcher reindexes — extend read_chunk's lazy-heal-on-read to the symbol path
## Problem

`symbol_lookup` (and the symbol lane of `impact_surface` / the grep-augment hook) returns **raw stored symbol rows with no current-source validation** — `SELECT symbols.start_byte, symbols.end_byte, … FROM symbols` (`crates/rag-rat-core/src/query/symbol.rs:158-160`). After a file is edited, those byte/line positions are stale until the debounced watcher reindex runs (`debounce_ms = 400`, capped at `max_latency_ms = 2500`; `watch.rs:82`, `watch.rs:385`). So for ~0.4–2.5 s after a write — exactly when an agent is iterating fastest — `symbol_lookup` reports pre-edit line numbers.

`read_chunk` already solves this for chunks: it validates the stored anchor against current source and relocates small drift (or reports stale/gone), and `semantic_search` heals stale files before returning. The symbol path never got that treatment.

## Proposal

Apply the same lazy-heal-on-read the chunk path uses to the symbol/impact paths:
- Before composing `symbol_lookup` / `impact_surface` results, check the matched file(s)' on-disk sha vs `files.sha256`; if drifted, run the existing bounded per-file heal (the capped, `needs_reindex`-on-overflow path already used by search/read_chunk) so positions reflect current source.
- At minimum, relocate the returned symbol span against current source the way `read_chunk` relocates a chunk anchor.

This closes the stale-position window **deterministically** instead of racing the 400 ms debounce. Do NOT lower `debounce_ms` as the fix — that fights the self-sustaining-reindex-loop risk (see the watcher `BugPattern`).

## Acceptance

- After editing a file, a `symbol_lookup` for a symbol in it returns current line numbers without waiting for the watcher.
- Heal stays bounded (same cap as the chunk path; `needs_reindex` beyond it).
- Covered by a test: edit shifts a symbol's lines, lookup returns the new lines.

### LINKED ISSUE #148: Surface a per-result staleness flag when a file is dirty relative to the index (symbol_lookup / search / impact)
## Problem

When a file has been edited but the debounced watcher reindex hasn't run yet (~0.4–2.5 s window, longer if queued behind the write lock), query results silently describe the pre-edit snapshot. `read_chunk` surfaces drift (relocated/stale/gone) and the graph envelopes carry a `coverage.stale_files` count, but `symbol_lookup`, `semantic_search` hits, and `impact_surface` don't give a per-result signal that the underlying file is dirty relative to the index. An agent iterating on a file can't tell a fresh answer from a stale one.

## Proposal

Add a cheap, uniform freshness signal on results whose file's on-disk sha differs from the indexed `files.sha256`:
- A boolean `stale: true` (and/or `indexed_sha` vs current) on `symbol_lookup` candidates, search hits, and impact sections — or at least a top-level `stale_files: [..]` note alongside the existing `coverage` block.
- It's a sha compare per touched file (cheap), no reindex required — pairs with #147 (which heals) but stands alone as honesty even before healing lands.

This is squarely the project's "no silent staleness / provenance on every result" posture: a truncated section is now flagged (#49); a stale-relative-to-disk result should be too.

## Acceptance

- A result drawn from a file whose disk content differs from the index is marked stale.
- A clean-file result is not marked.
- Negligible cost on the common (fresh) path — one sha read per touched file, skipped when nothing is dirty.

skakri:
Implements the freshness-after-write fixes surfaced from the agent dogfooding session.

## #147 — `symbol_lookup` lazy-heals stale positions
Symbol rows carry stored `start_byte`/`end_byte` with **no anchor relocation** (unlike `read_chunk`, which validates+relocates against current source). So after you edit a file, `symbol_lookup` returns pre-edit line numbers until the debounced watcher reindexes (~0.4–2.5 s) — exactly when you're iterating.

`symbol_candidates` now: detect matched files dirty vs the index (`disk sha != files.sha256`) → heal them inline (bounded by `MAX_AUTO_HEAL_FILES_PER_CALL`, `NeedsReindex` beyond it — same contract as `search_with_heal`) → re-resolve so positions/ids are current. A `symbol_id` selector that can't survive the reindex (ids are reassigned, see #149) keeps its pre-heal candidates flagged stale rather than returning empty.

## #148 — flag dirty result files
- `SymbolLookup` gains `stale_files` (residual after heal; empty in the common case).
- `impact_surface` populates `completeness.stale_files` over the selected symbol's file, the caller/callee call-site files, and the current-source item sections. **Flag-only** for impact (not heal): its neighbor set is unbounded, so an inline heal can't be bounded the way symbol_lookup's matched-file heal can — the agent re-runs `symbol_lookup` (which heals) for a specific symbol if it needs fresh positions.
- `semantic_search` already heals stale hits via `search_with_heal`, so it needs no separate flag.

Shared helpers on `IndexDatabase`: `stale_source_paths` (one read+hash per distinct result path), `heal_stale_paths` (bounded heal + FTS sync), `indexed_sha_for_path`.

## Tests
- `symbol_lookup_heals_stale_line_numbers_after_an_edit` — edit shifts a symbol; lookup by name returns the new position, `stale_files` empty.
- `impact_completeness_flags_dirty_result_files` — editing the symbol's file makes `completeness.stale_files >= 1`.

Full core (421) + mcp (20) + CLI green; clippy + nightly-fmt clean.

Closes #147
Closes #148

### DIFF
--- a/crates/rag-rat-core/src/index/query_api.rs
+++ b/crates/rag-rat-core/src/index/query_api.rs
@@ -416,8 +416,88 @@ impl IndexDatabase {
     ) -> anyhow::Result<crate::query::symbol::SymbolLookup> {
         let mut lookup =
             crate::query::symbol::lookup_candidates(self.storage.connection(), selector)?;
+        // #147: symbol rows aren't anchor-relocated like chunks, so a file edited since indexing
+        // returns stale line numbers. Heal the matched files inline (bounded, like
+        // search_with_heal) and re-resolve so positions/ids are current; #148: report any
+        // file still dirty after.
+        let paths: Vec<String> = lookup.candidates.iter().map(|c| c.path.clone()).collect();
+        let stale = self.stale_source_paths(&paths)?;
+        if !stale.is_empty() {
+            self.heal_stale_paths(&stale)?; // NeedsReindex beyond the cap
+            let healed =
+                crate::query::symbol::lookup_candidates(self.storage.connection(), selector)?;
+            if healed.candidates.is_empty() && !lookup.candidates.is_empty() {
+                // A `symbol_id` selector can't survive a reindex (ids are reassigned per #149), so
+                // the re-resolve finds nothing. Keep the pre-heal candidates rather than return
+                // empty, but flag the files as stale so the caller knows the positions are old.
+                lookup.stale_files = stale;
+            } else {
+                lookup = healed;
+                let healed_paths: Vec<String> =
+                    lookup.candidates.iter().map(|c| c.path.clone()).collect();
+                lookup.stale_files = self.stale_source_paths(&healed_paths)?;
+            }
+        }
         self.enrich_symbol_hits_with_load_bearing(&mut lookup.candidates)?;
         Ok(lookup)
+    }
+
+    /// The indexed `files.sha256` for `path` in the ACTIVE checkout (via the per-connection `files`
+    /// scope view), or `None` when the path isn't indexed in this scope.
+    fn indexed_sha_for_path(&self, path: &str) -> anyhow::Result<Option<String>> {
+        Ok(self
+            .storage
+            .connection()
+            .query_row("SELECT sha256 FROM files WHERE path = ?1 LIMIT 1", [path], |row| {
+                row.get::<_, String>(0)
+            })
+            .optional()?)
+    }
+
+    /// Of `paths`, those whose on-disk content differs from the indexed sha (or are unreadable) —
+    /// results drawn from them may be stale relative to the working tree. Deduped; paths not
+    /// indexed in scope are skipped (nothing to be stale against); no source root → empty
+    /// (bare/copied index). One file read + hash per distinct path — callers pass the small
+    /// result set, not the whole corpus.
+    fn stale_source_paths(&self, paths: &[String]) -> anyhow::Result<Vec<String>> {
+        let Some(root) = self.storage.source_root().map(Path::to_path_buf) else {
+            return Ok(Vec::new());
+        };
+        let mut stale = Vec::new();
+        let mut seen = std::collections::BTreeSet::new();
+        for path in paths {
+            if !seen.insert(path.as_str()) {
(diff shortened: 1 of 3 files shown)


skakri:
Addressed in cd043af:

- **Don't preserve stale candidates after name re-resolve (P1)** ✅ The pre-heal fallback is now guarded by `selector.symbol_id.is_some()`. A name/symbol_path/logical selector whose re-resolve comes back empty after the heal means the symbol was genuinely deleted/renamed by the edit → we return the empty healed result, not a ghost with dead ids/old offsets. Only a `symbol_id` selector (id reassigned on reindex, #149) keeps its pre-heal candidate, flagged stale. Test: `symbol_lookup_does_not_resurrect_a_deleted_symbol_after_heal` (+ `symbol_lookup_by_id_keeps_pre_heal_candidate_flagged_stale` pins the preserved id-selector behavior).

- **Track callee target files when counting impact staleness (P2)** ✅ `impact_surface`'s `completeness.stale_files` now resolves each returned direct-callee's target qualified name to its **definition file** (new `file_for_qualified_name` helper) and includes it, so a callee whose own file was edited (call-site file untouched) is counted. Test: `impact_completeness_flags_a_dirty_callee_definition_file` (edits only the callee's file, asserts `stale_files >= 1`).

- **Heal zero-hit symbol lookups from dirty files (P2)** → tracked as **#152**, deliberately not inline. The heal set on a zero-hit can only come from a `git status` (working-tree change set), which would run on *every* failed name lookup — and the common cause of a zero-hit is a genuine miss, not a just-added symbol. That's real per-call cost for a window the watcher already closes in ≤2.5 s. #152 has a hard-bounded design if we decide it's worth it.

Full core (424) + mcp (20) green; clippy + nightly-fmt clean.

PR #153, opened 2026-06-14
feat: opaque sym_<hex> symbol handle; drop ephemeral symbol_id from the wire (#149)

### LINKED ISSUE #149: Wire contract: drop ephemeral symbol_id, make logical_symbol_id an opaque `sym_<hex>` handle

skakri:
Implements the wire-contract change specced in #149.

## What changed

The MCP/CLI wire exposed two symbol ids, both footguns:
- **`symbol_id`** — autoincrement rowid, reassigned on every reindex (dogfooded: 875582→875938 after a one-line edit), so it could never be cached safely across an edit.
- **`logical_symbol_id`** — stable but a ~6×10¹⁸ number; decimal-string-serialized (#130) yet still *number-shaped*, inviting `parseInt` → silent 2⁵³ rounding.

Now: **one opaque, stable handle on the wire; the ephemeral rowid never leaves the engine.** Clean break, no transitional support.

- **`serde_big_id::sym_handle` / `sym_handle_opt`** — `logical_symbol_id` crosses serde as an opaque **`sym_<hex>`** token (hex of the u64 bits, lossless for any i64). Deserialize accepts **only** the token, so a stale numeric id passed by habit fails loudly. Replaces the now-dead `big_id` (deleted).
- **`symbol_id` is internal-only** — `#[serde(skip_serializing)]` on every output struct (`SymbolHit`, `LogicalSymbol{Hit,Member,Variant}`, `SymbolImportance`, graph query echoes, `ImpactSurfaceQuery`, memory binding), removed from the MCP arg structs, and `#[serde(skip_deserializing)]` on the memory bind input. It stays the rowid + FK target in-process.
- **`important_symbols`** now emits `logical_symbol_id` too (LEFT JOIN `logical_symbol_members`), so every symbol-returning result carries a usable handle.

## Granularity (as decided in #149)

The handle addresses the *logical* symbol; per-cfg-body wire addressing is gone (overloads/different-signature symbols still get distinct handles). `resolution: "exact"` on the handle returns the whole group's edges — the realistic query. `ffi_surface` is unaffected (it emits qualified-name strings, never an id).

## Docs + tests

- `docs/mcp-tools.md` + `AGENTS.md`: handle described; `symbol_id` documented as gone from the wire; examples use `sym_<hex>`.
- `serde_big_id` tests: token round-trip (incl. negative/`i64::MIN`/MAX), reject bare number / missing prefix.
- The MCP disambiguation test round-trips the handle through `find_callers` / `compare_graph_to_text` / `papertrail_for_symbol` and asserts candidates carry no `symbol_id` and a `sym_`-prefixed handle. Schema test asserts the selector schema lacks `symbol_id`.

Full core (421) + mcp (20) + CLI suites green; clippy + nightly-fmt clean.

Closes #149

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 .rag-rat/
 target/
+
+# Plans/specs are kept on disk for review, never committed (see dont-commit-plans).
+docs/plans/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,10 +27,12 @@ Reach for these first:
 - **`important_symbols`** — load-bearing symbols by (SCIP-aware) PageRank; pass `personalize` to
   bias toward what you're editing. Compiler-grade once `rag-rat oracle run` has run.
 
-**Symbol handle:** symbol-returning tools emit `logical_symbol_id`, an opaque `sym_<hex>` token —
-the stable handle to cache and pass back into graph/impact/memory tools (copy verbatim; never parse
+**Symbol handle:** symbol-returning tools emit `id`, an opaque `sym_<hex>` token — the stable handle
+to cache and pass back into graph/impact/memory tools as the `id` param (copy verbatim; never parse
 it as a number). There is no numeric `symbol_id` on the wire (it's an internal rowid reassigned on
-every reindex). Use `symbol_path` for the human-readable identity.
+every reindex). Use `ref` (the `path::name` qualified name) for the human-readable identity. The
+symbol-tool params are `ref` / `id` / `lang` (formerly `symbol_path` / `logical_symbol_id` /
+`language`).
 
 Why this beats grep here:
 - Results carry **provenance**: confidence labels, coverage warnings, and raw evidence, so you can
@@ -59,7 +61,7 @@ Codex's own store) is invisible to the others. rag-rat is the **cross-agent memo
 anything another agent would benefit from here.
 
 How to do it well:
-- **Anchor to the tightest stable target.** Prefer a `logical_symbol_id` binding (the `sym_<hex>`
+- **Anchor to the tightest stable target.** Prefer an `id` binding (the `sym_<hex>` logical-symbol
   handle — self-heals across cross-file moves via the relocation engine); fall back to a `path`
   binding for file/area-level notes, or a commit/GitHub ref for historical rationale.
 - **Pick the right `kind`:** `Invariant` (must stay true), `Decision`/`RejectedAlternative` (why it's
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,6 +27,11 @@ Reach for these first:
 - **`important_symbols`** — load-bearing symbols by (SCIP-aware) PageRank; pass `personalize` to
   bias toward what you're editing. Compiler-grade once `rag-rat oracle run` has run.
 
+**Symbol handle:** symbol-returning tools emit `logical_symbol_id`, an opaque `sym_<hex>` token —
+the stable handle to cache and pass back into graph/impact/memory tools (copy verbatim; never parse
+it as a number). There is no numeric `symbol_id` on the wire (it's an internal rowid reassigned on
+every reindex). Use `symbol_path` for the human-readable identity.
+
 Why this beats grep here:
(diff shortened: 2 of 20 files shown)


skakri:
@codex review

PR #157, opened 2026-06-15
fix(impact): signal truncation on the flat impact_surface shape (#150)

### LINKED ISSUE #150: impact_surface flat/query-only path still truncates silently (#49 only covered the structured symbol report)
## Problem

#49 added truncation reporting (`truncated_sections` + a caveat) to the **structured** `impact_surface` report — the symbol-selected path (`impact_surface_report_for_symbol`). But the MCP `impact_surface` handler also returns the **legacy flat shape** (`Vec<ImpactItem>` from `impact_surface_with_options`) for:
- free-text `query` requests (`handlers.rs:555`), and
- the `allow_ambiguous` symbol fallback when a name doesn't uniquely resolve (`handlers.rs:546`).

That flat path caps at `limit` via `surface.into_items(max_items)` (`impact/mod.rs`) with **no completeness field**, so those results can still be silently truncated and read as complete — #49 unfixed for the flat shape.

## Why it's separate from #49

The flat `Vec<ImpactItem>` is the documented compatibility shape ("Free-text `query` requests retain the older flat impact item shape for compatibility"). Adding completeness there means either changing that shape (a breaking change to the surface) or a side-channel (e.g. a trailing sentinel item) — a deliberate design decision, not a drop-in like the structured path. Hence tracked on its own rather than expanding the #49/#145 PR.

## Options

1. Give the flat path a structured envelope (`{ items, truncated_sections, caveats }`) — cleanest, but breaks the compat shape; needs a version/opt-in.
2. Append a synthetic trailing `ImpactItem` (category "completeness") when `items.len() == limit` — visible, no shape change, slightly hacky.
3. Route the ambiguous-symbol fallback (`handlers.rs:546`) through the structured report instead of the flat path, leaving only true free-text `query` on the flat shape — shrinks the gap.

## Acceptance

- A flat `impact_surface` result that hit `limit` signals it (chosen option), or the ambiguous/symbol paths no longer use the flat shape.
- No silent cap on any `impact_surface` invocation.

skakri:
Fixes #150.

## Problem
#49 added truncation reporting (`truncated_sections`) to the **structured** `impact_surface` report, but the **flat** `Vec<ImpactItem>` shape still capped at `limit` silently. The flat shape is returned for:
- free-text `query` requests (`handlers.rs`), and
- the `allow_ambiguous` symbol fallback when a name doesn't uniquely resolve.

Both route through `impact_surface_from_targets` → `ImpactSurface::into_items`, which sorts and `truncate(limit)`s with no completeness signal — so a capped flat result reads as complete.

## Fix
Chose **option 2** from the issue (visible signal, no breaking shape change): when the combined surface overflowed `limit`, append a trailing completeness sentinel `ImpactItem` (`category`/`kind` = `"completeness"`) carrying the dropped count. The compatibility `Vec<ImpactItem>` shape is unchanged; a reader can detect the sentinel structurally.

The fix lives in the flat path (`impact_surface_from_targets`) only. The structured report builds its per-section vecs through the *same* `into_items` but reports truncation via `truncated_sections` — so `into_items` itself is left untouched and no sentinel leaks into a structured section. This also covers both flat callers at once (free-text query *and* the ambiguous fallback), since both go through this function.

eval's impact scoring matches by `category`/`path` (the sentinel's `category="completeness"` / `path=""` match nothing) and never asserts vec length, so it's unaffected.

## Test
`impact_surface_flat_signals_truncation_when_capped`: a `hub` with four callees queried at `limit=2` emits the sentinel with exactly `limit` real items and a dropped-count in `evidence`; at `limit=100` no sentinel appears (no crying wolf). Confirmed it fails without the fix and passes with it.

`cargo build` / `clippy --all-targets` / `test` (435 core) / `+nightly fmt --check` all green.

### DIFF
--- a/crates/rag-rat-core/src/index/schema_bootstrap_tests.rs
+++ b/crates/rag-rat-core/src/index/schema_bootstrap_tests.rs
@@ -3081,8 +3081,8 @@ pub fn hub() {
         .expect("a capped flat result must carry a completeness sentinel (#150)");
     assert!(sentinel.reason.contains("capped"), "sentinel explains the cap: {sentinel:?}");
     assert!(
-        sentinel.evidence.iter().any(|evidence| evidence.contains("not shown")),
-        "sentinel carries the dropped count: {sentinel:?}"
+        sentinel.evidence.iter().any(|evidence| evidence.contains("beyond the limit")),
+        "sentinel signals more exist: {sentinel:?}"
     );
     let real_items = impact.iter().filter(|item| item.category != "completeness").count();
     assert_eq!(real_items, limit as usize, "exactly `limit` real items, plus the sentinel");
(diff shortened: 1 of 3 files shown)


skakri:
Good catch — both P2 comments are the same root cause: the per-section caps (`textual_fallback` via `remaining`, `historical_evidence` via `max_items`) clamp the surface to exactly `max_items`, so `surface.len() > max_items` never fired for the free-text fallback / history paths and they still truncated silently.

Fixed in 150e75d: probe one past the limit (fetch those sections with a `max_items + 1` budget), detect truncation as `surface.len() > max_items`, and let `into_items(max_items)` discard the probe item. Dropped the precise count from the sentinel since the per-section caps make the true total unknowable cheaply — it now states the cap and that more exist. Added a regression test for the capped-textual-fallback path (a query matching more files than the limit, no symbol/graph hits); verified it fails without the probe.

skakri:
@codex review

skakri:
Right again — the probe is a ROW budget, but the history sections collapse all commits/refs for a path into one item, so a hot file's commits exhausted the budget before later paths added their distinct items, and the probe never overflowed.

Fixed in fdb7d8b: `git_commits_for_paths` / `github_refs_for_paths` now budget by ITEMS ADDED (decrement once per path that contributes an item, per-path evidence bounded by the budget), so one hot file can't starve later paths and the probe overflows → sentinel fires. The structured callers (`git_commit_items`/`github_ref_items`) always pass a single path, so their output is byte-identical (no #49 regression). Added a unit test that a 3-commit hot path doesn't starve a second path's item under a 2-item budget.

skakri:
@codex review

PR #158, opened 2026-06-15
fix(index): heal a just-added symbol on a zero-hit name lookup (#152)

### LINKED ISSUE #152: symbol_lookup zero-hit doesn't heal: a just-added symbol isn't found until the watcher catches up
## Problem

`symbol_lookup`'s lazy heal (#147) derives its heal set from the INITIAL candidates' files. A lookup for a **just-added** symbol (or one renamed into existence) returns zero candidates → empty path set → no heal → it keeps returning nothing until the watcher reindexes (~0.4–2.5 s). So the symmetric case of #147 ("I just wrote `fn foo`, find it") still misses in the sub-debounce window.

Flagged in #151 review.

## Why it's deferred (not in #151)

The only way to know which files to heal on a zero-hit is the working-tree change set, i.e. a `git status` (`git_changed_paths`, as `diff_seed` uses). That would run on **every** failed name lookup — and the common cause of a zero-hit is a genuine miss (wrong/typo'd name), not a just-added symbol. Putting a `git status` on the miss path trades real per-call cost for a narrow window the background watcher already closes within ≤2.5 s. Not worth it inline.

## If we do it, bound it hard

- Trigger only on a **name / symbol_path** selector that returns zero (not `symbol_id`/`logical_symbol_id` — a miss there isn't "newly added").
- Heal only changed source files that are dirty-vs-index OR not-yet-indexed, capped at `MAX_AUTO_HEAL_FILES_PER_CALL`; if more, do nothing (do NOT raise `NeedsReindex` on a zero-hit — that's hostile for a plain miss).
- Then re-resolve once.

## Acceptance

- Adding a symbol then looking it up by name returns it without waiting for the watcher.
- A genuine miss (name that doesn't exist anywhere) adds at most one bounded `git status`, no heal, no error.

skakri:
Fixes #152.

## Problem
`symbol_lookup`'s lazy heal (#147) derived its heal set from the matched candidates' files. A lookup for a **just-added** symbol (or one renamed into existence) returns zero candidates → empty path set → no heal → keeps returning nothing until the watcher reindexes (~0.4–2.5 s). The symmetric case of #147 ("I just wrote `fn foo`, find it") still missed in the sub-debounce window.

## Fix
On a zero-hit for a **name / symbol_path** selector, index the working-tree change set (changed source files that are dirty-vs-index OR not-yet-indexed), re-derive logical symbols + edges, and re-resolve once. Bounded exactly as the issue's "bound it hard" section asks:
- **name/symbol_path selectors only** — a miss on a churning `symbol_id`/`logical_symbol_id` isn't "newly added", so those misses never pay a `git status`.
- **capped at `MAX_AUTO_HEAL_FILES_PER_CALL`** — over the cap does nothing; **never** raises `NeedsReindex` (a plain typo'd miss stays cheap).
- a genuine miss on a clean tree costs one `git status` and no write.

### Config retention
Indexing a brand-new (never-indexed) file needs the target config to classify language/kind (`collect_changed_index_files`). So `IndexDatabase` now retains the `Config` it was opened with, but **only on the config-bearing opens** (`open_config` / `try_open_config_read_only`) — `None` for `rebuild`/`open`/tests, where the heal is a graceful no-op. This mirrors the existing `github`-context "resolved only at the config boundary" pattern.

On the read-only MCP connection (#143) the heal's write trips `SQLITE_READONLY` and the dispatch retries read-write — the same path the #147 heal already uses. The heal reuses the incremental indexer's `collect_changed_index_files` / `assign_file_scopes` / `apply_incremental_file_plan` (widened to `pub(super)`).

## Test
`symbol_lookup_heals_a_just_added_symbol_without_waiting_for_the_watcher`: in a git repo, a brand-new file's symbol is found by name with no reindex; a genuine miss stays empty. Confirmed it fails (`[]`) without the heal.

`cargo build` / `clippy --all-targets` / `test` (435 core) / `+nightly fmt --check` all green.

### DIFF
--- a/crates/rag-rat-core/src/index/incremental.rs
+++ b/crates/rag-rat-core/src/index/incremental.rs
@@ -264,7 +264,7 @@ impl IndexDatabase {
         Ok((count, manifest_in_change_set))
     }
 
-    fn assign_file_scopes(
+    pub(super) fn assign_file_scopes(
         &self,
         files: Vec<IndexFile>,
         changes: &GitChangedPaths,
@@ -355,7 +355,7 @@ impl IndexDatabase {
         Ok(healed)
     }
 
-    fn apply_incremental_file_plan<F>(
+    pub(super) fn apply_incremental_file_plan<F>(
         &self,
         files: Vec<IndexFile>,
         deleted: BTreeSet<PathBuf>,
--- a/crates/rag-rat-core/src/index/lifecycle.rs
+++ b/crates/rag-rat-core/src/index/lifecycle.rs
@@ -50,6 +50,7 @@ impl IndexDatabase {
             active_commit_sha: String::new(),
             active_worktree_id: String::new(),
             github: github::GitHubContext::default(),
+            config: None,
         };
         if check_graph {
             db.ensure_graph_index_current()?;
@@ -65,6 +66,7 @@ impl IndexDatabase {
         // Real usage: resolve the GitHub repo context from the local `gh` CLI here, at the
         // boundary. rebuild/open (used by tests and the bare index command) leave it offline.
         db.github = github::GitHubContext::from_gh();
+        db.config = Some(config.clone());
         db.ensure_graph_index_current()?;
         Ok(db)
     }
@@ -109,6 +111,7 @@ impl IndexDatabase {
             active_commit_sha: String::new(),
             active_worktree_id: String::new(),
             github: github::GitHubContext::from_gh(),
+            config: Some(config.clone()),
         };
         db.set_context(&commit_sha, &worktree_id)?;
         Ok(Some(db))
@@ -165,6 +168,7 @@ impl IndexDatabase {
             active_commit_sha: String::new(),
             active_worktree_id: String::new(),
             github: github::GitHubContext::default(),
+            config: None,
         })
     }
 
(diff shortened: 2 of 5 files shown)


skakri:
All three are real — the bounded heal cut corners vs the incremental change-set apply. Fixed in 19c1232 by giving it the same discipline as `index_incremental_with_progress`:

- **deletions**: pass `changes.deleted` to `apply_incremental_file_plan` (was an empty set), so a removed file's rows don't survive the re-resolve;
- **packages**: `refresh_packages` before `resolve_edges`, so per-package import scope (#61) resolves against the fresh map when a Cargo.toml is in the change set;
- **transaction**: wrap apply + `refresh_packages` + `rebuild_logical_symbols` + `resolve_edges` + `sync_fts` in one `BEGIN IMMEDIATE`/COMMIT (ROLLBACK on error), so a concurrent reader never observes a half-rebuilt index. On the read-only MCP connection the BEGIN trips SQLITE_READONLY and the dispatch retries read-write, same as the #147 heal.

(I prototyped a deletions regression test but dropped it: the #147 `heal_file` path already cleans up a directly-looked-up deleted symbol, so a simple 'is it gone' test passes regardless — it can't isolate the edge-binding case. The disciplines are inherited by reusing the same tested incremental calls.)

skakri:
@codex review

PR #160, opened 2026-06-15
feat(mcp): nudge the agent to re-anchor stale memories via tool-result content

skakri:
Make the **agent** (not the user) eagerly notice and fix drifted repo-memory anchors.

## Why this mechanism
I checked the Claude Code / MCP docs (via the claude-code-guide agent). Key finding: **Claude Code's agent is pull-based.** Server-initiated MCP notifications (`notifications/message`, progress, resource-updated) reach the **UI/logs but not the model's context** — confirmed by [anthropics/claude-code#3174](https://github.com/anthropics/claude-code/issues/3174). The mechanisms that *do* reach the model are: tool-result content (multiple content blocks are shown to the model), and hooks' `additionalContext`. Of those, the only **MCP-server-native** one is tool-result content.

## What this does
Every rag-rat tool already funnels through one `call()` chokepoint in the server. When active memories have gone/stale anchors, `call()` appends a **second content block**:

> rag-rat: N active repo memories have stale/gone anchors. Call `memory_doctor` to list them with suggested re-anchor targets, then `memory_rebind` to fix …

The primary structured (TOON/JSON) block is untouched, so parsers are unaffected. It **self-limits** — once the agent runs `memory_rebind` the count drops to 0 and the block stops appearing, so it nags exactly while there's something to fix.

## Cost / safety
The count is `rag_rat_core::memory_attention_count(database)`: a **bare** read-only sqlite open + one indexed `COUNT` — no git resolution, scope view, or migration (so no `git rev-parse` subprocess per call). It mirrors `doctor_report`'s exact WHERE (active, `anchor_status IN ('gone','stale')`, excluding `scip_moniker`) so the nudge count matches what `memory_doctor` then lists. **Fail-open to 0** on any error — never blocks or fails a tool call.

## Tests
- `doctor_attention_count_counts_active_gone_and_stale_bindings` — counts active gone/stale, excludes obsolete.
- `memory_attention_count_reads_file_db_and_fails_open` — reads a file DB; missing DB → 0.

`cargo build` / `clippy --all-targets` / `test --workspace` (441 core, 25 mcp) / `+nightly fmt --check` all green.

Builds on #159 (the `memory_doctor` tool it points the agent to), now in main / v0.6.0.

### DIFF
--- a/crates/rag-rat-core/src/index/oracle/tests.rs
+++ b/crates/rag-rat-core/src/index/oracle/tests.rs
@@ -2903,8 +2903,8 @@ fn prune_oracle_runs_drops_dead_contexts_only() {
 // ---------------------------------------------------------------------------
 
 use crate::query::memory::{
-    RepoMemoryBindTarget, RepoMemoryCreate, create_memory, memory_by_id, split_active_stale,
-    validate_memories,
+    RepoMemoryBindTarget, RepoMemoryCreate, create_memory, doctor_attention_count, memory_by_id,
+    split_active_stale, validate_memories,
 };
 
 /// Build a multi-document SCIP index, serialized.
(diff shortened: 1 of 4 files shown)


skakri:
Good catch. Fixed in the latest push: the nudge now only rides TOON (the default, LLM-facing) results and is suppressed when launched as `rag-rat mcp --json` — that mode exists for clients parsing/concatenating tool text as JSON, and the nudge is agent-directed prose they shouldn't get. Added a server test asserting TOON carries the nudge block (2 content blocks) while JSON omits it (1 block).

skakri:
@codex review

PR #162, opened 2026-06-15
feat(eval): gate eval behind a non-default feature + add CI eval job

skakri:
Per the eval discussion: stop shipping the eval harness in the client, and make CI exercise it (it was latent — no CI ran it, and the baseline had drifted).

## 1. Removed from the built client (behind a `eval` cargo feature, OFF by default)
The `crate::eval` module + the `rag-rat eval` subcommand and its CLI plumbing (`commands::eval`, `default_eval_path`, `render::print_eval_summary` / `format_oracle_line`) are gated behind `eval` (rag-rat-core + rag-rat-cli). `cargo install` ships **no** `eval` command; `--features eval` brings it back. Verified: default `--help` has no `eval` subcommand; `--features eval --help` does.

## 2. CI eval job (`.github/workflows/eval.yml`)
Builds `--no-default-features --features eval` (hash embedder → deterministic, network-free, like the rest of CI) and runs:
- `clippy … --features eval` — the eval-only code is skipped by the main CI (built `--no-default-features`), so this keeps it from bit-rotting.
- `cargo test … --features eval eval_suite` — runs the eval suite over the `tests/fixtures/held-mini` corpus: asserts no current-source-safety violations and non-zero mrr@10 / recall@10.

## Finding: the strict baseline has drifted
While validating, I found the strict `must_include` baseline (`evals/expected_hits.toml`) currently **misses 3 expectations** — `useBridge` (TS const-arrow hook symbol), `MainBridge::create` (Kotlin companion method), and the `openDatabase` graph target — on the current index, with **both** hash and model2vec embedders (so they're embedder-independent symbol/graph-extraction gaps, not ranking noise). Because the eval was never CI-gated, this went unnoticed.

So this PR gates at the **machinery/smoke** level (green, deterministic). Promoting it to the full `rag-rat eval` `report.pass` gate needs a call: **fix the 3 extraction gaps**, or **rebaseline** `expected_hits.toml` to current behavior. I'll file a follow-up issue; happy to do either.

`cargo build`/`test`/`clippy`/`fmt` green in both feature states (default 437+25 tests; `--features eval` adds the eval suite).

### DIFF
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -1,0 +1,37 @@
+name: eval
+
+# Search-quality + SCIP-oracle eval gate. The eval harness is behind the `eval` cargo feature
+# (OFF in the shipped client), so it's compiled and run only here. Hash embedder
+# (`--no-default-features`) keeps it deterministic and network-free, like the rest of CI.
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+concurrency:
+  group: eval-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+
+jobs:
+  eval:
+    name: eval (search-quality)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+      - uses: Swatinem/rust-cache@v2
+      # The `eval` feature code is skipped by the main CI (which builds --no-default-features), so
+      # lint it here or it bit-rots.
+      - name: clippy (eval feature)
+        run: cargo clippy -p rag-rat-core -p rag-rat --no-default-features --features eval --all-targets --locked -- -D warnings
+      # Runs the eval suite over the held-mini fixture corpus: asserts no current-source-safety
+      # violations and non-zero retrieval (mrr@10 / recall@10), plus the oracle precision/recall
+      # formatting. Deterministic (hash embedder, no model download).
+      - name: eval suite
+        run: cargo test -p rag-rat-core --no-default-features --features eval --locked eval_suite
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -48,12 +48,14 @@ jobs:
         run: cargo clippy -p rag-rat-core -p rag-rat --no-default-features --features eval --all-targets --locked -- -D warnings
       - name: clippy (eval feature, hybrid)
         run: cargo clippy -p rag-rat-core -p rag-rat --features eval --all-targets --locked -- -D warnings
-      # Deterministic, network-free baseline gate: hash embedder → lexical retrieval. The fastembed
-      # reconcile block in the eval suite is compiled out here, so the baseline must hold lexically.
-      - name: eval suite (lexical, deterministic)
+      # Deterministic, network-free smoke gate: hash embedder → lexical retrieval (the fastembed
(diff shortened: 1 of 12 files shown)


skakri:
## Update: security fix + promoted to a real `report.pass` gate (resolves #163)

**Security (CodeQL alert).** Added `permissions: contents: read` to `eval.yml` — the job only checks out and runs cargo, so no write scopes.

**#163 root-caused and fixed.** The 3 baseline failures were not what they looked like:
- `kotlin-companion-create` — format bug. `symbol_path` is `{file_path}::{name}` with the receiver omitted, so `MainBridge::create` never matched. Corrected the expectation to the bare `create`.
- `useBridge` / `openDatabase` graph-neighbor — **not** extraction or embedder-ranking gaps. Their fixture chunks were below `MIN_EMBEDDING_CHARS = 80` (42 and 67 chars), so they were never embedded under *any* embedder; and the eval path (`rebuild`) never reconciled embeddings at all, so eval was lexical-only. Caching a MiniLM download would have changed nothing.

**Made the semantic eval real (the chosen route).**
- Enlarged the two fixture symbols past the 80-char threshold so they embed (they now also rank in top-K lexically).
- The eval suite installs the model + reconciles embeddings under `#[cfg(feature = "fastembed")]` and now asserts `report.pass` — every hand-authored `must_include_*` expectation must hold — over the real **hybrid** (lexical + semantic) retrieval path the product ships.
- CI runs **two** passes: deterministic lexical (`--no-default-features`, network-free) and hybrid (default features, MiniLM) with the model download cached via `RAG_RAT_MODEL_CACHE` + `actions/cache` (static key, immutable model). `report.pass` holds in both.

Verified locally: lexical pass=true (mrr@10 0.889, recall 1.000) and hybrid pass=true (mrr@10 0.833, recall 1.000, `useBridge`/`openDatabase` now retrieved via `hybrid`). clippy clean in both feature shapes; nightly fmt clean.

skakri:
@codex review

skakri:
Thanks — addressed all three Codex P2s in 8b1c4d5, plus the CodeQL finding.

**P2 · file-scope the companion factory** (`evals/queries.toml`) — agreed, bare `create` was too loose given `== || ends_with` matching + independent path/symbol hits. Changed to `src/Main.kt::create` (the actual indexed `symbol_path`), so it can't be satisfied by an unrelated `…recreate` / other-file `create`.

**P2 · avoid lexical seeding of the graph-neighbor fixture** (`tests/fixtures/held-mini/src/index.ts`) — agreed and good catch. Removed the comment that injected `native/bridge/open/database` into the `openDatabase` chunk; enlarged it with neutral code instead. **Consequence I leaned into:** the chunk now no longer ranks top-K on lexical/BM25 alone, so the full `report.pass` baseline genuinely requires embeddings. I gated the `report.pass` assertion behind `#[cfg(feature = "fastembed")]` — the lexical (`--no-default-features`) pass is now an explicit *smoke* gate (compile + current-source safety + non-zero retrieval), and the hybrid pass (MiniLM) enforces the full baseline. This makes the semantic-only expectations actually depend on semantic retrieval rather than seeded lexical terms.

**P2 · mark eval as feature-only where advertised** (`README.md:180`) — done; the command line now notes it requires a `--features eval` build (absent from the released binary).

**CodeQL · workflow permissions** — resolved: `eval.yml` carries `permissions: contents: read` (the alert predates that commit; should clear on re-scan).

All green locally: both eval passes (lexical smoke + hybrid full), clippy in both feature shapes, nightly fmt.

skakri:
@codex review

Issue #164, opened 2026-06-15
Multi-language SCIP oracle + Python (small-corpus runner)


skakri:
Tracking epic for the SCIP-oracle reorg, Python support, and a unified small-corpus oracle runner with a before/after resolution table. Plan: `docs/plans/2026-06-15-scip-oracle-multilang-runner.md` (local, gitignored).

## Goal
1. Reorganize GitHub Actions (consolidate the two near-duplicate `oracle-*` workflows; consistent naming + least-privilege permissions).
2. Add Python indexing + `scip-python` oracle backend.
3. One corpus-profile-driven runner over **small, real** repos (semver / cJSON / requests) emitting a typed before/after resolution report (JSON), with a per-PR Δ-vs-main comment via a glue script. Heavy repos (cargo/kernel) keep producing the Bencher series on release.

## Sequenced sub-tasks
- [ ] **A1** chore(ci): workflow naming + least-privilege `permissions:` sweep; bump deprecated actions (independent)
- [ ] **C0** feat(oracle): report/profile **schema contract + fixture-DB tests** — the stabilizer; locks `OracleResolutionReport` JSON (schema version + provenance envelope + `corpus_profile_hash`), denominator semantics, and fixture tests for every verdict transition (unresolved→upgraded, unresolved→resolved_external, exact→confirmed, exact→contradicted, oracle-only calls, skipped_drifted, monikers)
- [ ] **B1–B5** feat(lang): Python indexing (Language enum, tree-sitter grammar, symbols, edges, policy/fixture/tests incl. the import/alias/decorator bad-signal assertions)
- [ ] **B6** feat(oracle): `scip-python` backend (`OracleTool::ScipPython`, manifest, argv test)
- [ ] **C2** feat(oracle): typed before/after resolution report implementing the C0 contract (TOON + JSON only, **no markdown**); BMF derived from it
- [ ] **C1+C3** feat(ci): corpus-profile-driven `oracle.yml` + `tools/oracle-run.sh` with the per-corpus health gate; delete `oracle-rust.yml`/`oracle-kernel.yml`/`tools/{rust-scip,kernel-c}-oracle.sh`
- [ ] **C5** feat(ci): `tools/oracle-report-md.py` glue (JSON→markdown, Δ-vs-baseline, clean baseline-missing degradation) → job summary + fork-safe PR sticky comment
- [ ] **C-python** test(oracle): small Python corpus profile (package-dir scope, full venv) + validate scip-python resolution

[...]

PR #165, opened 2026-06-15
chore(ci): workflow hygiene — naming, least-privilege permissions, action bumps (A1)

skakri:
First step of #164 (A1 — independent workflow reorg).

## What
- **Hyphenate the bench workflow filenames** (`bench_pr_run` → `bench-pr-run`, etc.) so the whole `.github/workflows/` set is consistent. The `workflow_run` link resolves by workflow *name* (unchanged), so the rename is safe; `docs/bencher.md` updated to match.
- **Least-privilege `permissions:` on every workflow** — top-level `contents: read`, widened per-workflow only where needed (`checks: write` for the Bencher uploaders, `pull-requests: write` for the PR-comment tracker, `release-plz` keeps its `{}` + per-job scopes). Clears the CodeQL "workflow does not contain permissions" findings on `ci.yml` and `bench-pr-run`.
- **Bump the Node 20-deprecated actions** named in the runner warning: `actions/checkout` v4→v5 and `actions/cache` v4→v5 (both Node 24). Other actions left at their current majors to avoid unrelated breaking-change risk.

## Not in scope
The oracle consolidation, Python support, and the small-corpus runner are later items in #164 — this PR is just the independent hygiene pass.

All nine workflows parse; permissions verified least-privilege.

### DIFF
--- a/.github/workflows/bench-pr-run.yml
+++ b/.github/workflows/bench-pr-run.yml
@@ -2,31 +2,35 @@ name: bench-pr-run
 
 # Fork-safe step 1 of 2: runs benchmarks on a PR (forks included) with NO secrets, then uploads the
 # raw results + PR event as an artifact. The upload to Bencher (which needs the token) happens in
-# bench_pr_track.yml, triggered by this workflow completing — so a fork PR can never read the token.
+# bench-pr-track.yml, triggered by this workflow completing — so a fork PR can never read the token.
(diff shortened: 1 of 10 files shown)


skakri:
Good catch — fixed in cb74a22. `bencher run --github-actions` posts a Check *and* a comment, and since a permissions map zeroes omitted scopes, the tracker (contents:read + pull-requests:write only) had no Checks API access. Added `checks: write` to match Bencher's fork-PR example.

PR #166, opened 2026-06-15
feat(oracle): resolution-report + corpus-profile schema contract (C0)

skakri:
C0 of #164 — the stabilizer that goes in before the live computation/CLI (C2).

## Why first
A Δ-vs-main comparison is only meaningful when both sides share the same corpus, tool version, and report semantics. Locking the schema first means C2 implements an agreed contract rather than designing + coding in one step.

## What (all in `crates/rag-rat-core/src/index/oracle/report.rs`, no SQL/I/O)
- **`REPORT_SCHEMA_VERSION`** — bump on any shape/semantics change; consumers refuse to diff across a mismatch.
- **`CorpusProfile` (+ `CorpusHealth`) + `CorpusProfile::hash()`** — a stable sha256 over the whole profile (id/repo/rev/tool/prepare/bindings/health). It's the comparability key; a **golden test** pins it so an accidental profile edit surfaces in review.
- **`OracleResolutionReport` + `RunProvenance`** — the versioned, provenance-stamped before/after envelope (every field from the agreed list: `report_schema_version`, `corpus_profile_hash`, `corpus_id`, `tier`, `repo`, `rev`, `tool`, `tool_version`, `rag_rat_commit`, `worktree_id`, `production_sha`, `skipped_drifted`) + the resolution delta, verdict counts, monikers, and `OracleEvalMetrics`.
- **`ResolutionDelta` / `ResolutionBefore`** with documented denominators (resolved rate over total edges carrying a callee range).
- **`assemble()`** — pure mapping from a run's `OracleReport` + `OracleEvalMetrics` + before-counts; **`comparable_to()`** gates a Δ on schema version + profile hash.

## Tests
Golden profile hash; hash-changes-on-any-field; per-verdict-transition mapping (unresolved→upgraded / →resolved_external, exact→confirmed / →contradicted, oracle-only calls, skipped_drifted, monikers); envelope serialization (all required keys present); comparability rule.

## Not in scope (lands in C2)
The live index reads that produce the heuristic "before" counts + the moniker tally, the `rag-rat oracle report` CLI surface, and the BMF derivation — all implemented against this locked contract. The underlying `edge_oracle` accounting these consume is already fixture-tested in `oracle/tests.rs`.

Rust: no-default + default clippy clean, nightly fmt clean, 114 oracle tests + 5 new report tests green.

### DIFF
--- a/crates/rag-rat-core/src/index/oracle/mod.rs
+++ b/crates/rag-rat-core/src/index/oracle/mod.rs
@@ -21,6 +21,7 @@
 mod auto_run;
 mod join;
 mod manifest;
+mod report;
 mod run;
 mod scip;
 mod status;
@@ -34,6 +35,10 @@ use std::path::Path;
 pub use auto_run::{AutoRunDecision, AutoRunInputs, auto_run_decision};
 pub(crate) use join::package_of;
 pub use manifest::{ToolAvailability, ToolManifest};
+pub use report::{
+    CorpusHealth, CorpusProfile, OracleResolutionReport, REPORT_SCHEMA_VERSION, ResolutionBefore,
+    ResolutionDelta, RunProvenance,
+};
 use run::OracleRunInput;
 pub use run::{OracleEvalMetrics, RecallCalls};
 use rusqlite::Connection;
--- a/crates/rag-rat-core/src/index/oracle/report.rs
+++ b/crates/rag-rat-core/src/index/oracle/report.rs
@@ -206,12 +206,16 @@ impl OracleResolutionReport {
         }
     }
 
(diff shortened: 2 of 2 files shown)


skakri:
Good catch (P1) — fixed in the latest push. `comparable_to` now also requires `tool_version` to match. Kept it as a direct field comparison rather than folding `tool_version` into `corpus_profile_hash`: the profile hash is *static config* identity (repo/rev/tool/prepare/bindings/health), while `tool_version` is run-time provenance — but you're right that a different indexer build changes the SCIP output, so a cross-version Δ would mis-attribute the tool's changes to rag-rat. Added a test asserting a differing `tool_version` is incomparable.

PR #167, opened 2026-06-15
feat(lang): Python language support (symbols, graph edges, embeddings)

### LINKED ISSUE #170: Drive low-signal-chunk detection from the AST, not string prefixes
\`is_low_signal_chunk\` (`crates/rag-rat-core/src/index/ai/policy.rs`) decides whether a chunk is "pure plumbing" (imports/comments/braces) and should skip embedding. It currently classifies **lines by string prefix**, per-language (rust `use`/`mod`, ts `import`/`export`, kotlin `import`/`package`, python `import`/`from`, c/cpp `#include`; comment fences per-language).

This is a heuristic and somewhat brittle: it misses multi-line imports (`from x import (\n a,\n b,\n)`), spacing variants (`#  include`), `pub(crate) use`, etc.

**Fix:** drive low-signal from the tree-sitter parse we already do once per file — classify by **node kind** (e.g. `import_statement`/`import_from_statement`/`comment` for Python, `use_declaration`/`line_comment` for Rust, `preproc_include` for C) rather than text prefixes. Robust to spacing/multi-line; still a per-language enumeration, but of node kinds.

**Why not now:** the policy currently only receives chunk *text*, not its nodes, so this touches the chunk→parse seam — a real refactor that deserves its own review. The string-prefix heuristic is tested and fixes the concrete bugs (notably C `#define` macros are NOT low-signal) in the meantime.

No external crate fits — "is this line plumbing" is inherently parser-work, and tree-sitter (already a dep) is the right tool.

Context: Codex review on #167; design discussion confirmed keeping the heuristic + tracking the AST version here.

skakri:
Adds **Python** as a fully indexed language (tree-sitter-python) and replaces the string-prefix low-signal-chunk heuristic with **AST node-kind classification** (the brittleness the Python policy work surfaced — closes #170).

## Language + parsing
- `Language::Python` (`.py`/`.pyi`) wired through every match arm + the grammar.
- Symbols: functions, classes, **decorated defs** (the `decorated_definition` owns the chunk span so decorators are embedded, while the signature is read from the inner `def`/`class`), module/class-level **SCREAMING_SNAKE constants** (scope-gated — no lowercase-local spam), and **PEP 695 `type` aliases**.

## Graph edges
- **Calls** — `obj.method()` records a NameOnly receiver hint (never faked exact); `handlers[key]()` is not mis-recorded as the index var.
- **Imports** — module + each imported name (incl. comma/parenthesized `import_list`), never the `as` alias; relative imports normalize to their tail.
- **Inheritance** — base classes emit Implements + ReferencesType, resolving a parameterized base (`Generic[T]`) to its head and skipping `metaclass=`/splat kwargs.
- **Type references** — one recursive walker covers plain, generic (`Box[Item]`), qualified (`typing.Optional[Api]`), PEP 604 unions (`A | B`), nested (`Optional[list[Api]]`), and `Callable` param lists.
- **Decorators** — bare (`@requires_auth`) and qualified (`@pytest.fixture`, keeping the receiver) emit a call edge; parenthesized go through the call path.

## Embeddings / indexing safety
- **`is_low_signal_chunk` is AST-driven** (#170): classifies by node kind, not line prefixes — low-signal iff every top-level statement is plumbing (comment / import-include-package / bare docstring / `pass`). Robust to multi-line imports, multi-line docstrings, `#`-vs-`#define`, spacing; a symbol chunk is a definition, so never low-signal.
- The index walker **`FLOOR_DIRS` floors `.venv`/`venv`/`site-packages`/`__pycache__`**, so a `python = ["."]` config never ingests installed deps.
- `init` skips Python virtualenv/dependency trees and won't promote one as the default target.

## Surface + tests
- `symbol_lookup` MCP description + `AGENTS.md` list Python.
- held-mini `src/Main.py` fixture (bound) + ~27 Python/policy tests, including the adversarial-review finds (unions, subscript callee, nested generics, alias-not-import, metaclass filtering, decorator signature/span split). Hybrid eval (embeds held-mini) verified — no baseline regression.

## Deferred to follow-ups (resolver/init subsystems, not extraction)
#172 (Implements resolution preference), #173 (init dir-selection cosmetics), #174 (import-alias scope).

Unblocks **B6** (scip-python oracle backend).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,7 +15,7 @@ Reach for these first:
 
 - **`semantic_search`** — "where is this concept implemented?" Returns current source chunks with
   inline graph (callers/callees), git, and GitHub papertrail, all validated against current source.
-- **`symbol_lookup`** — exact/fuzzy symbol resolution (Rust/TS/Kotlin/C/C++), with any bound
+- **`symbol_lookup`** — exact/fuzzy symbol resolution (Rust/TS/Kotlin/C/C++/Python), with any bound
   memories attached.
 - **`impact_surface`** — the coding preflight before editing a symbol: graph callers/callees,
   tests, git history, papertrail, and **repo memories** crossing the call path. Run it before
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3160,6 +3160,7 @@ dependencies = [
  "tree-sitter-c",
  "tree-sitter-cpp",
  "tree-sitter-kotlin-ng",
+ "tree-sitter-python",
  "tree-sitter-rust",
  "tree-sitter-typescript",
  "ureq 2.12.1",
@@ -4248,6 +4249,16 @@ name = "tree-sitter-language"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "009994f150cc0cd50ff54917d5bc8bffe8cad10ca10d81c34da2ec421ae61782"
+
+[[package]]
+name = "tree-sitter-python"
+version = "0.25.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6bf85fd39652e740bf60f46f4cda9492c3a9ad75880575bf14960f775cb74a1c"
+dependencies = [
+ "cc",
+ "tree-sitter-language",
+]
 
 [[package]]
 name = "tree-sitter-rust"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,5 +55,6 @@ tree-sitter = "=0.26.9"
 tree-sitter-c = "=0.24.2"
 tree-sitter-cpp = "=0.23.4"
 tree-sitter-kotlin = { package = "tree-sitter-kotlin-ng", version = "1.1.0" }
+tree-sitter-python = "=0.25.0"
 tree-sitter-rust = "=0.24.2"
 tree-sitter-typescript = "=0.23.2"
--- a/crates/rag-rat-cli/src/init/mod.rs
+++ b/crates/rag-rat-cli/src/init/mod.rs
@@ -29,6 +29,15 @@ const SKIPPED_DIRS: &[&str] = &[
     ".next",
     ".turbo",
     ".venv",
+    // Python virtualenv / dependency / cache trees — never project source. Skipping them at scan
+    // time means their `.py` files never become dir candidates, so the "no default → promote the
+    // largest" fallback can't select a `site-packages` tree (#167 review).
+    "venv",
+    "virtualenv",
+    "site-packages",
+    "__pycache__",
+    ".tox",
+    ".nox",
     "build",
     "dist",
     "node_modules",
--- a/crates/rag-rat-cli/src/init/scan.rs
+++ b/crates/rag-rat-cli/src/init/scan.rs
@@ -96,8 +96,15 @@ pub(crate) fn candidate_dirs(scan: &RepoScan, language: Language) -> Vec<DirCand
             default: default_dir(scan, language, path),
         })
         .collect::<Vec<_>>();
+    // When nothing is a natural default, promote the largest candidate — but NEVER a Python
+    // dependency tree (`env/`, `site-packages`, …): in a repo whose only `.py` files live under a
+    // virtualenv, promoting it would write a binding into installed deps. If every candidate is a
+    // dependency dir, leave none defaulted (no binding beats the wrong binding).
     if !candidates.iter().any(|candidate| candidate.default)
-        && let Some(best) = candidates.iter_mut().max_by_key(|candidate| candidate.count)
+        && let Some(best) = candidates
+            .iter_mut()
+            .filter(|candidate| !fallback_excluded(language, &candidate.path))
+            .max_by_key(|candidate| candidate.count)
     {
         best.default = true;
     }
@@ -127,9 +134,44 @@ pub(crate) fn default_dir(scan: &RepoScan, language: Language, path: &Path) -> b
                 || text == "include"
                 || text.ends_with("/include")
                 || directly_contains_source(scan, language, path),
+        // Python packages typically sit at the repo root, under `src/`, or as a dir named after the
+        // package that directly contains `.py` files — but NEVER a virtualenv / dependency tree
+        // (`.venv/…/site-packages`), which would pollute the index with the whole dependency set.
+        Language::Python =>
+            !is_python_dependency_dir(&text)
+                && (text == "src"
+                    || text.ends_with("/src")
+                    || directly_contains_source(scan, language, path)),
         Language::Markdown => text == "docs" || text == ".",
     }
 }
+
+/// `true` for a path under a Python virtualenv / dependency tree — these hold installed third-party
+/// `.py` files (`site-packages`) that must never be indexed as project source.
+fn is_python_dependency_dir(text: &str) -> bool {
+    text.split('/').any(|component| {
+        matches!(
+            component,
+            ".venv"
+                | "venv"
+                | "env"
+                | ".env"
+                | "virtualenv"
+                | "site-packages"
+                | "__pycache__"
+                | ".tox"
+                | ".nox"
+                | "node_modules"
+        )
+    })
+}
+/// Whether the no-default fallback must NOT promote this candidate: a Python dependency/virtualenv
+/// tree (`env`/`.env`/`venv`/`site-packages`/…). `is_python_dependency_dir` covers the names
+/// `SKIPPED_DIRS` doesn't skip during the walk (e.g. `env`, which is too generic to skip globally).
+fn fallback_excluded(language: Language, path: &Path) -> bool {
+    language == Language::Python && is_python_dependency_dir(&display_rel(path))
+}
+
 pub(crate) fn directly_contains_source(scan: &RepoScan, language: Language, path: &Path) -> bool {
     path != Path::new(".")
         && scan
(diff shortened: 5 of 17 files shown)


skakri:
Addressed all five P2s in a61c468:

1. **Generic base classes** — `class Repo(Generic[T])` / `Base[Item]`: the base is now taken from the `subscript`'s `value`, so Implements/ReferencesType point at `Generic`/`Base`, not `T`/`Item`. Test: `generic_base_resolves_to_base_not_type_arg`.
2. **Constants → module/class scope only** — `assignment` is gated by `python_assignment_is_const_scope` (walks ancestors; a `function_definition`/`lambda` boundary ⇒ local, not a const). A function-local `LOCAL_MAX = 5` is no longer a symbol (asserted in `extracts_python_symbols`).
3. **Decorated defs own the span** — added a `decorated_definition` arm that carries the symbol span (so `@classmethod`/`@property`/`@app.get(...)` lines are in the chunk) using the inner name; the inner `function_definition`/`class_definition` arms are guarded against emitting a duplicate. Test asserts the `from_url` span starts at the `@classmethod` line.
4. **`#`-comment filtering Python-only** — `is_low_signal_chunk` is now fully **per-language** (comment fences *and* structural prefixes). `#` is a comment only in Python; C/C++ structural prefixes are `#include` only (never `#define` — a macro is a symbol). Regression test `c_macro_definition_is_not_low_signal`.
5. **venv exclusion** — the init scanner skips `.venv`/`venv`/`site-packages`/`__pycache__`/`.tox`/`node_modules` for Python, so `init -y` never selects installed deps as source.

Also folded in the review-thread suggestions: replaced the `starts_with` OR-chain with `matches!` + per-language prefix slices.

**On the failing Bencher check:** `bench-pr-run` itself passed — it's a Bencher threshold alert against a drifted main baseline (criterion is noisy; index_time runs the cargo corpus, unaffected by Python). #169 adds `workflow_dispatch` to `bench.yml` so the thresholds can be reapplied on demand (`--thresholds-reset` on main); after that runs, this PR's bench compares against a fresh baseline.

skakri:
@codex review

skakri:
Second review round addressed in 56bf9a0:

**P2s fixed:**
- **Metaclass/splat kwargs** — the base loop skips `keyword_argument`/`list_splat`/`dictionary_splat`, so `class Model(Base, metaclass=Meta)` emits a base edge for `Base` only, never `Meta`. Test: `metaclass_keyword_argument_is_not_a_base_class`.
- **Generic annotation head** — `x: Box[Item]` parses as `type > generic_type > identifier` (not `subscript`); a shared `python_type_head` now unwraps `type`/`generic_type`/`subscript` and anchors the head, so `Box` is emitted (was invisible). Test: `generic_annotation_anchors_the_head_type`.
- **Decorated signature** — the `decorated_definition` still owns the chunk span (decorators stay in the chunk, per round 1), but the **signature** is read from a threaded inner-def node, so logical-symbol member hashing + memory anchoring key on `def from_url(...)`, not `@classmethod`. Test asserts the signature is the `def` line.
- **venv fallback** — added `venv`/`virtualenv`/`site-packages`/`__pycache__`/`.tox`/`.nox` to `SKIPPED_DIRS`, so dependency `.py` files never become dir candidates and the 'no default → promote largest' fallback can't select a `site-packages` tree (fixes the root cause, not just `default_dir`).

**P3s deferred (noted in commit):**
- Multi-line docstring-only chunks: the line-prefix filter only catches the `"""` delimiters, not the body — this is exactly the brittleness the AST-based rewrite in **#170** fixes (docstring = a string node), so I'm folding it there rather than adding stateful string parsing to the heuristic.
- Pattern-list unpacked constants (`FOO, BAR = …`): real but rare; the single-return `symbol_node` shape doesn't emit multiple symbols per node. Can follow up if you'd like it.

454 core tests green, clippy clean both shapes, nightly fmt clean.

skakri:
Third round addressed in 6b3e4fa:

- **Type aliases** — PEP 695 `type UserId = int` (`type_alias_statement`) is now a `type` symbol, like the Rust/TS/C++ aliases. Test: `extracts_python_type_alias`.
- **Bare decorators** — `@requires_auth`/`@pytest.fixture` (an `identifier`/`attribute`, not a `call`) emit a NameOnly `CallsName` edge; `@foo(...)` was already handled via the call arm. Test: `bare_decorator_emits_a_call_edge`.
- **Qualified generic annotations** — `x: typing.Optional[Api]` is a `subscript` whose args the recursion misses; now emits ReferencesType for both head (`Optional`) and arg (`Api`). (Non-qualified `Box[Item]` = `generic_type`, args already reached by recursion.) Test: `qualified_generic_annotation_emits_head_and_arg`.
- **init venv fallback** — the 'no default → promote largest candidate' fallback now skips Python dependency dirs (`env`/`.env`/`site-packages`/…), so a repo whose only `.py` files live under a virtualenv never gets a binding into installed deps (fixes the `env`-not-in-SKIPPED_DIRS gap at the promotion site).

**Deferred → #172:** Python `Implements` edges resolve via the `trait`/`interface` preference, not `class`. It's a cross-language resolver change that deserves its own review, and the base class is already captured via the `ReferencesType` edge we emit alongside — so impact is bounded. Filed as a follow-up.

457 core tests green, clippy clean both shapes, nightly fmt clean.

Meta note: we're into the long tail of Python's grammar (each round surfaces narrower cases). The core — symbols, calls, imports, inheritance, annotations, decorators, type aliases — is solid and tested. Happy to keep iterating, but might suggest capping further niche cases (walrus/match/comprehension edges, etc.) into a 'Python edge cases' follow-up so this lands and B6/the corpus runner can build on it.

skakri:
Round 4 addressed in fe14dea:

**The structural keystone:**
- **Index walker floors venv/site-packages** — `FLOOR_DIRS` (core walker, all configs) gains `.venv`/`venv`/`site-packages`/`__pycache__`. `site-packages` is the load-bearing one: installed deps live there regardless of the venv dir's name, so a `python = ["."]` binding — from init OR a hand-written config — never ingests the dependency set. This resolves the venv-indexing concern at the source (both #4 and the *danger* in #3).

**Bounded graph fixes:**
- **Multi-imports** — `from pkg import A, B` nests names under `import_list`; `python_import_target` recurses into it, so each name gets an Imports edge (was: only the module). Test: `multi_import_emits_each_imported_name`.
- **Qualified bare decorators** — `@pytest.fixture` now records its `pytest` receiver + dotted path (was `EdgeContext::default()`), so the resolver can't fall back to a local same-named `fixture`. Test: `qualified_bare_decorator_keeps_its_receiver`.

**Deferred → #173** (init-scanner Python dir-selection cosmetics, now that the walker floor removed the safety risk): root entrypoints omitted when a package dir exists; env-only repos getting a broad `.` binding (safe now — deps are floored — just broader than ideal).

459 core tests green, clippy clean both shapes, nightly fmt clean.

**Strong recommendation to land:** we're 4 rounds deep; this round's findings were the walker floor (now fixed, structural) + two narrow graph cases (fixed) + two init cosmetics (deferred, no longer a safety issue). The Python core — symbols, calls, imports (incl. lists/aliases/relative), inheritance (incl. generics/metaclass), annotations (incl. qualified generics), decorators (bare + qualified), constants, type aliases, low-signal, and the venv floor — is comprehensive and tested (~20 Python-specific tests). Further cases (walrus/match/comprehensions) would be genuinely niche. I'd merge this and let B6/the corpus runner build on it, with #170/#172/#173 tracking the long tail.

skakri:
Round 5:

- **Nested generic args** — fixed (a2be5bc). `typing.Optional[list[Api]]` is a subscript whose nested args aren't `type`-wrapped; `emit_python_subscript_type_arg_edges` now recurses, emitting ReferencesType for `Optional`, `list`, AND the nested `Api`. Test: `nested_qualified_generic_annotation_emits_all_heads`.
- **Import-alias resolution** — deferred → **#174**. `from models import User as Account` then `Account()`: the import edge already targets `User` correctly, but binding alias *uses* back to `User` needs a Python import-alias **scope** the resolver consults — the analog of the Rust per-module `use`-scope subsystem (#61), not an extraction tweak. Sibling of #172.

460 core tests green, clippy clean both shapes, fmt clean.

This round was 2 items (down from 5→5→2): one bounded fix (done) and one resolver subsystem (deferred). The extraction layer is now very thorough. Reiterating the recommendation to land #167 — the open follow-ups (#170 low-signal AST, #172 Implements resolution, #173 init dir heuristics, #174 alias scope) are all resolver/init refinements that are cleaner as their own PRs than as more rounds here.

skakri:
Ran **two adversarial review passes** over this branch (parser/edges + integration) to get ahead of further rounds. Confirmed findings fixed in f75e946:

- **PEP 604 unions** (was uncovered, high-frequency): `x: A | B` is a `binary_operator` — only the last operand was referenced. Replaced the head+subscript-args logic with one recursive `emit_python_type_refs` walker (type/generic_type/subscript/binary_operator/list/tuple), so unions, qualified/nested generics, and `Callable[[int, A], B]` param lists all emit a ReferencesType per type.
- **Subscript callee**: `handlers[key]()` recorded the index `key` as the callee — now suppressed.
- **Generic base classes** now reference their type args (`class C(Mapping[str, Api])` → Implements Mapping + ReferencesType Api).
- **Type-alias self-reference**: `type X = int` no longer emits a ReferencesType to the alias name being defined.
- **Docs**: `symbol_lookup` MCP description + AGENTS.md now list Python.

New tests for each (union, subscript-callee, base type-args, alias self-ref, init venv-not-promoted). 464 core + 62 cli tests green, clippy clean both shapes.

The reviewers also confirmed a lot CORRECT under skepticism: async def, decorated span/signature split, parenthesized-vs-bare decorators, relative/aliased/list imports, chained/scoped constants, metaclass/splat base filtering, string forward-refs, no panics (all node text goes through checked accessors), FLOOR_DIRS consulted by both walker and watcher, no non-exhaustive `match Language`, no `.pyi` dup-symbol issue, policy language-arm always hits.

Remaining deferrals are resolver/init subsystems tracked as #172 (Implements resolution), #174 (import-alias scope), #173 (init dir cosmetics) — cleaner as their own PRs. The extraction layer is now thorough (24 Python tests). Recommend landing.

skakri:
Intertwined **#170** here (per your call): replaced the string-prefix `is_low_signal_chunk` heuristic with **AST node-kind classification** (464295f). A chunk is low-signal iff every top-level statement is plumbing (comment / per-language import-include-package / bare docstring / pass); any definition/expression is signal, so a symbol chunk is never low-signal. This de-brittles exactly what the Python policy work kept tripping on — multi-line imports, multi-line docstrings (interior prose no longer leaks), `#`-vs-`#define`, spacing. Per-language node kinds (rust use_declaration; ts import_statement; kotlin import_header/package_header; c/cpp preproc_include — not preproc_def; python import/from/future/pass + docstring); parse-failure → keep the chunk. Closes #170.

Verified the **hybrid eval** (which embeds held-mini) still passes — no baseline regression from the behavior change. 467 core tests, clippy clean both shapes, fmt clean.

So #167 now also carries the proper AST low-signal (no more deferred heuristic). Remaining deferrals are genuinely resolver/init subsystems: #172 (Implements resolution), #174 (import-alias scope), #173 (init dir cosmetics).

PR #168, opened 2026-06-15
feat(oracle): live before/after resolution report computation (C2 core)

skakri:
C2 of #164 — the live computation behind the C0 contract (#166, merged). Builds on the locked `OracleResolutionReport` schema.

## What
- **`store::resolution_before_counts`** — the heuristic 'before' counts over the active checkout's edge candidates: total (callee range present), `resolved_in_corpus` (Exact/Syntactic + `to_symbol_id`, matching the join's `heuristic_resolved_in_corpus` so before/after are consistent), unresolved (NameOnly/Ambiguous). One scoped query.
- **`store::count_symbols_with_moniker`** — per-tool `logical_symbol_monikers` tally (the 'symbols enriched' signal; re-run-safe, unlike `monikers_written`).
- **`oracle::resolution_report` + `IndexDatabase::resolution_report`** — read the before counts + moniker tally, diff the eval metrics from `edge_oracle`, and assemble the report. `tool_version` is sourced once from `provenance` and used for both the envelope and the metric scope key.
- **Fixture-DB test** over a synthetic edges + edge_oracle + monikers DB asserting the full path: before/after counts, every verdict transition, moniker tally, `skipped_drifted`, and precision/recall.
- Tightened the C0 `ResolutionDelta` doc to the actual definition (Exact/Syntactic + in-corpus, not 'same-file-name').

## Not in scope (remaining C2 slice)
The `rag-rat oracle report` **CLI surface** + **BMF derivation** — gated on **C1**'s corpus-profile file, since the report needs a `CorpusProfile` to stamp. `IndexDatabase::resolution_report` is the API the CLI/runner will call; the corpus file feeds the profile.

445 core tests green, clippy clean in both feature shapes, nightly fmt clean.

### DIFF
--- a/crates/rag-rat-core/src/index/oracle/mod.rs
+++ b/crates/rag-rat-core/src/index/oracle/mod.rs
@@ -146,6 +146,39 @@ pub fn oracle_eval_metrics(
     run::eval_metrics(conn, tool, tool_version, commit_sha, worktree_id, recall_calls)
 }
 
+/// Assemble the typed before/after [`OracleResolutionReport`] (C2) for a just-completed run: read
+/// the heuristic "before" resolution counts + the moniker tally from the index, compute the diffed
+/// eval metrics, and map everything onto the C0 schema. `run` is the run's [`OracleReport`] (its
+/// run-only counts — `skipped_drifted`, the recall call pair — aren't reconstructable from the side
+/// tables). The SCIP `tool_version` is read from `provenance` (the single source for both the
+/// report envelope and the metric scope key). Pure data: Markdown rendering is a glue concern, not
+/// here.
+pub fn resolution_report(
+    conn: &Connection,
+    profile: &report::CorpusProfile,
+    provenance: &report::RunProvenance,
+    tool: OracleTool,
+    commit_sha: &str,
+    worktree_id: &str,
+    run: &OracleReport,
+) -> anyhow::Result<report::OracleResolutionReport> {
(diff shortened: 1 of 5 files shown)


skakri:
All three P2s fixed in 69ecf34:

1. **Moniker count scoped to tool_version** — `count_symbols_with_moniker(conn, tool, tool_version)` now filters `tool_version` too, so the count describes the same run as the `edge_oracle`-derived metrics (passed `provenance.tool_version`).
2. **Match the oracle candidate filter** — `resolution_before_counts` now also requires `callee_end_byte IS NOT NULL`, matching `edge_join_candidates` exactly, so `total_edges`/`unresolved_before` can't diverge from `OracleReport.edges_examined`.
3. **Schema bump** — `REPORT_SCHEMA_VERSION` 1 → 2, since `resolved_before`/`resolved_after` semantics changed to the join's `heuristic_resolved_in_corpus` (Exact/Syntactic + in-corpus). Bumping follows the contract's own rule; safe here since C2 is the first producer (C0 had no live computation), but the version guards any future comparison.

On `codecov/patch`: the new logic is covered by the fixture-DB test; the one uncovered line is the thin `IndexDatabase::resolution_report` wrapper (3 lines passing active checkout ids). Happy to add a wrapper test if you want it gated.

445 core tests green, clippy clean, nightly fmt clean.

skakri:
@codex review

PR #169, opened 2026-06-15
chore(ci): workflow_dispatch on bench.yml to reapply Bencher thresholds on demand

skakri:
The bench job reapplies the Bencher thresholds (`bencher run --thresholds-reset` on `main`) **only on push to main**, so a drifted/bad threshold baseline (which makes PR bench checks fail spuriously — e.g. #167) can't be reapplied without a code push.

Add `workflow_dispatch` so thresholds can be reapplied on demand: **Actions → bench → Run workflow** (on `main`) re-establishes the baseline that PR runs clone via `--start-point-clone-thresholds`.

No behavior change to the push path.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -9,6 +9,10 @@ name: bench
 on:
   push:
     branches: [main]
+  # Manual trigger to REAPPLY the Bencher thresholds (this job runs `bencher run --thresholds-reset`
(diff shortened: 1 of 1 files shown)


skakri:
Good catch — fixed. Added `if: github.ref == 'refs/heads/main'` to the bench job so a `workflow_dispatch` from a feature ref is skipped (the publish steps hardcode `--branch main --thresholds-reset`). Dispatch from main works as intended; any other ref is a no-op.

PR #171, opened 2026-06-15
feat(oracle): corpus profiles + health-gate loader (C1)

skakri:
C1 of #164 — the declarative corpus inputs + health gate the runner (C3) and `oracle report` CLI consume. Builds on the merged C0/C2 report contract.

## What
- **`tools/oracle-corpora.toml`** — 5 corpora: **small** tier (semver / cJSON / requests — per-PR) + **heavy** tier (cargo / kernel — release→Bencher), each pinning `repo`/`rev`/`tool`/`prepare`/`bindings`/`health`.
- **`oracle/corpus.rs`** — `load_corpora` (TOML), `corpus_by_id`, `corpora_for_tier`, and **`check_corpus_health`**: the fail-on-nonsense gate (min heuristic edges / min oracle verdicts / max `skipped_drifted` / min `symbols_with_moniker`) returning typed `HealthViolation`s, so the runner exits non-zero on a bogus run even when the oracle command exits 0.
- `CorpusProfile`/`CorpusHealth` gain `Deserialize` (TOML); report contract otherwise unchanged.

## Tests
Committed corpora load with the expected ids/tiers (+ `py-requests` → `scip-python` tool/binding); **golden per-profile hashes** (the review tripwire for an accidental corpus/threshold edit — an edit changes the hash and makes prior reports incomparable); health-gate healthy + each-threshold-violation cases.

## Notes
- `rev`s are release tags as a first cut; pin to the tag's commit SHA before trusting the Bencher series. The golden hash test surfaces any such edit.
- `py-requests` uses `tool = "scip-python"` as data; the backend lands in **B6** (gated on #167). The loader doesn't validate the tool against `OracleTool`, so C1 is independent of B6.

449 core tests green, clippy clean both shapes, nightly fmt clean.

### DIFF
--- a/crates/rag-rat-core/src/index/oracle/corpus.rs
+++ b/crates/rag-rat-core/src/index/oracle/corpus.rs
@@ -17,9 +17,17 @@ struct CorpusFile {
     corpus: Vec<CorpusProfile>,
 }
 
-/// Parse the corpus profiles from an `oracle-corpora.toml` string.
+/// Parse the corpus profiles from an `oracle-corpora.toml` string. Fails closed on an empty result:
+/// a typo'd table name (`[[corpora]]` instead of `[[corpus]]`) deserializes to zero corpora, which
+/// would silently turn the CI matrix into a no-op and skip the health gate — so require at least
+/// one.
 pub fn load_corpora(toml_str: &str) -> anyhow::Result<Vec<CorpusProfile>> {
-    Ok(toml::from_str::<CorpusFile>(toml_str)?.corpus)
+    let corpora = toml::from_str::<CorpusFile>(toml_str)?.corpus;
+    anyhow::ensure!(
+        !corpora.is_empty(),
+        "oracle-corpora.toml has no [[corpus]] entries (wrong table name?)"
+    );
+    Ok(corpora)
 }
 
 /// The corpus with this id, if any.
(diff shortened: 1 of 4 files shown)


skakri:
Addressed in 7af62a2:

1. **Packaging** — dropped the out-of-package `include_str!` (the file lives in repo `tools/`, outside the `rag-rat-core` package, so packaged-crate tests couldn't compile). Logic/health tests now use an in-crate inline `SAMPLE`; a new `committed_corpora_file` test validates the real `tools/oracle-corpora.toml` at **runtime via `std::fs`**, skipping when absent (packaged crate) — so the golden per-profile-hash tripwire still runs in dev/CI without breaking `cargo package`/`publish`.
2. **Fail closed on empty** — `load_corpora` now errors when zero `[[corpus]]` entries parse (a typo'd `[[corpora]]` would otherwise yield a silent no-op matrix that skips the health gate). Test: `load_rejects_empty_or_misnamed_corpus_array`.
3. **Kernel build** — `linux-kernel` prepare is now `make defconfig` → **`make -j`** → `make compile_commands.json`, mirroring `tools/kernel-c-oracle.sh`; the compdb derives from per-object `.cmd` files, so without the build scip-clang gets an empty database. (Golden kernel hash updated accordingly.)

4. **py-requests in small tier** — this is the intended end state, and C1 is **data-only**: nothing in main consumes `load_corpora` yet (the C3 runner that materializes `python` bindings into `rag-rat.toml` isn't built). By the time C3 runs the small matrix, #167 (adds `python` to `Language::from_str`) + B6 (scip-python backend) will have landed, and the runner will gate corpora on tool/lang availability (it already has the `Blocked`-on-missing-tool pattern). So there's no live break; I'll make C3 skip unsupported-tool corpora explicitly. Happy to temporarily flag it if you'd prefer, but that'd churn the golden hash twice.

450 core tests green, clippy clean both shapes, nightly fmt clean.

PR #175, opened 2026-06-15
feat(oracle): `oracle report --corpus <id>` — run a corpus + emit its C2 resolution report (C2-CLI)

skakri:
Part of the multi-language SCIP-oracle runner epic (#164). Adds the **C2-CLI** surface over the C0/C2 report contract that landed in #166/#168 and the corpus profiles/health-gate from #171.

## What

`rag-rat oracle report --corpus <id> [--corpora <path>] [--scip <path>]`:

1. Loads the corpus profile from `tools/oracle-corpora.toml` (override with `--corpora`).
2. Maps the corpus's declared `tool` to an oracle backend.
3. Runs the oracle — produces a `.scip` with the corpus's tool, or consumes a pre-built `--scip` (deterministic; no tool needed).
4. Assembles the typed `OracleResolutionReport` (before/after edge resolution, verdict transitions, metrics, schema- + profile-stamped) and emits it as JSON/TOON.
5. Applies the per-corpus **health gate**: a violated threshold exits non-zero even when the oracle command itself succeeded — catching "scip emitted almost nothing" / "venv didn't resolve deps" / a broken parse.

## Decisions

- **Report printed unconditionally, before the gate.** A Δ/glue script consumes the JSON even for a failing run; violations go to stderr, then the command exits non-zero.
- **Missing tool is a hard error here** — deliberately the opposite of `oracle run`'s Blocked+exit-0 UX. This is a measurement runner over a corpus whose tool CI is expected to have installed; a silent skip would pass green on a broken environment.
- **run + `resolution_report` under one write lock**; `.scip` production stays outside it (the #82 P3 lock-free-production posture).
- **`rag_rat_commit` provenance** reads `$RAG_RAT_COMMIT` (CI's git SHA), falling back to the crate version off CI.

## Tests / verification

- New parse test `oracle_report_requires_corpus_and_takes_optional_paths` (mandatory `--corpus`, optional `--corpora`/`--scip`).
- Existing `corpus.rs` health-gate unit tests cover the threshold logic.
- Manually verified end-to-end against this repo's own index with rust-analyzer: a real report (42k edges, resolved 9767→41365 after the oracle), the failing-gate path (exit 1, violation on stderr, report still on stdout), the passing-gate path (exit 0), and the `--scip` prebuilt path.
- `cargo +nightly fmt --check`, `cargo clippy --all-targets`, and the full cli + core-oracle suites are green.

## Follow-up

`oracle report --corpus py-requests` errors with "unknown oracle tool" until **B6** (the scip-python backend) lands — `OracleTool::from_db_str` doesn't yet know `scip-python`.

### DIFF
--- a/crates/rag-rat-cli/src/cli.rs
+++ b/crates/rag-rat-cli/src/cli.rs
@@ -259,6 +259,9 @@ pub(crate) enum OracleCommand {
     Run(OracleRunArgs),
     /// Report oracle verdict counts + whether the indexer tool is installed.
     Status(OracleStatusArgs),
+    /// Run the oracle for a declared corpus and emit its typed before/after resolution report
+    /// (C2). Applies the corpus health gate: exits non-zero if the run falls outside thresholds.
+    Report(OracleReportArgs),
 }
 
 #[derive(Debug, Args)]
@@ -277,6 +280,20 @@ pub(crate) struct OracleStatusArgs {
     /// Report on one oracle tool only (default: every known tool).
     #[arg(long, value_enum)]
     pub tool: Option<OracleToolArg>,
+}
+
+#[derive(Debug, Args)]
+pub(crate) struct OracleReportArgs {
+    /// The corpus id to report on (must match a `[[corpus]]` entry's `corpus_id`).
+    #[arg(long)]
+    pub corpus: String,
+    /// Path to the corpus profiles file. Defaults to `<root>/tools/oracle-corpora.toml`.
+    #[arg(long)]
+    pub corpora: Option<PathBuf>,
+    /// Consume a pre-built `.scip` instead of invoking the corpus's tool. Deterministic; the tool
+    /// need not be installed.
+    #[arg(long)]
+    pub scip: Option<PathBuf>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, ValueEnum)]
@@ -548,5 +565,34 @@ mod tests {
             cli.command,
             Command::Oracle(OracleArgs { command: OracleCommand::Status(_) })
         ));
+    }
+
+    #[test]
+    fn oracle_report_requires_corpus_and_takes_optional_paths() {
+        // `--corpus` is mandatory; a bare `oracle report` must not parse.
+        assert!(Cli::try_parse_from(["rag-rat", "oracle", "report"]).is_err());
+        let cli = Cli::try_parse_from([
+            "rag-rat",
+            "oracle",
+            "report",
+            "--corpus",
+            "py-requests",
+            "--corpora",
+            "/tmp/corpora.toml",
+            "--scip",
+            "/tmp/x.scip",
+        ])
+        .expect("parse");
+        match cli.command {
+            Command::Oracle(OracleArgs { command: OracleCommand::Report(args) }) => {
+                assert_eq!(args.corpus, "py-requests");
+                assert_eq!(
+                    args.corpora.as_deref(),
+                    Some(std::path::Path::new("/tmp/corpora.toml"))
+                );
+                assert_eq!(args.scip.as_deref(), Some(std::path::Path::new("/tmp/x.scip")));
+            },
+            other => panic!("expected oracle report, got {other:?}"),
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


skakri:
Addressed in 3215f32.

**P2 — Don't publish unhealthy oracle verdicts** ✅ Fixed. `oracle report` now rolls the run back when the health gate fails: new `oracle::rollback_run` (+ `store::delete_oracle_run`, `IndexDatabase::rollback_oracle_run`) deletes the run's `edge_oracle` verdicts, `logical_symbol_monikers`, and `oracle_runs` row, run **inside the same write lock** as the run that wrote them (so it's atomic, in `finalize_corpus_report`). Verified end-to-end: after a failed gate the run exits non-zero, the report JSON is still emitted, and `edge_oracle`/`oracle_runs`/monikers are all empty with `oracle status` reporting no verdicts — so `latest_run_tool_version` can no longer select it.

**P2 — Validate the checkout before stamping the corpus profile** ✅ Fixed (bindings). New `ensure_checkout_matches_corpus` fails closed unless the active checkout's target bindings (language → directories) exactly equal the corpus profile's `bindings`, so the report can't stamp corpus X's `corpus_profile_hash` onto a different population. Added a unit test. On repo/rev specifically: an in-command check isn't tractable (the profile's `rev` is a tag, the checkout exposes a detached SHA — no remote/tag resolution available in-process), and the runner *guarantees* it by cloning `repo@rev`; the bindings are the part that independently changes the measured population, so that's what's validated here.

**P2 — Honor corpus timeout in report runs** — This is handled one layer up by design, not dropped. `timeout_minutes` is documented on `CorpusHealth` as "the corpus wall-clock budget the **runner** wraps the run in" (and `check_corpus_health` explicitly does not enforce it). The C3 runner (`tools/oracle-run.sh`, #177) wraps the whole `oracle report` invocation in `timeout ${timeout_minutes}m`, so a hung `rust-analyzer`/`scip-clang` fails the corpus (exit 124) rather than hanging CI. Enforcing it a second time inside the command — around only production+report, not the index step — would be redundant and muddy the orchestration/measurement split. Happy to add an in-command deadline too if you'd prefer defense-in-depth.

skakri:
**P2 — Reject custom target filters before stamping corpus reports** ✅ Fixed in 0940023. `ensure_checkout_matches_corpus` now requires each target to carry the simple `[target_bindings]` form's default filters (`include = ["**/*.<ext>"]`, empty `exclude`) in addition to the language→dir set match — so a `[[target]]` that keeps the same language+dirs but adds custom `include`/`exclude` (which the indexer applies, changing the file population) fails closed instead of stamping the corpus's `corpus_profile_hash` onto a filtered subset/superset. Unit test extended with a custom-`exclude` case and a narrowed-`include` case.

skakri:
Round 2 — addressed in d60a384.

**P2 — Preserve last good monikers / P2 — Keep previous same-version runs on failed reports** ✅ Both fixed by the same change: the report path is now **provisional**, not commit-then-rollback. You were right that a post-hoc delete can't restore anything — `run::run`'s authoritative clear destroys the prior `(tool, tool_version)` verdicts and the tool's monikers at the run's *start*, before the gate runs.

`run::run` is split into `run()` (the commit-on-success wrapper, unchanged for existing callers) and `run_in_tx()` (the body). New `oracle::run_oracle_report` runs `run_in_tx` + report assembly + the health gate inside **one transaction** and commits **only if healthy**; an unhealthy run drops the transaction, rolling back the whole pass — including the authoritative clear — so the previous healthy run's verdicts, monikers, and `oracle_runs` row are fully preserved (same version or not). Removed `rollback_run` / `delete_oracle_run` / `rollback_oracle_run` / `finalize_corpus_report`.

Verified e2e: a healthy run, then a gate-failing run on the same checkout (same tool version) — `oracle status` still reports the healthy run's verdicts afterward (`edge_oracle`/`oracle_runs`/monikers all intact).

**P2 — Support every committed corpus tool (`py-requests` → `scip-python`)** — This is a cross-PR sequencing point, not a defect in this PR: `OracleTool::from_db_str` learns `scip-python` in #176 (the B6 backend). #175 + #176 (+ #177's runner) land together; until then the unknown-tool path is a clean fail-closed error, not a crash. The runner's small-tier matrix only goes green for `py-requests` once #176 is merged — called out in #177's description. I'd rather not duplicate the B6 registration into #175.

PR #176, opened 2026-06-15
feat(oracle): scip-python backend — Python compiler-grade resolution (B6)

skakri:
Part of the multi-language SCIP-oracle runner epic (#164). Adds the **B6** scip-python backend so the SCIP oracle resolves Python edges to the `Compiler` tier — unblocking `oracle report --corpus py-requests` (the Python corpus was already declared in `tools/oracle-corpora.toml` from #171, and #175 added the report command; this makes its tool runnable).

## What

`OracleTool::ScipPython` added to the oracle tool registry:

- **Manifest entry**: program `scip-python`, language `python`, install hint.
- **Capability probe**: `scip-python index --help` exits 0 (the analog of rust-analyzer's `scip --help` check); `--version` reports the tool version.
- **Invocation**: `scip-python index --project-name <root-basename> --cwd <root> --output <abs>`. The project name gives in-corpus symbols a non-empty moniker package (what lets `count_symbols_with_moniker` see them); `--cwd` is where scip-python resolves the project + its installed deps.
- **No `compile_commands.json`-style prerequisite**: installing deps is the corpus `prepare` step's job, and a failed environment shows up as a near-zero moniker count the report health gate already catches — so `prerequisite_blocked` is `None`.
- CLI `--tool scip-python` + exhaustive enum coverage; argv unit test.

The SCIP occurrence→edge join is **language-agnostic** (keyed on document paths + positions), so no join changes were needed.

## Operational notes (documented in `docs/oracle.md`)

- scip-python derives `--project-version` from the git revision, so it expects a **git checkout** — real corpora are cloned at a pinned rev, so this is fine; a non-git dir makes it crash on an undefined version.
- Imports resolve against **installed** dependencies (a venv), so cross-package edges need the deps importable.

## Verification

- New manifest argv test `scip_python_indexes_a_cwd_with_a_project_name`; the existing exhaustive `every_tool_has_a_manifest_entry` now covers the third variant.
- Live probe: `oracle status --tool scip-python` → `available 0.6.6`.
- End-to-end on a small git-initialized Python project (3 files, 8 edges): **5 edges examined → 4 confirmed, 1 resolved-external, 4 monikers written**, verdicts persisted.
- `cargo +nightly fmt --check`, `cargo clippy --all-targets`, and the oracle core + cli suites are green.

## Stacking

Independent of #175 (C2-CLI) at the code level — both branch off main and touch disjoint code. Either can merge first; once both land, `oracle report --corpus py-requests` is fully runnable.

### DIFF
--- a/crates/rag-rat-cli/src/cli.rs
+++ b/crates/rag-rat-cli/src/cli.rs
@@ -285,6 +285,8 @@ pub(crate) enum OracleToolArg {
     RustAnalyzer,
     #[value(name = "scip-clang")]
     ScipClang,
+    #[value(name = "scip-python")]
+    ScipPython,
 }
 
 impl OracleToolArg {
@@ -292,6 +294,7 @@ impl OracleToolArg {
         match self {
             OracleToolArg::RustAnalyzer => rag_rat_core::index::oracle::OracleTool::RustAnalyzer,
             OracleToolArg::ScipClang => rag_rat_core::index::oracle::OracleTool::ScipClang,
+            OracleToolArg::ScipPython => rag_rat_core::index::oracle::OracleTool::ScipPython,
         }
     }
 }
--- a/crates/rag-rat-cli/src/main.rs
+++ b/crates/rag-rat-cli/src/main.rs
@@ -199,7 +199,20 @@ fn spawn_detached_oracle_auto_run(config: &rag_rat_core::Config) {
                 None => return Ok(()),
             }
         };
+        // The languages this checkout actually indexes. Gating background runs to these (#176)
+        // stops the auto-run loop from invoking a backend whose language isn't present —
+        // e.g. scip-python installed but no Python target: it would index nothing, fail,
+        // the error would be swallowed with no `oracle_runs` row recorded, and the loop
+        // would retry the doomed run every poll.
+        let configured_languages: std::collections::HashSet<&str> =
+            config.targets.iter().map(|target| target.language.as_str()).collect();
         for &tool in OracleTool::ALL {
+            // Skip a backend whose language this checkout doesn't index — never auto-run it here
+            // (the status registry stays broad; only background runs are gated).
+            let manifest = oracle::ToolManifest::for_tool(tool);
+            if !manifest.languages.iter().any(|lang| configured_languages.contains(lang)) {
+                continue;
+            }
             // Cheap probe before any decision: an uninstalled tool can never run, so don't even
             // read its run history.
             if matches!(oracle::probe_oracle_tool(tool), oracle::ToolAvailability::Blocked { .. }) {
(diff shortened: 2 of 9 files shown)


skakri:
Addressed in 84cc401.

**P2 — Route scip-python through graph surfacing** ✅ Fixed (and it covered the pre-existing scip-clang gap too). `enrich_hops_with_oracle` and `compare_graph_to_scip` no longer hardcode `OracleTool::RustAnalyzer` — new `oracle::latest_runs_in_scope` enumerates every backend with a run in the checkout, `enrich_hops_with_oracle` merges their verdicts, and `compare_graph_to_scip` aggregates across them (verdict sets are disjoint by edge language, so merging can't collide; the report's `tool`/`tool_version` now name all contributing backends). Added a regression test (`compare_graph_to_scip_surfaces_non_rust_analyzer_tools`) asserting a scip-clang verdict surfaces and the report names the tool.

**P2 — Pin scip-python project version** ✅ Fixed. `scip_command` now passes `--project-version _`, so the SCIP symbol strings (and the monikers persisted from them) stay stable across commits, preserving moniker-anchored memory relocation. Bonus: this also fixes scip-python's crash on a non-git checkout (the git-rev default was undefined there) — verified end-to-end on a non-git Python project (4 edges examined, 3 confirmed, 3 monikers, exit 0). Manifest argv test + docs updated.

**P2 — Filter auto-run tools by configured language** ✅ Fixed. The MCP background auto-run loop now skips any backend whose languages aren't in `config.targets` before probing/deciding — so scip-python installed without a Python target no longer spawns a doomed indexer (and silently retries) every poll. The status registry (`oracle status`, which lists all tools) stays broad; only background runs are gated.

`cargo +nightly fmt --check`, `clippy --all-targets`, and the cli + core-oracle suites (incl. the new surfacing test) are green.

skakri:
**P2 — Count Python constructor references in recall** ✅ Addressed in 40d12d2 — by making recall's two sides use the same callable population, which fixes the inflation you identified.

Root cause confirmed: `covered_calls` counted a `calls_name` edge matched to *any* SCIP occurrence, while `count_uncovered_calls` (the denominator) requires `scip::symbol_is_callable` (`).`). scip-python records a constructor call (`Foo()`, which our extractor emits as `CallsName`) as a reference to the **class** symbol (`Foo#`), so a *covered* constructor counted while a *missed* one — invisible to the callable-filtered uncovered side — never offset it → inflated recall.

The covered side now also requires `symbol_is_callable`, so both sides measure the same population (`).` method/function calls). I went with *exclude-on-both-sides* rather than *include constructors* because accurate inclusion isn't possible from SCIP alone: a `Foo#` occurrence is identical whether it's a constructor **call** (`Foo()`) or a type **annotation** (`x: Foo`), so counting `#` as callable on the uncovered side would over-count type refs as missed calls and *deflate* recall. Confining recall to `).` callables is the well-defined, symmetric metric; constructor-call recall would need a call-role SCIP doesn't carry. Rust/C are unaffected (their `calls_name` edges already match `).` symbols). Regression test `covered_side_requires_a_callable_scip_symbol` added.

PR #177, opened 2026-06-15
feat(oracle): unified tier-driven corpus runner + oracle.yml (C3)

skakri:
Part of the multi-language SCIP-oracle runner epic (#164). Replaces the two per-language demo workflows with **one declarative, tier-driven runner** over `tools/oracle-corpora.toml` — the C3 piece that ties the corpus profiles (#171), the report command (#175), and the scip-python backend (#176) into a CI-runnable whole.

## What

- **`tools/oracle-run.sh`** — run the oracle for **one** corpus end to end: read its profile → shallow-clone the repo at its pinned rev → run its `prepare` steps → render the checkout's `rag-rat.toml` from its `bindings` → `rag-rat index --full` → `rag-rat oracle report`. The health gate lives in `oracle report`; the script propagates its exit code (non-zero on a violation) and **always** leaves the report JSON behind so a Δ glue script can consume even a failing run.
- **`tools/oracle-corpus.py`** — stdlib (`tomllib`) reader the bash runner shells out to for the pre-index fields (repo/rev/prepare/bindings) and `--list-tier` for the CI matrix. `oracle report` reads tool/bindings/health from the same file itself — the helper is just bash's TOML eyes.
- **`tools/oracle-report-bmf.py`** — report JSON → Bencher BMF glue for the heavy tier (rag-rat emits JSON only; Bencher/markdown shaping is a glue concern).
- **`.github/workflows/oracle.yml`** — tier-driven:
  - **small** (PRs + push to main): GitHub-hosted matrix from `--list-tier small`, installs each corpus's SCIP tool by its `tool` field, runs the runner, uploads the report artifact. The health gate is the PR gate.
  - **heavy** (release / manual dispatch only): self-hosted big-memory box, serial (`max-parallel: 1`), bench image, pushed to Bencher as the headline resolution series.
- **Deletes** the superseded `oracle-rust.yml` / `oracle-kernel.yml` workflows and their `rust-scip-oracle.sh` / `kernel-c-oracle.sh` scripts (full migration, no shims); updates `docs/benchmarks.md`.

## Verification

End-to-end locally on the small tier (with the #175 `oracle report` binary):

| corpus | tool | edges | resolved before→after | result |
|---|---|---|---|---|
| rust-semver | rust-analyzer | 1056 | 412 → 936 | healthy, exit 0 |
| c-cjson | cmake compdb + scip-clang | 3941 | 2742 → 3408 | healthy, exit 0 |

Both clone → prepare → index → report → gate clean; the `corpus_profile_hash` matched the golden. The health gate was separately confirmed to fail non-zero on a violated threshold (in #175). `bash -n`, `py_compile`, and YAML parse all pass.

## Stacking

`oracle.yml`'s small tier needs the `oracle report` command (#175) and, for `py-requests`, the scip-python backend (#176). The script + helpers are independent of those at the file level (disjoint paths), but the workflow's `py-requests` leg goes green only once #176 lands and `rust-*`/`c-*` legs only once #175 lands. Merge order: #175 + #176, then this.


skakri:
Addressed in 202cbf0 — all six.

**P2 — Apply the corpus timeout to the whole run** ✅ The runner now re-execs itself once under `timeout -k 60s ${timeout_minutes}m`, so the budget covers clone + every `prepare` step + `index` + `report`, not just the report. An `EXIT` trap still removes the checkout on a timeout/gate-fail. Verified e2e (rust-semver: single re-exec, "budget 8m", completes healthy).

**P2 — Preserve Python virtualenv for scip-python** ✅ After the prepare steps, the runner activates a `.venv` it finds in the checkout (`VIRTUAL_ENV` + `PATH`) before indexing/reporting, so scip-python (pyright) resolves against the project's installed deps rather than the global interpreter.

**P2 — Include profile identity in the Bencher series** ✅ `oracle-report-bmf.py` now names the benchmark `{corpus_id}@{profile_hash[:12]}+{tool_version}/oracle` (was `{corpus_id}/oracle`), so a profile change (rev/bindings/prepare/threshold/tool) or a SCIP-indexer bump — both of which make reports incomparable — starts a fresh series instead of appending to the old one.

**P2 — Bound kernel make parallelism** ✅ `tools/oracle-corpora.toml` now uses `make -j$(nproc)` (was `make -j`, i.e. unlimited jobs). Recomputed the golden `linux-kernel` profile hash to match.

**P2 — Trigger the oracle on workspace dependency changes** ✅ Added root `Cargo.toml` and `Cargo.lock` to the `pull_request` + `push` path filters, so a workspace-dep/lock bump that changes the built binary can't skip the small-tier gate.

**P2 — Pin the rust-analyzer download** ✅ rust-analyzer is now installed as a rustup component (`dtolnay/rust-toolchain@stable` with `components: rust-analyzer`, resolved via `rustup which`), pinning the SCIP emitter to the stable toolchain (~6-week cadence) instead of `releases/latest` (weekly).

`bash -n`, `py_compile`, YAML parse, `cargo +nightly fmt --check`, `clippy`, and the core-oracle suite (incl. the recomputed golden) are all green.

skakri:
The two new **P1s** ("add the `oracle report` CLI before invoking it" and "add scip-python support before running py-requests") are both the same PR-stack ordering point, and they're correct: this PR is the **top of a 3-PR stack** and isn't meant to merge standalone.

- `oracle report` is added by **#175** (C2-CLI).
- the `scip-python` backend (`OracleTool::from_db_str` / `ALL`) is added by **#176** (B6).
- this PR (#177) is the runner + `oracle.yml` that drives them.

Merge order is **#175 → #176 → #177**: once the first two land on `main`, I rebase #177 onto `main` and its tree has both the `report` subcommand and the `scip-python` backend, so the small matrix (incl. the `py-requests` leg) and the heavy/Bencher path resolve. Until then #177's own CI is red purely because its dependencies aren't in this branch — the stacking artifact, not a defect in the runner. I'd rather not duplicate #175/#176 into this PR or temporarily exclude the Python corpus (scaffolding I'd just have to revert).

The earlier P2 round (pin rust-analyzer, trigger on `Cargo.toml`/`Cargo.lock`, preserve the venv, bound kernel `make -j$(nproc)`, profile-identity in the Bencher series) is addressed in 202cbf0 — those threads can be resolved.

PR #179, opened 2026-06-15
feat(resolve): Python from-import alias resolution scope (#174)

### LINKED ISSUE #172: Python Implements edges: resolver prefers trait/interface, not class
Python base-class inheritance emits an `Implements` edge (`extract.rs::python_edges` class_definition arm). But the resolver's `preferred_matches` for `implements` prefers `trait`/`interface` kinds, not `class` — so for a Python base (kind `class`), the `Implements` edge can mis-resolve to a same-named non-class symbol elsewhere in the index, or fall back ambiguously.

Impact is bounded today: python_edges ALSO emits a `ReferencesType` edge to the same base, and ReferencesType resolution DOES prefer type/class kinds — so the inheritance target is still captured. The risk is only the `Implements` edge resolving to the wrong symbol.

Options:
- Include class-like kinds in the `implements` resolution preference for Python (cross-language change — verify it doesn't mis-prefer in Kotlin/TS where `implements` targets interfaces).
- Or a Python-specific edge-kind/preference.

Context: Codex review round 3 on #167. Deferred from #167 because it's a resolver-side, cross-language change that deserves its own review; the base is already captured via ReferencesType in the meantime.

### LINKED ISSUE #174: Python import-alias resolution scope (alias → imported symbol)
Codex review on #167 (round 5). `from models import User as Account` then `Account()`: the Imports edge correctly targets `User`, but later `CallsName`/`ReferencesType` edges on the alias `Account` have no alias→target scope to connect them back to `User`, so they stay unresolved or bind to an unrelated local `Account`.

This needs a Python **import-alias scope** the resolver consults — analogous to the Rust per-module `use`-scope machinery (#61: `import_scope_start/end_byte`, `import_mod_id` columns + `imports::` resolution). It's a resolver-side subsystem, not an extraction tweak, so it's out of scope for the #167 indexing PR.

Sibling of #172 (Implements resolution preference) — both are Python resolution refinements. The import TARGETS are already captured correctly (so the dependency on `models.User` is visible); what's missing is binding alias *uses* to it.

skakri:
Fixes #174. `from models import User as Account` then `Account()` now resolves the alias use to the imported `User`, instead of staying unresolved or binding to an unrelated local `Account`.

## Approach

Mirrors the Rust per-module `use`-scope machinery (#61), but where Rust's `is_external_import` only *suppresses* a local same-named symbol, Python needs to *rebind* the alias to its in-corpus target. Three layers:

1. **Extraction** (`python_import_target`, gated to `from`-imports): an `aliased_import` emits its Imports edge to the **target** (`to_name = User` — keeps the in-corpus dependency) via `file_edge_scoped`, carrying the **alias** in `evidence` and a whole-file `ImportScopeRange` (Python imports are module/file-global). The edge is the alias carrier. Module aliases (`import x as m`) are left alone — they're a qualified-resolution problem, out of scope.
2. **Scope build** (`ImportScope`): a new `python_aliases` map (`alias → target`, byte-scoped) with `add_python_alias` / `python_alias_target`. The import-edge load branches on `files.language`, so Python imports register aliases instead of feeding `add_use` (which parses Rust `use` text).
3. **Rebind** (`resolve_symbol` callers): before resolution, a non-import reference whose name is an in-scope alias resolves under the **target** name. The stored `to_name` stays the surface alias; only `to_symbol_id` becomes the target's.

Applied in **both** resolution drivers (incremental `resolve_all_edges` and the full-rebuild `CompactEdge` loop) per the #61 both-driver-parity rule — `index --full` uses the full-rebuild path.

## Verification

- New unit test `python_from_import_alias_rebinds_to_the_imported_target` (resolve/tests.rs): with a target `models.User` and a **decoy** local `other.Account`, the `Account()` edge binds to `User`.
- Manual e2e on a temp project confirmed the same (resolved to `models.py::User`, not the decoy).
- Full `rag-rat-core` suite green; `cargo +nightly fmt --check` + clippy clean. The import target itself was already captured, so existing import/resolution tests are unaffected.

## Scope

Symbol aliases via `from x import T as A` only; whole-file scope (a safe over-approximation). Module aliases (`import x as m; m.f()`) remain a follow-up — sibling of the already-closed #172.

### DIFF
--- a/crates/rag-rat-core/src/index/edges/extract.rs
+++ b/crates/rag-rat-core/src/index/edges/extract.rs
@@ -628,13 +628,42 @@ pub(crate) fn python_edges(
                     EdgeConfidence::NameOnly,
                 ));
             }
+            // Record an alias for the rebind ONLY when the from-import is BOTH module-bound AND
+            // RELATIVE (`from .compat import X as Y`). Two gates:
+            //  - module-bound: a `def`/`class`-nested import binds the alias only in that local
+            //    scope, so a whole-file alias scope would rebind unrelated same-name references; it
+            //    binds file-wide at top level and inside transparent `if`/`try`/`with`/`for`
+            //    blocks.
+            //  - relative: a relative import provably names an IN-CORPUS sibling module, so
+            //    rebinding its alias to the in-corpus target is safe. An ABSOLUTE import (`from
+            //    urllib3.util import Timeout as TimeoutSauce`) is usually EXTERNAL — rebinding
+            //    `TimeoutSauce` → bare `Timeout` would mis-bind to a same-named LOCAL class
+            //    (`requests.exceptions .Timeout`), a real precision regression measured on
+            //    psf/requests (#174 review). Distinguishing absolute-in-corpus from
+            //    absolute-external needs a Python package model we don't have; relative is the
+            //    correct-by-construction in-corpus subset.
+            // A non-recorded alias still emits its plain target Imports edge (dependency captured).
+            let record_alias =
+                is_python_module_bound(node) && python_from_import_is_relative(node, text);
+            let import_start = node.start_byte();
+            // The module root is needed to bound the alias's scope at the next module-scope
+            // rebinding of the alias name (#174 review) — see `python_import_target`.
+            let module_root = record_alias.then(|| python_module_root(node)).flatten();
             let module_id = module.map(|m| m.id());
             let mut cursor = node.walk();
             for child in node.named_children(&mut cursor) {
                 if Some(child.id()) == module_id {
                     continue;
                 }
-                python_import_target(child, text, path, out);
+                python_import_target(
+                    child,
+                    text,
+                    path,
+                    record_alias,
+                    import_start,
+                    module_root,
+                    out,
+                );
             }
         },
         // `import <module>` / `import <module> as alias` — Imports edge to the module, not the
@@ -642,7 +671,7 @@ pub(crate) fn python_edges(
         "import_statement" => {
             let mut cursor = node.walk();
             for child in node.named_children(&mut cursor) {
-                python_import_target(child, text, path, out);
+                python_import_target(child, text, path, false, node.start_byte(), None, out);
             }
         },
         // Function / method / constructor call. Mirror the C handler: the callee is the LAST
@@ -795,7 +824,7 @@ fn emit_python_type_refs(
                 emit_python_type_refs(child, symbols, text, out);
             }
         },
-        "identifier" | "attribute" | "dotted_name" => {
+        "identifier" =>
             if let Some(name) = last_identifier_text(node, text) {
                 out.push(symbol_edge(
                     symbols,
@@ -805,8 +834,35 @@ fn emit_python_type_refs(
                     EdgeConfidence::NameOnly,
                     last_identifier_node(node).map(final_segment_node).map(CalleeRange::of_node),
                 ));
-            }
-        },
+            },
+        // A QUALIFIED type reference (`pkg.Account`, `a.b.C`, `Account.Inner`): record the receiver
+        // AND a dotted qualified name. The receiver_hint marks it qualified (so a bare-leaf alias
+        // rebind skips it — `pkg.Account` is NOT the local alias `Account`) AND lets a
+        // RECEIVER-alias rebind rewrite the root: `Account.Inner` with `Account` an alias
+        // for `User` resolves `User::Inner`, not bare `Inner` (#174 review). Without the
+        // qualified name the rebind would have nothing to rewrite and resolution would fall
+        // back to the ambiguous bare tail.
+        "attribute" | "dotted_name" =>
+            if let Some(name) = last_identifier_text(node, text) {
+                let identifiers = identifiers_under(node, text);
+                let receiver = node
+                    .child_by_field_name("object")
+                    .map(|object| node_text(object, text))
+                    .or_else(|| identifiers.first().cloned());
+                out.push(symbol_edge_with_context(
+                    symbols,
+                    node,
+                    "",
+                    name,
+                    EdgeKind::ReferencesType,
+                    EdgeConfidence::NameOnly,
+                    EdgeContext {
+                        receiver_hint: receiver,
+                        target_qualified_name: dotted_qualified_name(&identifiers),
+                    },
+                    last_identifier_node(node).map(final_segment_node).map(CalleeRange::of_node),
+                ));
+            },
         _ => {},
     }
 }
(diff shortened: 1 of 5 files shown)


skakri:
All four addressed in 447e5e2.

**P2 — Use file language in the full-rebuild path** ✅ The full-rebuild driver now builds a `file_language` map from the DB (`SELECT id, language FROM files`) instead of `index.file_language`, which is populated only from files that contributed a symbol. A symbol-less entrypoint (`from m import X as A; A()` with no def/class of its own) was skipped under `index --full`; it now registers and resolves. Verified e2e (a `run.py` that only imports + calls resolves `Account → models.User` under `--full`). The incremental driver already joined `files.language`, so the two are back in lockstep.

**P2 — Stop rebinding after same-module shadowing** & **P2 — Preserve the imported module** ✅ Both handled by deferring to a same-file definition: the rebind now fires only when the file defines **neither** the alias name **nor** the target's bare name locally (new `SymbolIndex::file_defines` + `python_alias_rebind`). So `from m import User as Account; class Account: ...; Account()` binds to the local `class Account`, and a file that also defines its own `User` won't have `Account` rewritten into the same-file `User` (it stays on normal resolution). Conservative — when either name is local, skip the rebind (no edge beats a wrong one). New unit test covers the shadow case + e2e confirms it.

**P2 — Don't give function-local imports file-wide scope** ✅ Extraction now records the alias only for **top-level** `from … import … as …` (the import statement's parent is the `module` node). An import nested in a function/class body emits its plain target edge (dependency still captured) but no file-wide alias, so unrelated same-name references aren't rebound.

Full `rag-rat-core` suite (483) green; fmt + clippy clean.

skakri:
Pushed `4feb8e8` reworking the alias rebind to model Python's order-dependent, scope-local name binding — addressing all five review points:

1. **Latest covering binding wins** — `python_alias_target` now selects the binding with the greatest `scope_start ≤ ref_byte`, so a re-import of the same alias (`from a import X as A; from b import Y as A`) resolves a later reference under the most recent import. (`python_alias_rebind_picks_the_latest_reimport`)
2. **Module-bound detection through transparent blocks** — `is_python_module_bound` walks ancestors to the `module` root, treating `if`/`try`/`with`/`for` as transparent and stopping at `def`/`class`/`lambda`. So `try: from x import Y as Z except` still binds `Z` file-wide, while a function-nested import gets no file-wide alias (the plain dependency edge is still emitted). (`try_block_from_import_alias_is_module_bound`, `function_nested_from_import_alias_is_not_recorded`)
3. **Qualified references are not rebound** — extraction marks qualified type/call refs (`other.Account`, `pkg.Account`) with a receiver hint; `python_alias_rebind` takes `is_qualified` and bails on it. (`python_alias_rebind_skips_a_qualified_reference`, `qualified_type_reference_carries_a_receiver_hint`)
4. **Order-aware shadow** — a same-name local definition *after* the import reassigns the name and shadows the alias from that point (`latest_def_before` vs the alias's `scope_start`); a definition *before* the import does not. (`python_alias_rebind_defers_to_a_same_file_definition`, `python_alias_rebind_overrides_a_definition_before_the_import`)
5. **Alias scope starts at the import statement** — not byte 0; a use before the import is out of scope and stays unresolved. (`python_alias_rebind_skips_a_use_before_the_import`)

Both drivers (incremental `resolve_all_edges` + full-rebuild `CompactEdge` loop) call the same `python_alias_rebind`. Verified e2e against a 4-file fixture: bare `Account()` → `User`; re-import → `Customer`; def-after-import → local class; `other.Account()` → unresolved. `cargo +nightly fmt --check`, `clippy --all-targets`, and the full `rag-rat-core` suite (491 tests) are green.

skakri:
Pushed \`bf0d4a3\` — implemented the full module-scope model for both points.

**1. Module-scope binding/order (resolve/mod.rs:48).** You're right that \`latest_def_before\` over all indexed symbols was the wrong basis — it can't see a lowercase \`Account = factory()\` (not indexed) and it counted nested defs. The shadow now lives in **extraction**: an alias's \`scope_end\` is bounded at the next **module-scope** rebinding of the name, computed by \`python_next_module_binding\` — it scans def/class/assignment/re-import statements, recurses through transparent \`if\`/\`try\`/\`with\`/\`for\` blocks, but never descends into a \`def\`/\`class\`/\`lambda\` body. So:
- \`Account = make_account()\` after the import now correctly stops the alias (verified e2e: the later \`Account()\` is no longer rebound to \`User\`).
- a nested \`class Account\` inside a function no longer false-shadows the module alias.
- an attribute set (\`Account.config = …\`) and a def *before* the import don't shrink the scope.

\`latest_def_before\` is removed; the resolver just honors the bounded \`scope_end\`.

**2. Receiver-position rebind (resolve/mod.rs:282).** \`from m import User as Account; Account.from_id()\` now rebinds the **receiver root** (\`Account.from_id\` → \`User.from_id\`, receiver \`User\`), so the method resolves on the imported class instead of staying unresolved (verified e2e: resolves \`scope_exact\` to \`User.from_id\`). A qualified reference only ever treats its **receiver** as an alias, never its leaf — \`pkg.Account\` is \`pkg\`'s member, not the local alias.

**Documented model boundary:** this is a module-scope model (matching "based on module-scope binding statements"). It does not do per-nested-scope lexical shadowing — a reference *inside* a \`def\` whose body locally rebinds the alias's exact name is still rebound to the module import. That's rare and would need full lexical-scope analysis at the reference site; I noted it in the \`python_alias_rebind\` doc comment. Happy to file a follow-up if you'd like it tracked.

Tests: resolve-side (\`scope_end\` shadow, qualified-receiver rebind, latest re-import, pre-import skip, non-alias-receiver skip) + extraction-side scope_end (class redef, lowercase assignment, nested-ignored, attribute-set-ignored, def-before-import). Both drivers share \`python_alias_rebind\`. fmt/clippy/full core suite (497) green.

skakri:
Pushed \`7c39061\` — all six addressed.

1. **Type aliases (935):** `type Account = int` is now in the rebinding detector, so it ends the alias scope.
2. **Value-less annotations (943):** an `assignment` only counts as a rebinding when it has a `right` value, so `Account: type[User]` no longer prematurely ends the alias.
3. **Plain dotted imports (980):** `python_import_binds_name` now distinguishes forms — a from-import binds the leaf, a plain `import other.Account` binds the top-level `other` (never the tail), so it doesn't end an `Account` alias.
4. **Assignment RHS timing (1023):** an assignment's effective rebinding byte is now the statement END, so `Account = Account()` keeps the RHS inside the alias scope (resolves to the import before the new binding takes effect).
5. **Mutually-exclusive branches (1023):** `python_next_module_binding` only counts UNCONDITIONAL top-level statements (direct module children), so neither `try`/`except` import shrinks the other; and `python_alias_target` returns None when two bindings cover the same byte with different targets — so the fallback is left ambiguous instead of resolved by byte order. Branches that agree on the target still resolve.
6. **Qualified type receivers (827):** a qualified type ref (`x: Account.Inner`) now carries a dotted `target_qualified_name`, so the receiver-alias rebind rewrites the root → `User::Inner` rather than searching the bare `Inner`.

Documented limitation (def/class header timing, 1023): a header annotation referencing the alias by the def's own name (`def Account(x: Account)`) stays bound to the new name; body refs are correctly excluded. Resolving the header would need true lexical/eval-order analysis at the reference site.

Tests: extraction-side (type alias, value-less annotation, plain dotted import, RHS-covers-scope, conditional-rebinding-ignored, qualified-type qn) + resolve-side (branch ambiguity → unresolved, agreeing branches → resolved, non-overlapping sequential re-import). fmt/clippy/full core suite (505) green.

skakri:
Checked the numbers on the \`py-requests\` oracle corpus (scip-python), and they surfaced a real regression — now fixed. Pushed \`dbd424d\`.

**What the oracle showed (main vs this branch, before the fix):** precision 89.91% → **89.47%**, recall flat, +2 contradicted edges. Both regressions were the same pattern:

\`\`\`python
from urllib3.util import Timeout as TimeoutSauce   # external
timeout = TimeoutSauce(...)
\`\`\`

The rebind rewrote \`TimeoutSauce\` → bare \`Timeout\`, which then bare-name-resolved to requests' OWN \`exceptions.Timeout\` (a different, in-corpus class). Root cause: Python has no external-import suppression, so rebinding an alias whose target is **external** lets bare resolution grab a same-named local symbol.

**Fix:** only record/rebind **relative** from-import aliases (\`from .compat import X as Y\`). A relative import provably names an in-corpus sibling module, so its alias is safe to rebind; an absolute import may be external and can't be cheaply distinguished from absolute-in-corpus without a Python package model. This is idiomatic-aligned — in requests, every in-corpus alias is relative and every external one is absolute.

**Re-measured on requests after the fix:** precision **90.00%** (now *above* the 89.91% baseline), recall unchanged, **+3 confirmed, 0 contradicted** — the feature is net-positive instead of break-even-with-2-wrong-edges.

Documented limitation: an absolute *in-corpus* alias (\`from myproject.models import X as Y\`) is no longer rebound (conservative; needs an in-corpus Python package model to do safely). Test: \`absolute_external_from_import_alias_is_not_recorded\`; the alias-scope tests now use relative imports. Full core suite (506) green.

skakri:
Pushed \`d5ea4ca\`. Two of the three are bounded module-scope rebinding forms the scope_end scan missed — fixed:

- **Starred unpacking (990):** `*Account, rest = rows` rebinds `Account`. `python_assignment_target_binds` now recurses `splat_pattern`/`list_splat_pattern` (and `expression_list`). Test: `alias_scope_ends_at_a_starred_unpacking_rebinding`.
- **`del Account` (976):** a module-level delete removes the binding — `delete_statement` is now a scope terminator. `del Account.attr`/`del Account[i]` correctly do NOT end the scope (attribute/subscript targets don't rebind the name). Tests: `alias_scope_ends_at_a_module_level_del`, `alias_scope_ignores_a_del_of_an_attribute`.

- **Relative import outside indexed targets (904):** this is the genuinely fundamental one — when the index covers only a *subpackage*, a parent relative import (`from ..models import X`) can point outside the indexed targets, and the rebind resolves the bare target against an unrelated in-corpus symbol. Closing it needs an **in-corpus Python module model** (resolve the relative module to an indexed file) — the same machinery the absolute-in-corpus and external-base cases all want, which rag-rat doesn't have yet. Documented as a limitation on the gate for now; I'm measuring whether this class is worth building the model for (see the cross-PR note I'll add).

fmt/clippy/full core suite (509) green.

skakri:
**Cross-PR note (the recurring root gap).** The fundamental issues reviewers keep finding here — external imported names binding to same-named local symbols — all trace to one missing piece: rag-rat has no Python external-import suppression / in-corpus module model (the Rust path has `local_crate_roots`/#61; Python doesn't). Filed as #182 with the concrete cases and the proposed minimal model.

Decision (data-backed): defer the model. On the `py-requests` oracle the measured impact is small — the alias feature is +3 confirmed / 0 contradicted after the relative-only gate (precision 89.91% → 90.00%), and the implements preference is neutral (identical to baseline). The current PRs are conservative so they don't emit wrong high-confidence edges; the residual cases are documented limitations tracked in #182, to be revisited if a heavier Python corpus shows the gap materially hurting precision.

skakri:
Triaged — none need action on this PR; all three fold into the #182 tracker:

- **647 (absolute in-corpus alias):** recall miss (not a wrong edge) — the relative-only gate deliberately doesn't rebind absolute imports because we can't tell in-corpus from external. Recording it needs the in-corpus Python module model (#182), which is exactly what would resolve the module to an indexed file.
- **991 (`for`/`with` module-level targets):** wrong-edge but only when an imported class-alias name is reused as a module-level loop/with variable — vanishingly rare and oracle-negligible. Added to #182 to handle (`for_statement` left / `with` as_pattern) when the module model lands, alongside the other binding forms.

Keeping the (now-squashed) branch stable rather than adding per-form code for ~zero measured impact.

skakri:
These fold into #182 — both are the lexical/module-scope class we're deferring:

- **Nested-scope shadowing** (`def f(Account): Account()`): this is the documented model boundary in `python_alias_rebind`. A function parameter or lowercase local that shadows the alias isn't indexed as a symbol, so detecting it needs reference-site scope analysis.
- **Wildcard re-export** (`from .other import *`): knowing whether `other` exports `Account` needs the in-corpus module model.

Both are wrong-edge but require reusing an imported alias name as a function param/local or having a wildcard re-export it — rare and oracle-negligible. Keeping the squashed branch stable; tracked in #182.

PR #180, opened 2026-06-15
fix(resolve): prefer a base CLASS for Python implements edges (#172)

### LINKED ISSUE #172: Python Implements edges: resolver prefers trait/interface, not class

skakri:
Fixes #172. Python base-class inheritance (`class Sub(Base)`) emits an `implements` edge to `Base`, which is a **class** — Python has no traits/interfaces. The resolver's `implements` preference was `["trait", "interface"]`, so a Python base never matched and the edge could mis-resolve to a same-named non-class or fall back ambiguously.

## Fix

`preferred_matches` is now **language-aware**: for Python, `implements` prefers `["class", "object"]`; Kotlin / TS / Rust are **unchanged** (`["trait", "interface"]`) — there `implements` / `: I` targets an interface, and a same-named class must not be preferred over it. The source language is already threaded into `resolve_symbol` (`request.source_language`), so this is a one-line branch in the existing preference table.

## Verification

- New resolve test `python_implements_prefers_a_base_class_over_a_non_class`: with a base `class Base` and a **decoy** same-named module-level `function Base`, the Python `implements` edge binds to the class.
- e2e on a temp project: `class Sub(Base)` resolves `implements → base.py::Base (class)` at Syntactic confidence (was unresolved/ambiguous before).
- Full `rag-rat-core` suite (482) green; fmt + clippy clean. Kotlin/TS `implements` tests unaffected (the branch is gated to Python).

Sibling of #174 (alias resolution, PR #179) — both Python resolver refinements deferred from #167. The remaining `import x as m` module-alias case is still out of scope (noted in #174).

### DIFF
--- a/crates/rag-rat-core/src/index/edges/extract.rs
+++ b/crates/rag-rat-core/src/index/edges/extract.rs
@@ -688,20 +688,40 @@ pub(crate) fn python_edges(
                     {
                         continue;
                     }
-                    // Implements points at the base HEAD only (`Generic`, not `T`); ReferencesType
-                    // covers the head AND every type argument (`Mapping[str, Api]` → `Mapping`,
-                    // `str`, `Api`).
-                    let head = python_type_head(base);
-                    if let Some(name) = last_identifier_text(head, text) {
+                    // Emit Implements ONLY for a STATIC head — a plain identifier, or an attribute
+                    // whose receiver chain is all identifiers and not `self`/`cls` (`pkg.Base`),
+                    // after unwrapping generic/subscript/paren wrappers (#172 review). A DYNAMIC
+                    // base has no compile-time class — `factory()`,
+                    // `factory().Base`, `self.Base`, `Base if flag else Other`,
+                    // a lambda, … — so claiming an Implements edge would
+                    // let the Python class preference mis-bind it to a same-named local class. An
+                    // allowlist (vs blocklisting each dynamic form) is robust to new expression
+                    // kinds.
+                    //
+                    // Implements targets the base HEAD's LEAF name (`Base` for
+                    // `pkg.Base`/`Generic[T]` → `Generic`); ReferencesType
+                    // (below) covers the head and every type argument. The edge
+                    // is bare-name (no qualified context): a module-qualified base `pkg.Base`
+                    // is resolved by the leaf `Base` exactly like a bare base, because a top-level
+                    // Python class's `scope_path` is the bare name, not `pkg::Base`. The cost is
+                    // that an EXTERNAL `pkg.Base`/bare imported base can still
+                    // bind a same-named local class — the general "Python has
+                    // no external-import suppression" gap (#172/#174
+                    // review), which needs an in-corpus Python module model to close, not a
+                    // per-base special case.
+                    if let Some(head) = python_static_base_head(base, text)
+                        && let Some(name) = last_identifier_text(head, text)
+                    {
+                        let callee = last_identifier_node(head)
+                            .map(final_segment_node)
+                            .map(CalleeRange::of_node);
                         out.push(symbol_edge(
                             symbols,
                             base,
                             name,
                             EdgeKind::Implements,
                             EdgeConfidence::NameOnly,
-                            last_identifier_node(head)
-                                .map(final_segment_node)
-                                .map(CalleeRange::of_node),
+                            callee,
                         ));
                     }
                     emit_python_type_refs(base, symbols, text, out);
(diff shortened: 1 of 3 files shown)


skakri:
Both addressed in a528135.

**P2 — Avoid preferring foreign classes for Python bases** ✅ The Python `implements` branch in `preferred_matches` now filters by **language** as well as kind (in-corpus Python `class`/`object` only). And — since an empty preferred set otherwise falls back to the all-matches set — a Python `implements` that finds no in-corpus Python class now stays **unresolved** (mirroring the `references_type` no-fallback rule) instead of binding a same-named TS/Kotlin/C++ class or non-class. The `references_type` edge still records the base. Test: a TS `class Base` is not bound by a Python `implements`.

**P2 — Avoid forcing call-style Python bases to classes** ✅ Extraction no longer emits an `Implements` edge for a **call-shaped** base (`class Sub(factory())`) — a dynamic base whose head is a callable, not the base class, where the class preference would mis-bind to a same-named class. The base call is still captured as a `CallsName`, so the dependency on `factory` isn't lost. Test: a call-shaped base emits no `Implements` (keeps the `CallsName`).

Full `rag-rat-core` suite (484) green; fmt + clippy clean.

skakri:
Good catch — pushed \`3b20178\`. The guard only checked the base's immediate kind, so a parenthesized call (\`class Sub((factory()))\`, where tree-sitter nests the \`call\` under a \`parenthesized_expression\`) slipped through and mis-bound to \`factory\`.

Replaced \`base.kind() != "call"\` with a new \`python_base_is_dynamic\` helper that unwraps \`parenthesized_expression\` layers (bounded loop) before deciding. Added two tests: \`parenthesized_call_shaped_base_emits_no_implements\` (no Implements, but \`factory\` still captured as a CallsName) and \`parenthesized_class_base_still_emits_implements\` (a parenthesized *static* base \`(Base)\` keeps its Implements edge). fmt/clippy/tests green.

skakri:
Right again — pushed \`024df5f\`. \`class Sub(factory()[T])\` exposes the base as a \`subscript\` whose value is the \`call\`, and since the Implements path runs \`python_type_head\` (which unwraps \`subscript\` → \`value\` → the call), the dynamic check has to unwrap the same way.

Aligned \`python_base_is_dynamic\` with \`python_type_head\`: it now unwraps \`subscript\` (→ value), \`type\`/\`generic_type\` (→ first child), and parentheses — so any \`call\` reachable as the Implements head is recognized as dynamic and gets no Implements edge. Invariant: if the Implements path would land on a \`call\`, the dynamic check catches it. Added \`subscript_call_shaped_base_emits_no_implements\`; the existing \`Generic[T]\`/\`Mapping[str, Api]\` base tests still pass (their head is an identifier, not a call). fmt/clippy/tests green.

skakri:
Pushed \`3b34110\`. \`class Sub(factory().Base)\` — the base is an \`attribute\` whose receiver is a \`call\`, so \`Base\` comes off the factory result, not a static class. \`python_base_is_dynamic\` now follows the \`attribute\` \`object\` chain: a call receiver → dynamic (no Implements), a static \`pkg.Base\` bottoms out at an identifier → still Implements. The subscripted form \`factory().Base[T]\` is handled too (subscript → value → attribute → call). Tests: \`attribute_on_call_base_emits_no_implements\` (both forms) + \`static_attribute_base_still_emits_implements\`. fmt/clippy/tests green.

skakri:
Pushed \`e6eacd5\` — addressed both remaining points by switching from a dynamic-base **blocklist** to a static-head **allowlist**, which is robust to new expression forms.

**Conditional (and other) dynamic bases:** \`python_static_base_head\` now emits an Implements edge ONLY when the base's head is static — a plain identifier, or an attribute whose receiver chain is all identifiers (\`pkg.Base\`) — after unwrapping generic/subscript/paren wrappers. Everything else is excluded by default: \`Base if flag else Other\` (conditional_expression), \`factory()\`, \`factory().Base\`, \`factory().Base[T]\`, lambdas, binary operators, etc. Test: \`conditional_expression_base_emits_no_implements\`.

**Qualified base context:** a qualified static base (\`pkg.Base\`) now emits its Implements edge WITH the receiver hint (\`pkg\`) and dotted qualified name (\`pkg::Base\`), so resolution can leave an external base unresolved instead of binding it to a same-named local \`class Base\`. Updated \`static_attribute_base_still_emits_implements\` to assert the carried context.

(\`python_type_head\` folded into \`python_static_base_head\`.) fmt/clippy/tests green.

skakri:
Pushed \`1a08279\` — both were gaps the qualified-context (round 5) opened for an external/non-class qualified base:

1. **Empty evidence (extract.rs:715).** The qualified Implements edge passed \`""\`, so \`allow_unqualified_fallback\` couldn't see the dotted form and still fell back to bare \`Base\`. Now passes \`text\` → the edge carries dotted evidence (\`pkg.Base\`), and the Python branch of \`allow_unqualified_fallback\` (dotted evidence + non-self receiver → Kotlin-only) refuses the bare fallback, so an external \`pkg.Base\` stays unresolved.

2. **Qualified arms skipped the class filter (extract.rs:721 / resolve).** The scope/qualified lookup arms only consulted \`kind_matches\` (macro-only), so \`class Sub(Pkg.Base)\` where \`Base\` is a method could bind via \`scope_exact\` before the bare-name \`preferred_matches\` class filter ran. Extended \`kind_matches\` to apply the Python-implements class+language restriction to the qualified arms too; a non-class qualified target now falls through to the existing Python-implements no-fallback guard → unresolved.

Tests: \`python_qualified_external_base_does_not_fall_back_to_local_class\`, \`python_qualified_implements_does_not_bind_a_non_class\`, and \`python_qualified_implements_binds_an_in_corpus_class\` (the restriction doesn't over-reach — a qualified base that IS an in-corpus class still resolves). fmt/clippy/tests green.

skakri:
Pushed \`f3a1491\`. Triaged the three; one is a bounded fix, two are the same fundamental gap.

**Fixed — `self.Base` (871):** `python_attribute_is_static` now returns false for a `self`/`cls`-rooted attribute chain, so `class Sub(self.Base)` is dynamic → no Implements. Test: `self_qualified_base_emits_no_implements`.

**Reverted the qualified-context-on-Implements (728 + 772):** that round-5 change regressed in-repo module-qualified bases (728: `import base; class Sub(base.Base)` went unresolved because a top-level Python class's `scope_path` is the bare `Base`, not `base::Base`, so the qualified arms could never hit it — and the dotted-evidence fallback-suppression then left it unresolved). It also never actually fixed the external case (772). A qualified base now emits a **bare** Implements to the leaf `Base`, resolved like any bare base. The `kind_matches` qualified-arm change and its now-unreachable tests are reverted with it.

**The remaining gap (772, and round-5's `pkg.Base`) is fundamental:** an external bare/qualified imported base can bind a same-named local class — because Python has no external-import suppression. It's the *same* root as #179's `TimeoutSauce` and the relative-out-of-scope case: rag-rat lacks an in-corpus Python module model. A per-base special-case can't fix it consistently (that's what caused the 728 regression). Documented in the base-loop comment.

**Data:** I ran the `py-requests` oracle on this branch vs main — **identical** (precision 89.91%, recall 86.79%, 303 confirmed / 34 contradicted). The implements class preference is neutral on requests (it has no same-named base-class collisions to exercise it), so the feature isn't harming precision and the flagged external-base risk doesn't manifest here. fmt/clippy/core suite (491) green.

See the cross-PR note below on whether to build the Python module model.

skakri:
**Cross-PR note (the recurring root gap).** The fundamental issues reviewers keep finding here — external imported names binding to same-named local symbols — all trace to one missing piece: rag-rat has no Python external-import suppression / in-corpus module model (the Rust path has `local_crate_roots`/#61; Python doesn't). Filed as #182 with the concrete cases and the proposed minimal model.

Decision (data-backed): defer the model. On the `py-requests` oracle the measured impact is small — the alias feature is +3 confirmed / 0 contradicted after the relative-only gate (precision 89.91% → 90.00%), and the implements preference is neutral (identical to baseline). The current PRs are conservative so they don't emit wrong high-confidence edges; the residual cases are documented limitations tracked in #182, to be revisited if a heavier Python corpus shows the gap materially hurting precision.

skakri:
Same lexical/module-model class as the rest (#182). `def make(provider): class Sub(provider.Base)` has a parameter receiver, not a module — and there's no name-free way to tell a module namespace from a local without scope/import resolution (content evidence like `pyvenv.cfg` only works for venv dirs, not receivers). It's a wrong-edge but requires a class whose base comes off a local/param, which is rare and oracle-negligible. Deferred to #182 with the other module-model cases; keeping the squashed branch stable.

PR #181, opened 2026-06-15
fix(init): Python root-entrypoint + env-only directory selection (#173)

### LINKED ISSUE #173: init scanner: Python directory-selection edge cases (root entrypoints, env-only repos)
Follow-ups from Codex review on #167, after the index walker floors venv/site-packages (`FLOOR_DIRS`) — so the *danger* (indexing installed deps) is resolved; these are init-UX completeness refinements:

1. **Root-level entrypoints omitted when a package dir exists.** A repo with both root `manage.py`/`main.py` AND a package dir: the package dir is the natural default, so the fallback never promotes `.`, and `directly_contains_source` returns false for `.` — so `init -y` writes only the package binding and omits the root entrypoints. Want: include the repo root (or root `.py` files) when it directly contains Python source alongside a package dir.

2. **env-only repos still get `python = ["."]`.** When the only `.py` files are under `env/`/`.env/`, `add_file_to_dir_counts` increments the aggregate `.` candidate, which the fallback promotes (`.` isn't a dependency-named path). Now SAFE — the walker floors `site-packages`/`.venv`/`venv`/`__pycache__`, so `python=["."]` won't ingest the deps — but cosmetically the binding is broader than ideal. Want: don't promote `.` when its only Python files live under a dependency tree.

Both are in `crates/rag-rat-cli/src/init/scan.rs` (`default_dir` / `candidate_dirs` fallback / `directly_contains_source`). Deferred from #167 to keep the language PR focused; the walker floor + the fallback dependency-dir filter already shipped there cover the safety-critical part.

skakri:
Fixes #173. Two init-scanner UX refinements for Python directory selection. The safety-critical part (the walker flooring `site-packages`/`.venv`/`venv`/`__pycache__`) already shipped in #167 — these are about what `init -y` *writes*.

## 1. Root entrypoints alongside a package dir

A repo with root `manage.py`/`main.py` **and** a package dir wrote only the package binding, omitting the root files: `.` (the aggregate bucket every file increments) was never recognized as *directly* containing source. Now root-level files key under `.` in `direct_dir_counts`, and a new `python_root_has_direct_source` makes `.` a default when the root holds genuine entrypoints — so both `.` and the package dir are offered.

## 2. env-only repos no longer promote `.`

When every `.py` lives under a dependency tree (`env/`, `.venv/…/site-packages`, …), the no-default fallback would pick `.` on aggregate count and write `python = ["."]` (over installed deps). The fallback now skips `.` for Python unless the root directly contains real source — so an env-only repo gets no Python binding rather than a wrong one.

## Tests

- `root_entrypoints_default_alongside_a_package_dir` — `manage.py` + `myapp/` → both `.` and `myapp` are defaults.
- `fallback_does_not_promote_dot_when_python_lives_only_under_a_dependency_tree` — realistic env-only repo (`.` carries the aggregate count) → no default.
- The existing `fallback_does_not_promote_a_venv_only_python_repo` stays green; full `rag-rat` suite + fmt + clippy clean. The count-keying change is Python-guarded in the selection logic, so other languages' dir selection is unaffected.

Last of the #167 Python follow-ups (with #172 / PR #180 and #174 / PR #179).

### DIFF
--- a/crates/rag-rat-cli/src/init/mod.rs
+++ b/crates/rag-rat-cli/src/init/mod.rs
@@ -11,6 +11,7 @@ use rag_rat_core::config::EmbeddingBackend;
 use rag_rat_core::index::ai::{
     FASTEMBED_MODEL_ID, HASH_MODEL_ID, MODEL2VEC_MODEL_ID, ReconcileOptions,
 };
+use rag_rat_core::index::ignore_rules::{IgnoreMatcher, is_virtualenv_dir};
 use rag_rat_core::language::Language;
 use rag_rat_core::{Config, IndexDatabase};
 pub(crate) use render::*;
@@ -31,9 +32,10 @@ const SKIPPED_DIRS: &[&str] = &[
     ".venv",
     // Python virtualenv / dependency / cache trees — never project source. Skipping them at scan
     // time means their `.py` files never become dir candidates, so the "no default → promote the
-    // largest" fallback can't select a `site-packages` tree (#167 review).
+    // largest" fallback can't select a `site-packages` tree (#167 review). NB: `virtualenv` is NOT
+    // here — it's a real first-party package name; a virtualenv literally named `virtualenv/` is
+    // caught by content (`pyvenv.cfg`) in `scan_dir` instead (#181).
     "venv",
-    "virtualenv",
     "site-packages",
     "__pycache__",
     ".tox",
@@ -70,6 +72,14 @@ pub(crate) struct RepoScan {
     dir_counts: BTreeMap<Language, BTreeMap<PathBuf, usize>>,
     direct_dir_counts: BTreeMap<Language, BTreeMap<PathBuf, usize>>,
     total_source_bytes: u64,
+    /// The scan found a real Python virtualenv (a dir with a `pyvenv.cfg`) ANYWHERE the index
+    /// would walk — not gitignored, not floored. The indexer floor can't cover a venv under an
+    /// ambiguous name (`env`/`virtualenv`), so a `python = ["."]` walk WOULD index it; when
+    /// one is present we must not auto-bind `.`. Gitignored / conventionally-floored venvs
+    /// (`.venv`/`venv`) and first-party packages that merely share a venv-ish NAME (no
+    /// `pyvenv.cfg`) are NOT recorded here — content detection, not the name, decides (#181
+    /// review).
+    has_python_virtualenv: bool,
 }
 
 impl InitOptions {
@@ -284,6 +294,7 @@ mod tests {
             ]),
             direct_dir_counts: BTreeMap::new(),
             total_source_bytes: 0,
+            has_python_virtualenv: false,
         };
 
         let plan = default_plan(".".to_string(), &scan);
@@ -319,6 +330,7 @@ mod tests {
                 ]),
             )]),
             total_source_bytes: 0,
+            has_python_virtualenv: false,
         };
 
         let defaults = candidate_dirs(&scan, Language::C)
--- a/crates/rag-rat-cli/src/init/run.rs
+++ b/crates/rag-rat-cli/src/init/run.rs
@@ -11,6 +11,17 @@ pub(crate) fn run(args: &crate::cli::InitArgs, config_path: &str) -> anyhow::Res
     } else {
         prompt_plan(root, root_value, &scan)?
     };
+    // A plan with no bindings would render an empty `[target_bindings]` and index nothing — a
+    // useless config. `init -y` can reach this when every detected language drops to a dependency
+    // tree (e.g. Python whose only `.py` live under a virtualenv); the interactive path already
+    // bails, so guard the non-interactive path the same way rather than write an unusable file
+    // (#181 review).
+    if plan.bindings.is_empty() {
+        anyhow::bail!(
+            "init found no indexable source to bind — every detected language resolved to a \
+             dependency tree (e.g. a virtualenv). Nothing to configure."
+        );
+    }
     let config_text = render_config(&plan);
 
     if options.dry_run {
@@ -100,9 +111,22 @@ pub(crate) fn prompt_plan(
         let dirs: Vec<_> =
             selected.into_iter().map(|index| candidates[index].path.clone()).collect();
         if dirs.is_empty() {
-            anyhow::bail!("init needs at least one selected root for {}", language.as_str());
+            // No roots selected — drop the language (don't bind it), matching the non-interactive
+            // plan which omits a language with no safe default (e.g. env-only Python, whose
+            // candidates are all dependency trees and so default to none). Bailing here would make
+            // accepting the offered defaults fail for exactly that case (#181 review).
+            continue;
         }
         bindings.insert(*language, dirs);
+    }
+    // Keep `languages` consistent with the bindings actually chosen (a dropped language must not
+    // linger and make `render_config` emit an empty list).
+    let languages = languages
+        .into_iter()
+        .filter(|language| bindings.contains_key(language))
+        .collect::<Vec<_>>();
+    if bindings.is_empty() {
+        anyhow::bail!("init needs at least one selected root to index");
     }
 
     let backend = prompt_backend(scan)?;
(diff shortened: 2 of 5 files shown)


skakri:
✅ Fixed in fb1bc83. You're right — `candidate_dirs` left env-only Python with no default, but `default_plan` reconverted an empty default list back to `["."]`, so `init -y` still wrote `python = ["."]` over the dependency tree.

`default_plan` now **omits the Python binding** (and drops Python from the language list) when there are candidates but none is a safe default — instead of the `vec![PathBuf::from(".")]` fallback. For other languages `.` stays the reasonable "index everything" default; only Python's env-only case (which `candidate_dirs` deliberately refuses to promote) is carried through as "no binding." Added `default_plan` tests: env-only repo → no Python binding; a real package dir still binds.

skakri:
Pushed \`f8035b7\`. The real gap was floor parity: the indexer's hard floor (\`FLOOR_DIRS\` in \`ignore_rules.rs\`) covered \`.venv\`/\`venv\`/\`site-packages\`/\`__pycache__\` but not \`virtualenv\`/\`.tox\`/\`.nox\`, so once \`.\` is a Python binding the walk could ingest the venv scaffolding *outside* \`site-packages\`.

I floored the unambiguous venv/tooling dir names (\`virtualenv\`, \`.tox\`, \`.nox\`) so they're never walked regardless of config — that's the "floor parity" option, and it's the right home for "never index a virtualenv tree" since it also protects a hand-written \`python = ["."]\`.

I deliberately left \`env\`/\`.env\` **out** of the global floor, matching the existing decision in the floor's own doc comment (and the init scanner's \`is_python_dependency_dir\` vs \`SKIPPED_DIRS\` split): \`env\` is too generic to assume a virtualenv globally, and its installed deps are still caught by the \`site-packages\` floor. Flooring \`env\` globally would refuse to index a legitimately-named \`env/\` source dir (and a Django root with a venv named \`env/\` + \`manage.py\` is common — we don't want to drop root indexing entirely there). New test \`python_venv_floor_dirs_ignored_but_generic_env_indexed\` pins both halves. fmt/clippy/tests green.

If you'd rather be maximally conservative and also exclude \`env\`/\`.env\` when emitting \`.\`, say the word — but it'd need the richer \`[[target]]\` exclude form (the simple \`[target_bindings]\` render can't carry excludes), so I held off.

skakri:
Pushed \`453e07a\` — addressed all three.

**1. Floor regression (ignore_rules.rs:79).** You're right, \`virtualenv\` is a real PyPI package (\`src/virtualenv/…\` is first-party). Removed it from \`FLOOR_DIRS\`. The floor now keeps only names that can NEVER be a first-party import package: \`.venv\`/\`.tox\`/\`.nox\` (dotted — a package dir can't start with a dot) and \`venv\`/\`site-packages\`/\`__pycache__\` (reserved). \`env\`/\`.env\`/\`virtualenv\` are handled by the scanner instead. Test updated to assert \`src/virtualenv/__init__.py\` indexes.

**2. Dot-default leak (scan.rs:165).** \`scan_dir\` now records when a Python dependency/virtualenv dir sits directly at the repo root — checked *before* the scan-skip, so \`venv\`/\`virtualenv\` (which the scan otherwise skips) still count — and \`python_root_has_direct_source\` refuses the \`.\` default when one is present. So a \`manage.py\`-plus-\`env/\` repo no longer writes \`python = ["."]\`; Python is omitted rather than indexing the venv (\`.\` still binds for a clean root entrypoint). I went with "avoid \`.\` when a dependency dir is present" rather than per-target excludes because the simple exclude-glob matcher (\`env/**\` → prefix \`env\`) would also catch a legit \`environment/\` dir.

**3. Interactive parity (scan.rs:126).** \`prompt_plan\` now *drops* a language when no roots are selected (env-only Python) instead of bailing, matching \`init -y\`'s omission, and keeps the language list consistent with the bindings (with a final bail only if nothing at all is selected).

Tests: \`root_entrypoint_binds_dot\` (clean root → \`.\`), \`root_entrypoint_with_root_venv_does_not_bind_dot\` (root venv → omitted), updated floor test. fmt/clippy/tests green.

skakri:
Pushed \`052b403\` — good catch, that was an over-reach. The previous round recorded ANY root-level dependency dir, including ones the indexer already floors (\`.venv\`/\`venv\`/\`node_modules\`/\`site-packages\`/…). A \`python = ["."]\` walk never ingests those, so suppressing \`.\` for them wrongly dropped real \`manage.py\`-style entrypoints in the common project-with-a-local-\`.venv\` case.

Narrowed the detection to a new \`is_unfloored_python_dependency_dir\` (\`env\`/\`.env\`/\`virtualenv\` — exactly the names the floor can't safely carry, so the only ones a \`.\` walk would actually ingest). Renamed the flag to \`root_has_unfloored_python_dep_dir\` and added \`only_unfloored_dependency_dirs_block_the_dot_default\` pinning that \`.venv\`/\`venv\`/\`node_modules\` do NOT block \`.\` while \`env\`/\`.env\`/\`virtualenv\` do. fmt/clippy/tests green.

skakri:
Pushed \`89f4477\`. All three were the same root cause — the scanner re-implemented a crude subset of the index walk (only \`SKIPPED_DIRS\` + a name heuristic) and diverged from it. Fixed the class rather than patching each:

**Scan now honors the same ignore rules as the index** (scan.rs:56 + :231). \`scan_dir\` consults the shared \`IgnoreMatcher\` (gitignore + floor — exactly what the index walk uses), so "the scan sees a dependency dir" ⟺ "the index would ingest it under \`.\`". Consequences:
- The unfloored-env check (\`env\`/\`.env\`/\`virtualenv\`) is now **depth-agnostic** — a nested \`tools/env/\` is caught, not just a root one (comment 1).
- A **gitignored** \`env/\`/\`.venv/\` is \`is_ignored\` → skipped by the scan (as the index skips it) → it no longer blocks \`.\`, so a \`manage.py\`-plus-gitignored-venv repo keeps its root entrypoint (comment 2).
- A non-gitignored, unfloored env (root or nested) still blocks \`.\` — because the index genuinely would ingest it.

**Empty-plan bail** (run.rs:204). \`init -y\` now fails when the plan has no bindings instead of writing an empty \`[target_bindings]\` and indexing nothing — matching the interactive path (comment 3).

Verified e2e: gitignored \`.venv\`+\`env\` + \`manage.py\` → \`python = [".", "myapp"]\`; nested non-gitignored \`tools/env/\` + package → \`python = ["myapp"]\` (\`.\` dropped, package kept); non-gitignored env-only → init fails. Added an integration test (\`init_dir_selection.rs\`) covering the gitignored-venv and empty-config cases. fmt/clippy/tests green.

skakri:
Pushed (re-squashed to one commit, `1d32657`). All three addressed — and the test isolation one was a real bug that mutated developer config, so thanks for catching it.

**Test isolation (init_dir_selection.rs:39) — the real issue.** The tests ran the full `rag-rat init -y`, which after rendering also calls `offer_mcp_install`/`offer_hooks_install` and indexes — so they actually installed MCP servers and shelled out to `claude`/`codex`. They now run `init -y --dry-run`, which renders the config to stdout and returns BEFORE any write/index/install — pure planning, no side effects, no git needed. (I also cleaned up a stray `rag-rat` entry a prior run left in my `~/.codex/config.toml` pointing at a deleted temp dir.)

**`virtualenv` package over-blocked (213) + nested env ancestor (140) — fixed at the root by content detection.** The env/virtualenv *name* ambiguity is what caused both (and several earlier rounds), so I replaced the name heuristic with `ignore_rules::is_virtualenv_dir` — a real venv has a `pyvenv.cfg`; the `virtualenv` *package* doesn't. The scanner now skips a dir only when it's a real venv by content (and doesn't count its files), so:
- `src/virtualenv/` (first-party, no `pyvenv.cfg`) → bound normally (`python = ["src"]`), not dropped (213).
- a real `tools/env/` venv (`pyvenv.cfg`) → skipped, so `tools` is never promoted and `.` is blocked; only the real package binds (`python = ["myapp"]`) (140).

`virtualenv` is also removed from the scan's `SKIPPED_DIRS` (name-skip) for the same reason — content detection handles the venv case. New integration tests cover both (`first_party_virtualenv_package_is_bindable`, `nested_real_venv_does_not_promote_its_ancestor`) plus a `is_virtualenv_dir` unit test. fmt/clippy/tests green.

PR #184, opened 2026-06-16
feat(oracle): scip-typescript backend + ts-ky corpus

skakri:
Adds **scip-typescript** as a fourth SCIP oracle backend, so TypeScript/TSX
edges get upgraded to the `Compiler` confidence tier and TypeScript appears
as a row in the oracle resolution report table.

## What's in it

- **`OracleTool::ScipTypescript`** variant (`oracle/mod.rs`): `as_db_str`/`from_db_str` round trip + `ALL`.
- **Manifest entry** (`oracle/manifest.rs`): program `scip-typescript`, languages `[typescript]`, install hint. Invocation `scip-typescript index --cwd <root> --infer-tsconfig --output <abs>`.
  - **No `--project-name` / `--project-version` pin** (unlike scip-python): `package.json` supplies a commit-stable name + version, so monikers don't churn across commits.
  - `--infer-tsconfig` synthesizes a `tsconfig.json` from `package.json` when the project doesn't ship one.
  - Shares scip-python's `can_emit_scip` (`index --help`) and the no-prerequisite arm (cross-package refs resolve against installed `node_modules`; a non-installed env is caught by the moniker health gate).
- **CLI** `oracle run --tool scip-typescript` (`cli.rs`).
- **`ts-ky` corpus** (`tools/oracle-corpora.toml`): sindresorhus/ky v1.7.2, tier `small`, `npm install` prepare, `bindings = { typescript = ["source"] }`; golden profile hash pinned in `oracle/corpus.rs`.
- **Docs + CI**: `docs/oracle.md` tool list; `.github/workflows/oracle.yml` install case with `SCIP_TYPESCRIPT_VERSION` pinned to `0.4.0`.
- New manifest test + round-trip test now iterates `OracleTool::ALL`.

## Validation

E2E-validated against ky locally: **616 edges, 220 confirmed / 24 contradicted, 87 monikers**, scip-typescript v0.4.0. Full `cargo +nightly fmt --check`, `cargo clippy --all-targets`, and `cargo test` (core 521 + CLI) green.

### DIFF
--- a/.github/workflows/oracle.yml
+++ b/.github/workflows/oracle.yml
@@ -58,11 +58,12 @@ concurrency:
 
 env:
   # Pinned SCIP toolchain for the small tier so a PR number isn't perturbed by an unrelated indexer
-  # release. scip-clang/scip-python are pinned to explicit versions; rust-analyzer is installed as a
+  # release. scip-clang/scip-python/scip-typescript are pinned to explicit versions; rust-analyzer is installed as a
   # rustup component (below), pinning it to the stable toolchain (a ~6-week cadence) instead of the
   # weekly `releases/latest` — so a fresh RA build can't change `rust-semver`'s numbers mid-PR.
   SCIP_CLANG_VERSION: v0.4.0
   SCIP_PYTHON_VERSION: 0.6.6
+  SCIP_TYPESCRIPT_VERSION: 0.4.0
 
 jobs:
   matrix:
@@ -129,6 +130,9 @@ jobs:
             scip-python)
               npm install -g "@sourcegraph/scip-python@${SCIP_PYTHON_VERSION}"
               scip-python --version ;;
+            scip-typescript)
+              npm install -g "@sourcegraph/scip-typescript@${SCIP_TYPESCRIPT_VERSION}"
+              scip-typescript --version ;;
             *)
               echo "unknown tool '$tool'" >&2; exit 1 ;;
           esac
--- a/crates/rag-rat-cli/src/cli.rs
+++ b/crates/rag-rat-cli/src/cli.rs
@@ -304,6 +304,8 @@ pub(crate) enum OracleToolArg {
     ScipClang,
     #[value(name = "scip-python")]
     ScipPython,
+    #[value(name = "scip-typescript")]
+    ScipTypescript,
 }
 
 impl OracleToolArg {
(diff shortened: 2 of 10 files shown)


skakri:
Thanks — all five addressed in effa75f.

**1. `--infer-tsconfig` dirties the tree** ✅ Confirmed it writes a `tsconfig.json` into the checkout (v0.4.0). Dropped the flag; scip-typescript now requires a root `tsconfig.json` via `prerequisite_blocked` (the TS analog of scip-clang's `compile_commands.json`). ky ships one, so the corpus is unaffected.

**2. Moniker version churn** ✅ Confirmed empirically — a project at version `9.9.9` yields `scip-typescript npm <name> 9.9.9 …`, and there's no `--project-version` flag. Added `scip::stabilize_moniker_version`, which pins the version component to `_` at the moniker-write site (same stable form scip-python's `--project-version _` produces). My original "commit-stable" claim was wrong; docs/comments corrected.

**3. Health gate isn't a dependency signal for TS** ⏭️ Correct — local monikers come from `package.json` regardless of `node_modules`, so `symbols_with_moniker` doesn't detect missing deps (only external resolution drops). This applies to the runner generally; tracked in #185 (external-resolution health metric) and the comment now says so.

**4. Indexer dep tree not fully pinned** ⏭️ Valid and pre-existing (scip-python's `typescript`/runtime deps resolve via ranges too). Tracked in #185 (shrinkwrap/containerize the indexer or fold the resolved compiler version into tool identity).

**5. Corpus install not locked** ⏭️ Valid and shared with `py-requests`'s `pip install .`. Tracked in #185 (committed lockfile + `npm ci` / pinned constraints).

3–5 are deferred to keep this a focused backend addition; none is a regression here.

@codex review

skakri:
Round 2 addressed in 4eeec7d — all three confirmed empirically against scip-typescript v0.4.0 output before fixing.

**1. UTF-16 ranges** ✅ Confirmed scip-typescript leaves `position_encoding` unset (protobuf default Unspecified) while emitting UTF-16 columns; the reader's Unspecified→UTF-32 fallback misaligned columns past astral chars. Added `OracleTool::default_position_encoding()` (UTF-16 for scip-typescript) + `ScipIndex::parse_with_default`, applied only when a document leaves the field unset — an explicit `position_encoding` still wins. Test: `unspecified_encoding_uses_the_supplied_default` (emoji + identifier).

**2. Synthetic file definitions** ✅ Confirmed the `0..0` per-file definition with a namespace symbol ending in `/` (`…`a.ts`/`, DEF role set). It containment-maps to the first symbol and wins shortest-moniker selection. `scip::symbol_is_module` now filters namespace/module/file symbols out of the moniker pass. Test: `synthetic_file_definition_does_not_overwrite_a_symbols_moniker`.

**3. Arrow-function recall** ⏭️ Confirmed: arrow functions / function-valued vars are `Term` (`…f.`), not `Method` (`…).`), so they're excluded from both sides of the recall ratio (symmetric, narrower; precision unaffected). No clean local fix — SCIP has no call role, and admitting bare `.` re-introduces the field/const over-count the `).` filter exists to prevent. Documented at `symbol_is_callable`; tracked in #185.

Gate: nightly fmt + clippy clean, 526 core tests pass; `small (ts-ky)` health gate green in CI.

@codex review

PR #186, opened 2026-06-16
feat(oracle): C++ corpus (yaml-cpp) + index .h headers under a cpp binding

skakri:
Adds a **C++ corpus** to the oracle report table and fixes the root reason C++ couldn't resolve well: **`.h` headers weren't indexable as C++.**

## The `.h` problem

`.h` is ambiguous (C or C++). Bare detection (`Language::from_path`) resolves it to **C** (the safe default), and the index walk gated target membership on `from_path(path) == target.language` + a glob rendered from `simple_extensions` (which, for `cpp`, is `cc/cpp/cxx/hh/hpp/…` — **not `.h`**). So a `cpp` binding could never index a C++ library's `.h` header API — and most C++ libraries put their API in `.h`. Cross-file calls into those headers resolved to nothing.

## Fix — `.h` resolves as C++ under an explicit `cpp` binding

Centralized in `language.rs` (the single home for extension/glob logic):
- **`target_extensions()`** — what an *explicit* binding claims; `cpp` adds `"h"`. (`simple_extensions` / bare `from_path` unchanged — `.h` still defaults to C with no binding.)
- **`claims_extension` / `claims_path`** and **`default_include_globs()`** — the one source of truth for rendering target filters *and* validating a corpus checkout, so they never drift.
- `config::resolve_targets`, `commands::ensure_checkout_matches_corpus`, `index::walker`, and `index::discovery` all route through these (removed the scattered `simple_extensions().map("**/*.{ext}")` duplication). Files are parsed as the **target's** language, so `.h` under `cpp` → C++, under `c` → C. A `.c` is never C++.

## The corpus

`cpp-yaml` (yaml-cpp 0.8.0, scip-clang, small tier, `bindings = { cpp = ["src", "include"] }`). Binding `include/` exercises the fix — yaml-cpp's classes are declared in `include/yaml-cpp/*.h` and defined in `src/*.cpp`:

| | src only | src + include (this PR) |
|---|---|---|
| files indexed | 31 | **90** |
| resolved_external | 2364 | **72** (header calls now resolve in-corpus) |
| precision | 0.917 | 0.825 |
| recall | 0.95 | 0.402 |
| monikers | 353 | **1038** |

Recall drops because the heuristic's genuine C++ call-coverage gap is now honestly measured against the header API; precision (~0.83) tracks the earlier fmt measurement (the `logical_variant` overload/decl-vs-def tiebreak is the lever). Reuses oracle.yml's existing scip-clang install; auto-added to the small matrix.

## Validation

fmt + clippy clean, 530 core tests + CLI tests (`--features eval`) pass. Validated `cpp-yaml` end-to-end locally (healthy; profile hash pinned in `corpus.rs`).

### DIFF
--- a/crates/rag-rat-cli/src/commands.rs
+++ b/crates/rag-rat-cli/src/commands.rs
@@ -586,7 +586,7 @@ fn ensure_checkout_matches_corpus(
     // form's defaults; anything else measures a different population under the same hash.
     for target in &config.targets {
         let default_include: BTreeSet<String> =
-            target.language.simple_extensions().iter().map(|ext| format!("**/*.{ext}")).collect();
+            target.language.default_include_globs().into_iter().collect();
         let include: BTreeSet<String> = target.include.iter().cloned().collect();
         anyhow::ensure!(
             target.exclude.is_empty() && include == default_include,
--- a/crates/rag-rat-core/src/config.rs
+++ b/crates/rag-rat-core/src/config.rs
@@ -176,6 +176,26 @@ pub struct ResolvedTarget {
     pub kind: TargetKind,
 }
 
+impl ResolvedTarget {
+    /// Indexing precedence when more than one target claims the same file (lower sorts first /
+    /// wins). Two levels: (1) kind — generated/tests/docs before source, the existing order;
+    /// (2) within a kind, a language that claims an ambiguous extension as an upgrade
+    /// ([`Language::upgrades_ambiguous_extension`]) wins it — so a `.h` covered by both a `c` and a
+    /// `cpp` binding indexes as C++ (the deliberate intent), not C (alphabetical accident). Both
+    /// the full-rebuild walk (first-claimer-wins) and the incremental per-path resolver use
+    /// this.
+    pub fn index_precedence(&self) -> (u8, u8) {
+        let kind_rank = match self.kind {
+            TargetKind::Generated => 0,
+            TargetKind::Tests => 1,
+            TargetKind::Docs => 2,
+            TargetKind::Source => 3,
+        };
+        let upgrade_rank = u8::from(!self.language.upgrades_ambiguous_extension());
+        (kind_rank, upgrade_rank)
+    }
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum TargetKind {
     Source,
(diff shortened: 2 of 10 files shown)


skakri:
Both confirmed empirically; addressed in fef11a1.

**1. C++ wins `.h` when C targets overlap** ✅ Confirmed: both `collect_index_files` (first-claimer-wins) and `target_for_path` sorted only by kind, then config order — and simple bindings sort alphabetically (`c` before `cpp`), so C took the shared `.h`. Added `Language::upgrades_ambiguous_extension()` (true for cpp — it claims `.h` as an upgrade over C's default) and `ResolvedTarget::index_precedence()` `(kind, upgrade-claimers-first)`; both paths now sort by it. A `.h` covered by both bindings indexes as **C++**; a `.c` (claimed only by C) is unaffected. Test: `target_for_path` resolves `.h`→cpp / `.c`→c / `.cpp`→cpp with both bindings present.

**2. `init` binds C++ header dirs as `c`** ⏭️ Correct — `init` discovers via `from_path` (every `.h`→C), so a generated config puts `include/` under `c` and the new support is unreachable without hand-editing. This is an `init`-wizard change (detect C++ projects, bind their header dirs as cpp, handle mixed repos) that's separate from the indexing capability here. Filed as **#188**, together with a related gap you flagged — `init` only emits the bare `[target_bindings]` stub, not the full commented config syntax (`[[target]]` blocks, all tables). `docs/config.md` documents how to bind C++ header dirs as `cpp` manually in the meantime.

Gate: fmt + clippy clean, 532 core tests pass.

@codex review

skakri:
Both confirmed and fixed in 71ff70c.

**1. Reindex headers when target language changes** ✅ Real bug — `discovery_plan` compared only path→sha, so an existing index that stored a `.h` as C would never be reindexed as C++ after the upgrade/binding change (sha unchanged), leaving the header upgrade ineffective until `--full`. `indexed_file_map` now also carries the stored `(language, kind)`, and discovery treats a mismatch against the discovered target as a change. Test `discover_relanguages_h_when_binding_changes_c_to_cpp`: a `.h` flips `c`→`cpp` on re-discover with identical content.

**2. Raise cpp-yaml thresholds above the src-only baseline** ✅ Correct — the old gate (`examined ≥ 20`, `monikers ≥ 10`) sat far below even the src-only run, so a regression that stopped indexing `.h` headers would still pass green. Raised to `edges ≥ 7000`, `examined ≥ 4500`, `monikers ≥ 700` — between the src-only (~4600 / ~3400 / ~350) and src+include (~9200 / ~5800 / ~1040) numbers — so the corpus now fails if header resolution regresses. Golden hash re-pinned; re-validated healthy.

Gate: fmt + clippy clean, 533 core tests pass.

@codex review

Issue #2, opened 2026-06-06
M1: Create SQLite schema bootstrap


skakri:
## Goal
Create the initial SQLite schema for files, chunks, symbols, parser failures, and index metadata.

## Acceptance criteria
- [ ] `rag-rat index` creates `.rag-rat/index.sqlite`.
- [ ] Schema creation is idempotent.
- [ ] Storage tests cover bootstrap on an empty repo.
- [ ] The schema supports later FTS5, git, GitHub, and local AI tables.

skakri:
Closed by 747188e and 472d69a. Schema bootstrap is idempotent, existing DBs migrate anchor columns, and index::schema_bootstrap_tests::rebuild_bootstraps_sqlite_schema_for_empty_target_root covers an empty configured target root creating the SQLite schema and FTS table.

Issue #1, opened 2026-06-06
M1: Keep SQLite/rusqlite storage facade and default config


skakri:
## Goal
Keep a slim SQLite/rusqlite-backed storage facade without changing user-visible CLI behavior.

## Acceptance criteria
- [ ] Existing storage callers route through the storage facade.
- [ ] Default index path is `.rag-rat/index.sqlite`.
- [ ] Existing public output structs remain compatible.
- [ ] Heavy embedded database dependencies such as DuckDB are not added.

skakri:
Closed by 4193f6a, 747188e, and 472d69a. Storage is rusqlite-backed with the default .rag-rat/index.sqlite path, DuckDB is absent from dependencies, and doctor now reports the SQLite storage backend explicitly. Public status uses the current SQLite FTS sync fields rather than legacy DuckDB refresh fields.

Issue #191, opened 2026-06-16
Audit resolver overconfidence: assert Exact/Syntactic only when certain (avoid misresolution)


skakri:
## Problem

The heuristic edge resolver asserts `Exact`/`Syntactic` confidence on references it actually *guesses*. Because the SCIP oracle only counts a heuristic edge as `Confirm`/`Contradict` when it's `Exact`/`Syntactic`, every overconfident wrong guess becomes a **contradiction** — which both (a) tanks reported precision and (b) means we ship a *wrong high-confidence edge* that misleads `find_callers`/`trace_callees`/`impact_surface`. A wrong confident edge is worse than an honest unresolved one.

Two overconfidence sources were already fixed (see the PR linking this issue):
1. `same_logical_symbol` grouped distinct same-`(file,name,kind)` items (ignored `scope_path`), so `logical_variant` confidently picked one of N — e.g. serde's 50 `Value` associated types. Fixed by requiring equal `scope_path`.
2. Rust `references_type` to a generic parameter (`T`) or a `Self::`/param projection (`Self::Value`) was bound to a same-named concrete type. Fixed by suppressing those.

Combined: rust-serde precision 0.356 → 0.763, rust-semver 0.796 → 0.831.

## What's still suspected (needs a systematic audit)

Measure with the oracle's **per-`resolution_path` contradiction rate** as ground truth (`edge_oracle.kind='contradict'` grouped by `edges_data.resolution_id`), across all small-tier corpora and every language. Known residual misresolution patterns from the serde/rust-time analysis:

- **Multi-candidate type references** bound at `Syntactic` via `target_name_fallback` / `same_file_name` when N>1 same-named definitions exist (serde `Value`/`Error`/`Deserializer`; time `Second`/`Hour`/`Minute`). Picking the same-file or first candidate is a guess — should it be `NameOnly`/`Ambiguous`?
- **External / prelude / alias shadowing**: a reference to `Result`, `Option`, `fmt::Result`, `crate::Error` binds to a same-named *local* symbol (e.g. time's own `type Result`) when the reference means `std`/another crate. ~100+ contradictions on rust-time. Needs import-aware suppression.
- **`calls_name` trait-method ambiguity**: same-named methods across many impls (serde calls_name precision ~0.46).
- Re-audit the C++/C `logical_variant` path (the prior 29%-precision finding) now that `same_logical_symbol` is stricter.

## Deliverable

[...]

PR #192, opened 2026-06-16
fix(resolve): stop asserting high confidence on guessed Rust type references

skakri:
Fixes two **resolver overconfidence** bugs that made the heuristic assert `Exact`/`Syntactic` on references it was actually guessing — which the SCIP oracle counts as contradictions (and which ship wrong high-confidence edges to `find_callers`/`impact_surface`). Surfaced while normalizing the Rust oracle corpus, where precision was anomalously low (serde 0.36, far below semver's 0.80).

### 1. `same_logical_symbol` ignored `scope_path`
It grouped same-name candidates by `qualified_name` (`{file}::{name}`) + name + kind — not unique within a file. serde's `de/impls.rs` has 31 `Value` rows across 10 `impl Visitor` blocks; they collapsed to "one logical symbol", so `logical_variant` confidently picked one (93% wrong). Now also requires equal `scope_path` (`Visitor::Value` ≠ `IgnoredAny::Value`); genuine variants (decl/def, `#[cfg]` copies) share a scope and stay grouped.

### 2. Rust generic params / associated-type projections
`references_type` to `T` (a generic parameter) or `Self::Value` / `V::Value` (associated-type projections) was bound to a same-named concrete type. These have no in-corpus definition — only type inference knows the target. On serde this was ~640 contradictions (`T` alone: 494). Now suppressed (bare single-uppercase-letter+digits, or a `Self`/param-rooted projection); a module-qualified path like `de::Deserializer` (lowercase root) still resolves.

### Measured precision (oracle, language-agnostic — no regressions)

| corpus | before | after |
|---|---|---|
| rust-serde | 0.356 | **0.763** |
| rust-semver | 0.796 | **0.831** |
| rust-time | 0.625 | 0.687 |
| cpp-yaml | 0.825 | 0.836 |
| py-rich | 0.938 | 0.950 |
| ts-rxjs | 0.871 | 0.873 |

rust-time's residual is corpus-specific (a local `type Result` shadowing `std` + macro-heavy code), not a resolver bug.

### Follow-up
The remaining overconfidence sources (multi-candidate type refs, external/prelude/alias shadowing, `calls_name` trait-method ambiguity) are tracked for a systematic audit in #191.

Both drivers (incremental + full-rebuild) share `resolve_symbol`, so both get the fixes. Tests added for each; fmt + clippy clean, 537 core tests pass.

### DIFF
--- a/crates/rag-rat-core/src/index/edges/resolve/mod.rs
+++ b/crates/rag-rat-core/src/index/edges/resolve/mod.rs
@@ -808,12 +808,14 @@ pub(crate) fn resolve_symbol<'a>(
     let matches = if preferred.is_empty() { matches.as_slice() } else { preferred.as_slice() };
     match matches {
         [symbol] => Some((*symbol, EdgeConfidence::Syntactic, "target_name_fallback")),
-        // Multiple same-name candidates. For a `references_type` this is the associated-type case —
-        // don't guess (see `multi_pick_ok`); leave it unresolved (NameOnly) for the oracle to
-        // upgrade.
-        [_, ..] if !multi_pick_ok => None,
         [_, ..] => {
-            if same_logical_symbol(matches) {
+            // `logical_variant` (an arbitrary pick among same-`scope_path` candidates) is gated by
+            // `multi_pick_ok`: for a Rust `references_type` the same-scope group is distinct
+            // associated types (serde's many `impl Trait { type Value }` share a trait scope), so
+            // picking one is a wrong guess. `same_file_name` is KEPT even for `references_type` — a
+            // type defined AND used in its own file resolves reliably to the local definition even
+            // when the name recurs in other files.
+            if multi_pick_ok && same_logical_symbol(matches) {
                 return Some((matches[0], EdgeConfidence::Syntactic, "logical_variant"));
             }
             let same_file = matches
@@ -823,7 +825,7 @@ pub(crate) fn resolve_symbol<'a>(
                 .collect::<Vec<_>>();
             match same_file.as_slice() {
                 [symbol] => Some((*symbol, EdgeConfidence::Syntactic, "same_file_name")),
-                [_, ..] if same_logical_symbol(&same_file) =>
+                [_, ..] if multi_pick_ok && same_logical_symbol(&same_file) =>
                     Some((same_file[0], EdgeConfidence::Syntactic, "logical_variant")),
                 _ => None,
             }
--- a/crates/rag-rat-core/src/index/edges/resolve/mod.rs
+++ b/crates/rag-rat-core/src/index/edges/resolve/mod.rs
@@ -678,6 +678,18 @@ pub(crate) fn resolve_symbol<'a>(
     {
         return None;
     }
+    // Whether to accept a pick among MULTIPLE same-name candidates (the `logical_variant` /
+    // `same_file_name` arms). A RUST `references_type` with several same-named definitions is
+    // almost always an associated type the reference can't disambiguate by name (serde defines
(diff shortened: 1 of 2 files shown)


skakri:
Good catch — confirmed and fixed in 7ca7af5.

You're right: the extractor emits `Self::Value` / `T::Output` via `last_identifier_text` (`extract.rs:262`) to the **bare** last segment, so those projections arrive as `Value`/`Output` with no root, and the projection guard alone doesn't fire on them (it catches the ~2k edges that *do* carry `::`, plus bare generic params like `T`). I verified against serde: 8674 bare references_type names vs 2035 with `::`, and the dominant bare contradictions are exactly these associated types (`Value`×147, `Deserializer`×38, `Error`×21).

The bare production path is now covered by a separate guard: a Rust `references_type` with **multiple** same-named candidates is not guessed (the `logical_variant`/`same_file_name` arms are skipped — a unique type still binds via the single-match arms; the edge stays `NameOnly` for the oracle to upgrade). Added `references_type_multi_candidate_does_not_guess_in_rust` exercising the bare-`Value` shape (the real extractor output), alongside the existing `::`-form test.

Net on serde: 0.763 → **0.873** precision (recall unchanged). The single remaining gap — a bare projection with exactly ONE same-named concrete type (`Output` where only one `Output` exists) — is indistinguishable from a real type by name and is tracked in #191.

skakri:
Fixed in the latest commit — you're right, the early bail dropped legitimate same-file resolution.

Narrowed it: for a Rust `references_type` I now gate **only** the `logical_variant` arms (the arbitrary pick among same-scope associated types, ~0.25 precision). `same_file_name` and `target_name_fallback` still resolve, so a type defined and used in its own file binds locally even when the name recurs elsewhere. Added `references_type_resolves_same_file_definition_despite_name_collision` for exactly your `struct Error` scenario.

Net: serde 0.763 → 0.806 (down from the 0.873 the over-aggressive bail gave, but no dropped local edges), rust-time 0.687, no regression elsewhere. The `same_file_name` path itself is mixed (0.565 on serde — legit local types vs bare `Self::Value` projections that happen to have one same-file candidate); separating those needs module-scope verification, tracked in #191.

skakri:
Both addressed in df73ca9.

**cfg-split twin types (P2-b)** ✅ You're right — the multi-candidate gate blocked `logical_variant` for `references_type`, which also blocked genuine `#[cfg]`-split twins (same `qualified_name` + `scope_path` = one logical symbol). I reverted that gate entirely: it only helped serde (+0.04, not a report corpus) and added nothing to rust-time, so the cfg-variant regression wasn't worth it. Added `references_type_resolves_cfg_split_twin_types` to lock it in.

**concrete single-letter types (P2-a)** — keeping the suppression, with rationale: a bare single-uppercase-letter type ref (`T`/`V`/`K`, optionally + digits) is overwhelmingly a generic parameter in Rust; a concrete `struct T` is vanishingly rare, and even then a bare `T` reference is genuinely ambiguous with a generic param (no name-based way to tell them apart). The guard degrades such refs to **unresolved (NameOnly), never to a wrong edge** — the oracle upgrades the real ones, and graph consumers see no edge rather than a false one. It fixes 494 generic-param contradictions in serde alone. The residual (a real single-letter concrete type) is noted in the #191 audit. Happy to narrow it further if you feel strongly, but the convention is strong enough that I think unresolved-over-wrong is the right call here.

Net (fix #1 same_logical_symbol scope_path + fix #2 generic-param/projection guard): rust-serde 0.356 → 0.763, rust-semver 0.796 → 0.831, rust-time 0.687; recall unchanged; cpp-yaml/py-rich/ts-rxjs/c-libuv unaffected.

Issue #182, opened 2026-06-16
Python in-corpus module model / external-import suppression


skakri:
## The recurring root gap

Across the Python PRs (#172/#180 implements, #174/#179 from-import aliases) the same fundamental limitation keeps surfacing in review: **rag-rat has no Python external-import suppression / in-corpus module model.** The Rust path has this (`local_crate_roots` / `is_external_import`, #61); Python does not.

Without it, a bare name imported from an EXTERNAL module can bind to a same-named LOCAL symbol. Concrete cases reviewers found:

- `from urllib3.util import Timeout as TimeoutSauce; TimeoutSauce()` → rebound to a local `requests.exceptions.Timeout` (#179; mitigated by restricting alias rebind to **relative** imports).
- `class Sub(pkg.Base)` / `from vendor import Base; class Sub(Base)` where `Base` is external but a local `class Base` exists → implements binds the local class (#180, 772 / round-5).
- `import base; class Sub(base.Base)` where `base` IS in-repo → the qualified form can't resolve without knowing `base` is in-corpus (#180, 728; currently emitted bare).
- `from ..models import X` when the index covers only a subpackage and `models` is outside the indexed targets (#179, 904).

## What would close it

A minimal Python in-corpus module model: resolve an import's module (relative or absolute) to an indexed file (`a.b.c` → a file ending `a/b/c.py` or `a/b/c/__init__.py`), giving an `is_external_python_import` predicate analogous to the Rust one. That lets resolution:
- suppress binding an external imported name to a local same-named symbol, and
- confirm a relative/absolute in-repo import so its alias/base CAN rebind/resolve.

## Why deferred (measured)

On the `py-requests` oracle corpus the impact is small: the alias feature is +3 confirmed / 0 contradicted after the relative-only gate (precision 89.91% → 90.00%); the implements preference is **neutral** (identical to baseline — requests has no same-named base-class collisions). The current PRs are conservative (relative-only aliases, bare implements, allowlist static heads) so they don't produce wrong high-confidence edges. The model is worth building if a heavier Python corpus (more internal hierarchies + external bases) shows the gap materially hurting precision — worth adding such a corpus to `tools/oracle-corpora.toml` first to quantify.

skakri:
Further #179 review (09:19) surfaced more cases that belong to this same gap / the module-scope binding model:

[...]

PR #199, opened 2026-06-17
feat(resolve): Python external-import suppression (#182, conservative)

skakri:
Gives Python the external-import suppression Rust already has (`is_external_import`): a bare reference to a name imported from an **out-of-corpus module** no longer binds to a local same-named symbol. The integration is clean — `resolve_symbol` already suppresses a local bind when `imported_external` is true; this just makes that true for Python.

## How (conservative)

- **Extractor** carries each plain absolute `from M import name`'s module on the per-name edge (`EdgeCandidate.import_source_module`, interned into `CompactEdge`). **Not** aliased imports (the file binds the alias, not the target — avoids the #174 `TimeoutSauce` regression), **not** relative / `import x` (fail-open).
- **`python_module_set`** builds the in-corpus dotted-module set from indexed `.py` paths (`a/b/c.py`→`a.b.c`, `a/b/__init__.py`→`a.b`; root-layout assumption, documented).
- **`is_external_python_module`** classifies external (neither `M` nor `M.name` in-corpus).
- The **full-rebuild resolve driver** populates `ImportScope.python_external`; the existing `imported_external` path suppresses the bind. No `resolve_symbol` change.

## Measured (no regression)

| corpus | contradictions | precision | recall |
|---|---|---|---|
| py-django | 224 → **216** | 93.7 → **93.9** | 84.1 (flat) |
| py-rich | 113 → **109** | 95.0 → **95.1** | 81.3 (flat, confirm identical) |

rust/c/ts corpora are untouched (Python-only). Both Python corpora improve slightly; neither regresses recall or confirm.

## Key finding (redirects the lever to #191)

The diagnostic on py-django shows the **dominant** remaining external contradictions are **not imports** — they're method/dunder over-binding: `get` (42), `update` (9), `filter` (7), `join` (7), `__new__`, `__init__`… bound via `target_name_fallback`/`same_file_name` when the receiver is external. That's **#191 (resolver overconfidence)**, not external-import suppression — so the import gap's real ceiling here is the ~8 cut. The big precision lever is #191 (starting next).

## Scope

This wires the **full-rebuild** path (what `index --full` + the oracle corpus run use). Incremental (DB-driver) persistence of `import_source_module` (a dedicated edge column) is a documented follow-up — fail-open until then (a freshness gap, not corruption). Unit tests cover the module set, classifier, and the `ImportScope` external flag; 552 core tests pass, clippy clean.

Refs #182 (the in-corpus module model is delivered for the full-rebuild path; closes the import-suppression portion).

### DIFF
--- a/crates/rag-rat-core/src/index/edges/extract.rs
+++ b/crates/rag-rat-core/src/index/edges/extract.rs
@@ -76,6 +76,7 @@ pub(crate) fn contains_edges(symbols: &[IndexedSymbol]) -> Vec<EdgeCandidate> {
                 source_span: child.span(),
                 callee_span: None,
                 import_scope: None,
+                import_source_module: None,
                 edge_kind: EdgeKind::Contains,
                 confidence: EdgeConfidence::Exact,
             });
@@ -302,6 +303,7 @@ pub(crate) fn rust_impl_edges(
             // located identifier node, so there is no clean callee range to record (#67).
             callee_span: None,
             import_scope: None,
+            import_source_module: None,
             edge_kind: EdgeKind::Implements,
             confidence: EdgeConfidence::NameOnly,
         });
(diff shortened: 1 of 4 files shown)


skakri:
Closing: the measurement refuted the hypothesis. The external-import suppression is a small win (py-django contradictions 224→216, py-rich 113→109; +0.1-0.2pp precision) AND has three real correctness P2s from review — src-layout wrong-suppression (needs __init__.py package-root detection), no order/rebinding shadowing (needs byte-range scoping like the alias path), and no DB persistence (incremental re-resolve loses it). Meanwhile the diagnostic showed the dominant py-django contradictions are method-name over-binding (get/filter/dunders), and a fix for THAT (#191) measured as a clean win across all corpora (precision up everywhere, recall flat, cpp-yaml -70 contradictions). Shipping #191 instead. #182 stays open with the finding; the import model can be revisited (with package-root/rebinding/persistence) only if a heavier Python corpus shows the import gap is material.

PR #203, opened 2026-06-17
fix(symbol): accept a sym_<hex> handle in the ref slot (#201)

### LINKED ISSUE #201: impact_surface/symbol tools: a sym_<hex> handle passed as `ref` returns null (disambiguation→drill-down handoff)
Reported by a coding agent working in the parent (held) repo.

## Symptom
`impact_surface(symbol=…)` returned `disambiguation_required: true` with candidates carrying `id: sym_<hex>`. The natural next step — feeding the chosen id back as **`ref`** — produced `null`. The agent had to switch to a qualified-name `ref` on a different tool.

## Root cause
The wire (#149) splits the symbol selector: `ref` = the qualified name (`symbol_path`), `id` = the opaque `sym_<hex>` handle (`logical_symbol_id`). The disambiguation output emits the candidate as `id`, but `ref` expects a qualified name — so `ref="sym_abc123"` resolves `symbol_path="sym_abc123"`, matches nothing, and returns null ("not found"). It's a real usability trap: the candidate gives you an `id`, and the obvious retry (`ref`) silently fails.

## Proposed fix
In the shared symbol selector, if `ref`/`symbol_path` starts with `sym_`, route it through `parse_sym_handle` to the `id` path (accept a handle in either slot). Cheap, removes the trap. (Alternatively/additionally: document the drill-down as 'pass the candidate's `id` back as `id`', and/or have the disambiguation payload spell out the exact next-call params.) Applies to every symbol-shaped tool (impact_surface, find_callers, trace_callees, docs_for_symbol, …) since they share the selector.

skakri:
Fixes #201 — reported by a coding agent in the parent repo.

## Bug
`symbol_lookup` (and the other symbol tools) return candidates carrying their handle as `id: sym_<hex>`. The obvious drill-down — feeding that token back as **`ref`** — returned `null`: `ref` maps to `symbol_path`, which resolves against qualified names, so `symbol_path="sym_abc123"` matched nothing. The disambiguation→drill-down handoff silently dead-ended.

## Fix
`SymbolSelector::effective_logical_symbol_id()` = `logical_symbol_id` if set, else a `sym_<hex>` handle parsed out of `symbol_path` (via `parse_sym_handle`). Used by `candidates_for_selector` (routes to the logical-member lookup) and `select_one`'s unambiguous-pick. A real qualified name never parses as a `sym_` handle, so there's no ambiguity — the handle now works in **either** slot (`id` or `ref`) across every symbol-shaped tool (impact_surface, find_callers, trace_callees, docs_for_symbol, memory_for_symbol, …).

## Tests
Unit test covers: explicit `id` wins, a handle in `ref` resolves, a real qualified name falls through to the symbol_path lookup, empty → none. 550 core tests pass; clippy clean.

Closes #201.

### DIFF
--- a/crates/rag-rat-core/src/index/query_api/mod.rs
+++ b/crates/rag-rat-core/src/index/query_api/mod.rs
@@ -594,6 +594,10 @@ mod name_based_tests {
         assert!(!super::selector_is_name_based(&selector(None, Some("sym_zzzz"))));
         // A real qualified name in the same slot stays name-based (heal-eligible).
         assert!(super::selector_is_name_based(&selector(None, Some("crates/x/src/a.rs::foo"))));
+        // A qualified name that merely STARTS with `sym_` (path-qualified, or a `sym_…/` path) is a
+        // name, not a handle — it keeps the #152 heal (P3 review follow-up).
+        assert!(super::selector_is_name_based(&selector(None, Some("sym_helpers.rs::build"))));
+        assert!(super::selector_is_name_based(&selector(None, Some("sym_dir/mod.rs::build"))));
         // A bare name is name-based.
         assert!(super::selector_is_name_based(&selector(Some("foo"), None)));
     }
--- a/crates/rag-rat-core/src/index/query_api/mod.rs
+++ b/crates/rag-rat-core/src/index/query_api/mod.rs
@@ -503,11 +503,14 @@ impl IndexDatabase {
 /// isn't a "just added" symbol, just a stale or wrong id, so re-indexing the change set wouldn't
 /// recover it and would put a `git status` on every such miss.
 fn selector_is_name_based(selector: &crate::query::symbol::SymbolSelector) -> bool {
-    // A `sym_<hex>` handle in the ref/symbol_path slot resolves as a logical id (#201), not a name,
-    // so use the EFFECTIVE id here: a stale/mistyped handle must fail cheaply like `id`, never be
-    // misread as a name/path miss that trips the #152 zero-hit heal + reindex on every bad handle.
+    // A `sym_<hex>` handle in the ref/symbol_path slot is id-based (#201), not a name — exclude
+    // both a handle that RESOLVES and one that's merely handle-SHAPED but malformed (typo/bad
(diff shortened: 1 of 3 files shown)


skakri:
Both P2s addressed in b6ef223:

- **Ref-slot handle heal classification** — `selector_is_name_based` now uses `effective_logical_symbol_id()`, so a stale/mistyped `sym_<hex>` in the `ref` slot fails cheaply like `id` instead of tripping the #152 zero-hit heal/reindex.
- **Disambiguation on ref-slot handle** — `lookup_candidates` no longer flags `disambiguation_required` when the selector resolved via a handle (id or ref-slot `sym_<hex>`); a cfg-split/overload group's members all share that one handle. Mirrors `select_one`.

Added `ref_slot_handle_is_not_name_based` (unit) + a handle-in-ref assertion on the cfg-split integration test. Full core+mcp suites green, clippy clean.

@codex review

skakri:
Follow-up P2 fixed in cccc8b3: a malformed `sym_`-prefixed ref (typo/bad hex/truncated) doesn't parse, so `effective_logical_symbol_id()` returned None and `selector_is_name_based` still let it trip the #152 heal. Added `ref_is_handle_shaped()` (symbol_path starts with `sym_`; a qualified name never does) and excluded it from the name-based check — any handle-shaped ref now fails cheaply like `id`, parse or not. Covered in `ref_slot_handle_is_not_name_based`. Core (551) + mcp (26) green, clippy clean.

@codex review

skakri:
Rebased onto main (post-#204) — resolved the `lookup_candidates` overlap (combines #204's `include_generated` param with the handle-aware disambiguation).

P3 fixed in d473105: `ref_is_handle_shaped` now also requires no `::` and no `/`, so a real qualified name that merely starts with `sym_` (`sym_helpers.rs::build`, `sym_dir/…`) stays name-based and keeps the #152 zero-hit heal; only a bare malformed handle (`sym_zzzz`) is treated as id-based. Covered in `ref_slot_handle_is_not_name_based`. Core (554) + mcp (26) green, clippy clean.

@codex review

PR #204, opened 2026-06-17
feat(symbol): exclude generated bindings from symbol search by default (#202)

### LINKED ISSUE #202: symbol_lookup/impact_surface surface generated bindings (wasm-bindgen/uniffi *.d.ts) as candidates
Reported by a coding agent working in the parent (held) repo.

## Symptom
`symbol_lookup` / `impact_surface` on a Rust symbol surface generated artifacts — `generated*/`, wasm-bindgen / uniffi `*.d.ts` `uniffi_*` consts — as candidates. For source-level navigation that's noise.

## Proposed fix
Default `file_kind != generated` filter on the symbol-candidate path (the index already classifies `files.kind`, incl. `generated` — see ffi_surface's generated detection and the discovery generated classification), with an opt-in flag (e.g. `include: ["generated"]`, matching the existing include-array convention) to surface them when wanted. Keeps source navigation clean by default without losing the ability to inspect generated FFI.

skakri:
## What

`symbol_lookup` (and `db.symbols`) drowned in generated bindings — codegen / ubrn FFI output — that shadow the hand-written source symbol a search is after. The real-world case is codegen living **under a source target** (e.g. `.../src/generated/`): it keeps `kind = source`, gets full symbol extraction, so a `kind != 'generated'` filter could never catch it.

## How

Unify the notion of "generated" on the `files.generated` column the rest of the query layer already filters on (search, orientation, tree, clusters):

- New `crate::index::file_is_generated(kind, path)` = `kind == Generated || is_generated_path(path)`, feeding the `files.generated` column at both insert sites. Codegen under a source target is now flagged `generated = 1` — but **still gets symbols** (extraction is gated on `kind`, untouched), so the graph keeps them for callers/callees while default search/lookup hides them.
- Symbol search filters `files.generated = 0` on the **name + symbol_path** arms only. An explicit `id` / `logical_symbol_id` selection is never filtered.
- MCP opt-in: `symbol_lookup include: ["generated"]` via a dedicated `SymbolInclude { Memories, Generated }` enum (`read_chunk` keeps `MemoriesInclude`).
- `select_one` (impact / graph / memory drill-down) excludes generated; PageRank/importance seeding keeps it (whole-graph nodes).

## Notes

- Requires a **reindex** to take effect on existing DBs (git hooks / watcher handle this).
- A `kind = generated` *target* already produces no symbols (chunk-only), so it never polluted search — the gap was purely the path-heuristic case.

## Test

`symbol_search_excludes_generated_bindings_unless_opted_in` (schema_bootstrap_tests): a `src/generated/` file under a source target is excluded by default, returned with the opt-in, and still resolvable by explicit id. Full core + mcp + cli suites green; clippy clean; nightly fmt.

Closes #202.


skakri:
All three P2s addressed in 0b30794:

- **Backfill existing indexes** — added a `GENERATED_FLAGS_VERSION` gate (mirrors `graph_index_version`). `ensure_generated_flags_current` re-derives `files.generated` from `is_generated_path` over `main.files` on a write-bearing open; read-only opens see the stale version and fall back so they heal once. Full rebuild stamps it current. Bump the version whenever the heuristic's definition changes again.
- **Root-level generated dirs** — `is_generated_path` is now segment-based, matching a `generated`/`generated-web` segment at any depth incl. repo root, plus the `.d.ts`/`_bg.wasm.d.ts` suffixes. Segment equality also drops the `pre-generated-data/` false-positive.
- **Don't advertise an ignored opt-in** — `git_history_for_symbol`/`papertrail_for_symbol` now take a selector-only `SymbolRefArgs` (no `include`); only `symbol_lookup` carries the `generated` opt-in it actually honors.

Tests: `generated_dirs_match_at_any_depth_including_root`, `stale_generated_flags_are_rederived_on_open`, schema assertions that history/papertrail lack `include`, + a `SymbolRefArgs` serde round-trip. Core (552) + mcp (26) + cli green, clippy clean.

@codex review

Issue #200, opened 2026-06-17
Graph misses message-dispatch (actor-channel / enum) edges; completeness_risk reads low anyway


skakri:
Reported by a coding agent working in the parent (held) repo.

## Symptom
`find_callers` on `upsert_journal_embedding` returned **0 callers, completeness_risk: low** — but it genuinely has callers. They reach it through an `MlReq` enum → actor-channel dispatch (a `match` arm in the backend handler), which tree-sitter call-edges can't traverse. The held codebase routes a lot of control flow through `service_channel` / `MlReq`/`DbReq` message enums, so the call graph silently misses those edges.

## Why it matters
Two problems: (1) the dispatch edges are absent, and (2) reporting **low** completeness_risk there is misleading — the agent trusts "0 callers" as complete when it isn't.

## Proposed directions
- **Synthetic dispatch edge:** model "enum variant CONSTRUCTED here → handler `match` arm there" as a synthetic edge for known dispatch enums (a message/command enum matched in a handler). Conservative, lower-bound.
- **At minimum, raise completeness_risk** when a symbol is only reachable via a known dispatch enum (or has zero static callers but is a handler-arm target), so "0 callers" isn't reported as low-risk.

Related but distinct from #119 (effect/capability graph spike) — this is concrete dispatch-edge modeling for the actor pattern. Cross-language (the pattern appears in Rust actors here; also common in TS reducers/Kotlin sealed-class dispatch).

PR #206, opened 2026-06-17
feat(graph): synthesize message-dispatch (actor-channel / enum) edges (#200)

skakri:
## What

Closes the second half of #200. A handler reached only through enum-message dispatch — construct a variant in one function, handle it in a `match` arm elsewhere (the actor-channel pattern) — had no static caller edge to the leaf, so `find_callers` returned 0. tree-sitter sees `sender → constructs Enum` and `handler → calls leaf`, but never `sender → leaf`.

(The first half — `find_callers` 0-callers no longer reads `low` completeness — shipped in #205.)

## How

A synthesized **`dispatches`** edge, formed at resolve time via a global variant-key join:

1. **Extraction (Rust)** records two internal FACT edge kinds:
   - `dispatch_construct` — fn → `Enum::Variant` key, for a tuple `Enum::Variant(..)` or struct `Enum::Variant { .. }` construction.
   - `dispatch_handle` — `Enum::Variant` key (in `evidence`) + handler (in `to_name`), for a `match` arm whose pattern names a variant and whose body calls a handler.
   - Conservative: both path segments must be PascalCase (excludes `Foo::new()`, `T::CONST`); a plain integer/`_` arm yields nothing.
2. **`synthesize_dispatch_edges`** (post-resolve, both drivers) joins constructors to handlers on the variant key and emits a `dispatches` edge from each sender to each resolved handler. Idempotent (drops the active scope's prior `dispatches` first, so incremental re-runs cleanly); fan-out capped per variant to bound a catch-all enum.

## Why it's safe

- The handler name is resolved by the **normal resolver** (the fact's `evidence` is never read by `resolve_symbol` — verified — only by synthesis), so an unresolved handler is never guessed.
- FACT rows carry **no callee range**, so the SCIP oracle (which gates on `callee_start_byte IS NOT NULL`) skips them — oracle metrics unaffected.
- FACT kinds are **excluded from PageRank** (they'd double-count the dispatcher's existing `calls_name`); the synthesized `dispatches` edge IS counted (weight 0.8, just below a direct call).
- It is purely **additive** — a new edge kind, never rewriting `calls_name`; `Syntactic` confidence, double-gated (constructed AND handled).

## Surfacing

`dispatches` rides the call edge-kind set, so `find_callers`/`trace_callees` show it by default; new `McpGraphEdgeKind::Dispatches` on the wire.

## Test

`find_callers_sees_message_dispatch_via_synthetic_edge`: an enum + sender (`enqueue` constructs `MlReq::UpsertJournalEmbedding`) + dispatcher (`handle` matches it → `upsert_journal_embedding`). `find_callers(upsert_journal_embedding)` returns both the direct `handle` (calls_name) and the synthesized `enqueue` (dispatches, evidence = the variant); a non-handler gets no dispatch edge. Core (555) + mcp (27) + cli green, clippy clean, fmt.

## Follow-up

Cross-language (TS discriminated-union reducers, Kotlin sealed-class `when`) reuses the shared synthesis — per-language fact extraction, each gated by corpus measurement. Design: `docs/plans/2026-06-17-200-dispatch-edges.md`.

### DIFF
--- a/crates/rag-rat-core/src/index/edges/extract.rs
+++ b/crates/rag-rat-core/src/index/edges/extract.rs
@@ -243,7 +243,62 @@ pub(crate) fn rust_edges(
                         .map(CalleeRange::of_node),
                 ));
             }
+            // #200 dispatch construct fact: a tuple enum-variant construction `Enum::Variant(..)`.
+            // Key off the FULL call path's last two PascalCase `::` segments, so
+            // `crate::m::Msg::Start(..)` still yields `Msg::Start` (the bare receiver would be
+            // `crate`). `Foo::new()` / `T::CONST` are excluded (tail not PascalCase).
+            if let Some(key) =
+                node.child_by_field_name("function").and_then(|f| enum_variant_key(f, text))
+            {
+                out.push(dispatch_fact(
+                    symbols,
+                    node,
+                    key,
+                    EdgeKind::DispatchConstruct,
+                    EdgeContext::default(),
+                    None,
+                ));
+            }
         },
+        "struct_expression" =>
+        // #200 dispatch construct fact: a struct enum-variant construction `Enum::Variant { .. }`.
+        {
+            if let Some(key) =
+                node.child_by_field_name("name").and_then(|n| enum_variant_key(n, text))
+            {
+                out.push(dispatch_fact(
+                    symbols,
+                    node,
+                    key,
+                    EdgeKind::DispatchConstruct,
+                    EdgeContext::default(),
+                    None,
+                ));
+            }
+        },
+        "scoped_identifier" =>
+        // #200 dispatch construct fact: a UNIT enum-variant `Enum::Stop` in a VALUE position — a
+        // call argument (`send(Msg::Stop)`), a `let` initializer (`let m = Msg::Stop;`), an
+        // assignment RHS, or a `return`/`break` value. Not a `call_expression`, so the arms above
+        // miss it. The value-position gate keeps an ordinary type/module/use path out;
+        // over-emitting for a non-enum `Foo::Bar` is harmless — synthesis only joins a
+        // variant whose head is a unique in-scope `enum`, so a non-enum head never yields a
+        // `dispatches` edge.
+        {
+            if scoped_identifier_in_value_position(node)
+                && let Some(key) = enum_variant_key(node, text)
+            {
+                out.push(dispatch_fact(
+                    symbols,
+                    node,
+                    key,
+                    EdgeKind::DispatchConstruct,
+                    EdgeContext::default(),
+                    None,
+                ));
+            }
+        },
+        "match_arm" => rust_dispatch_handle_facts(text, node, symbols, out),
         "macro_invocation" =>
             if let Some(name) = first_identifier_text(node, text) {
                 out.push(symbol_edge_with_context(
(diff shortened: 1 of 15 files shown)


skakri:
All six P2s addressed in 2fb344b:

- **Cross-enum join** — synthesis now scopes by the enum definition: only joins variants whose enum name resolves to exactly one in-scope `enum` symbol; an ambiguous `Msg` is skipped (no cross-enum edge). Test: `dispatch_join_is_scoped_to_a_unique_enum_definition`.
- **Full-path / unit-variant construct** — keys off the call path's last two PascalCase segments (`crate::m::Msg::Start(..)` → `Msg::Start`); `send(Msg::Stop)` unit variants now produce a construct fact (gated on an `arguments` parent).
- **Arm delegate only** — binds only the arm's tail/delegate call, so `metrics::inc()` / nested arg-calls no longer become handlers. Test asserts an arm side-effect call gets no dispatch caller.
- **Receiver context** — handle facts carry the same `target_qualified_name`/`receiver_hint` as a normal call, so `self.handle_a`/`Self::handle_a` resolve to the right method.
- **Callable preference** — `preferred_matches` treats `dispatch_handle` like `calls_name`.
- **Fan-in exclusion** — fact kinds excluded from `scoped_weighted_fan_in` too (not just PageRank); weight 0 as defense in depth.

Core (556) + mcp (27) + cli green, clippy clean, fmt.

@codex review

skakri:
Round 2 — all four addressed in 36f017b:

- **Guard/scrutinee calls** — the delegate is now found via `collect_tail_calls` (descends if/match/block/return tails), so `Msg::X => if ready() { a() } else { b() }` binds `a`/`b`, never `ready`. Tested: `should_stop()` guard is not a handler.
- **Or-patterns** — `A | B => handle()` emits a handle fact per variant, each anchored at its own variant-pattern node so the span-keyed full-rebuild dedup (which ignores evidence) keeps both. Tested: both `Start` and `Resume` senders dispatch.
- **Unit variants in value positions** — `let m = Msg::Stop; send(m)` (and assignment/return RHS) now produce construct facts, not only direct call args. (Over-emitting for a non-enum head is harmless — the enum-uniqueness gate filters it.)
- **Rust-scoped enum uniqueness** — the parser stores C/C++ `enum_specifier` defs as `kind='enum'` too, so the uniqueness check now filters `language='rust'`; a C++ `enum Msg` no longer suppresses a Rust `Msg` actor enum.

Core (557) + mcp (27) + cli green, clippy clean, fmt.

@codex review

skakri:
Round 3 — all four addressed (still one squashed commit, `1e72631`):

- **`.await` tails** — `collect_tail_calls` now unwraps `await_expression`, so an async `Msg::X => self.handle().await` binds the handler.
- **Nested payload variants** — only the top-level constructor of each OR-pattern alternative is a handled variant; `Outer::Wrapped(Inner::Start) => run()` handles `Outer::Wrapped`, not the payload `Inner::Start` (so an unrelated `Inner::Start` sender isn't a false caller of `run`).
- **Graph index version** — bumped `GRAPH_INDEX_VERSION` 7→8 so an existing v7 index re-extracts the dispatch facts and synthesizes on next open (the read-only gate already keys on it), instead of carrying no `dispatches` edges until a manual rebuild.
- **Aliased enum heads** — the join now skips only a head that names MULTIPLE in-scope Rust enums (ambiguous); a head with zero local definitions (an imported/aliased enum like `use proto::Message as Msg`) is admitted since both sender and handler write the same head. Cross-enum protection unchanged.

New tests: await tail + external-head dispatch; nested-payload variant is not a handled variant. Core (559) + mcp (27) + cli green, clippy clean, fmt.

@codex review

skakri:
Additional review hardening (513f309, still one squashed commit) — three fixes:

- **P1 — `Self::Variant` cross-enum collapse.** `enum_variant_key` accepted `Self::Ripe` and the head `Self` (zero enum defs) was admitted, so two unrelated enums each writing `Self::Ripe` in construct+handle cross-linked (the actor pattern `impl MlReq { fn enqueue() { send(Self::Upsert) } }` is exactly this). Fix: rewrite `Self::V` to the enclosing impl type at extraction — recall preserved, no cross-link. Regression test added.
- **P1 — FACT-row leak into `repo_brief` spine + `repo_clusters`.** Both fan-in queries `JOIN ... edges.to_symbol_id` with no edge-kind allowlist, so resolved `dispatch_handle` facts double-counted a handler's fan-in. Fix: exclude the FACT kinds in both (matching the PageRank/load_bearing exclusions).
- **P2 — `dispatch_construct` resolved to a same-named symbol.** Its synthetic `Enum::Variant` `to_name` was run through the resolver and bound a bogus `to_symbol_id`. Fix: skip resolving construct facts in both drivers (synthesis reads only their `from_symbol_id`).

Verified clean: both-driver idempotency, `files`-scope correctness, span-keyed dedup, fan-out cap, oracle exclusion, and the `GRAPH_INDEX_VERSION` re-extract. Core (560) + mcp (27) + cli green, clippy clean, fmt.

skakri:
Follow-up review found two more leaks of the same class — `repo_brief.rs::file_rows` (fan_in/fan_out CTEs) and `grep_augment.rs::edge_counts` — both confirmed by tests. That's 6 leaky consumers; the inline `NOT IN` approach doesn't scale.

**Rethought it structurally** (326d555). The FACT rows share `edges_data` with real edges, so every reader had to remember an exclusion. Fix: hide the two FACT kinds at the **`edges` compatibility view** — the chokepoint all 11 query-layer readers (traversal, repo_brief, clusters, grep-augment, orientation, impact, memory, …) go through. Now:
- Every view reader is structurally safe — present and future — with no per-consumer change. The two found leaks are fixed *and* the 4 inline patches added earlier (orientation, clusters) are removed.
- Resolution + synthesis read `edges_data` directly, so they still see the facts.
- The two direct `edges_data` importance readers (PageRank, load-bearing fan-in) keep explicit exclusions (small, stable set).
- The synthesized `dispatches` edge is a real edge → stays visible.

Delivered via migration **V023** (recreates the view; baseline covers fresh DBs; existing indexes auto-migrate on open). New regression test `dispatch_fact_rows_are_hidden_from_the_edges_view` asserts the facts are in `edges_data` but absent from `edges`, and `dispatches` stays visible.

Also confirmed clean: incremental idempotency/staleness, multi-worktree scope, GRAPH_INDEX_VERSION, span dedup, and `lexical` (relation_weight 0.0). Residual: a documented P3 (two distinct *external* enums sharing a bare name — accepted recall/safety tradeoff).

Core (561) + mcp (27) + cli green, clippy clean, fmt. Squashed to one commit.

Issue #224, opened 2026-06-18
Index size: the non-text levers (edges, FTS, symbols) — where <1 GB lives (#77 follow-up)


skakri:
Follow-up to #77. Measured composition of the real kernel index (3.6 GB, 1.07M chunks, via `dbstat`) shows **chunk text is only ~31% of the DB**, so #77's text compression alone takes it to ~2.8 GB, not <1 GB. The rest — tracked here — is where the remaining reduction lives.

## Measured composition (kernel index, 3.6 GB)

| object | size |
|---|---|
| `chunks` (table) | 1974 MB (~1114 MB is `text` → #77) |
| `edges_data` | 533 MB |
| edge indexes (`idx_edges_*`, 5×~70 MB) | ~350 MB |
| `chunk_fts_data` | 259 MB |
| `symbols` | 147 MB |
| `logical_symbols` | 117 MB |

Non-text total ≈ **1.7 GB**, untouched by #77.

## Levers (rough priority by size)

1. **Edges (`edges_data` 533 MB + indexes ~350 MB ≈ 880 MB)** — the biggest. #79 already interned the high-repeat strings into `edge_strings`; remaining ideas: the deferred #79 step-3 lazy-materialization of `evidence` (≈40% distinct, kept inline); audit whether all 5 `idx_edges_*` indexes are needed or can be narrowed/merged; tighter column types.
2. **`chunk_fts_data` (259 MB)** — the FTS5 index stores plaintext tokens; text compression does NOT shrink it. Options: tokenizer/prefix-index tuning, or accept it as the search index (and note: if #77 moves to contentless FTS, the external-content overhead changes here).
3. **`symbols` + `logical_symbols` (264 MB)** — consider interning repeated `qualified_name` / `signature` (same pattern as `edge_strings`).
4. **VACUUM is NOT a lever** — measured 3619 → 3578 MB (~1%); the index is already dense (freelist ~1 MB). Don't rely on it for size.

## Notes
- Each lever is independent and can be its own PR; this is the tracking issue.
- Validate every change against the no-indexing-delay / no-retrieval-delay constraint (same bar as #77).

PR #225, opened 2026-06-19
feat(index): dictionary-zstd chunk-text compression + drop chunks.text (#77 Phase 2)

### LINKED ISSUE #77: Compress stored chunk text — dictionary-trained zstd at the blob level (~3.8 GB → <1 GB on the kernel)

skakri:
Implements #77: stored chunk text moves to a dictionary-trained zstd `chunk_text` store, and the inline `chunks.text` column is retired. Phase 1 (#223) removed the `impact_surface` text scan + added `files.has_test_code`; this is the compression itself.

## What changed

**Storage**
- New `chunk_text(chunk_id, blob, raw_len, dict_version)` holds each chunk's text as a dictionary-trained zstd blob; `chunks.text` is dropped (V027). The store is built from `chunks.text` during forward-migration, then the column goes.
- The dictionary is **immutable and versioned** (`chunk_text_dict(version, dict)` + per-blob `dict_version`), not a single mutable global slot. A zstd blob only decodes against the dict it was made with, so the dict is a per-blob decode key. The first index trains version 1; nothing ever replaces it. A future retrain would *add* a version (new blobs use it; old blobs keep theirs; both stay resident), so a rebuild of one worktree context can never orphan another context's blobs. One shared dict (measured within ~3% of per-language).

**FTS**
- `chunk_fts` becomes **contentless** (`content=''`, `contentless_delete=1`) — an external-content index re-reads the content column, which would block dropping it. Tokens are written inline from the in-memory chunk text on every path (full rebuild + incremental); only `MATCH`/`bm25()` are used (snippets come from `chunk_text`), so contentless loses nothing. Requires SQLite ≥ 3.43 (the codebase already required ≥ 3.37 for STRICT / 3.35 for DROP COLUMN).

**Reads**
- A version-aware `ChunkTextDecoder` decodes a batch across resident dict versions, reusing one decompressor per version. Every reader (`read_chunk`, lexical bm25/vector, graph local-context, embedding scan, policy skip-summary, FTS repopulate) decodes from `chunk_text`; the per-call dict load is hoisted to once-per-batch.

**Migrations**: V025 (chunk_text tables), V026 (contentless `chunk_fts`, idempotent), V027 (build store from `chunks.text` then drop it). Forward-migrating an existing index rebuilds the store from the column before dropping it; the column drop is irreversible.

## Benchmark (the gate)

Kernel subtree `mm/fs/kernel/lib` (128,875 chunks), pre-Phase-2 base vs this branch:

| Metric | Before | After |
|---|---|---|
| DB size | 396.0 MB | **329.5 MB** (−17%) |
| Full index time | 150.7 s | **110.6 s** (−27%, faster) |
| chunk_text compression | — | 2.65× (74.3 MB raw → 28.1 MB blobs) |
| Query latency | 1.13 s | 1.02 s |

Indexing got *faster* (contentless FTS skips the bulk external-content `'rebuild'` re-read), and retrieval is correct (snippets decode) and not slower — honoring the no-indexing-delay and no-retrieval-delay constraints.

## Result vs the issue's estimate

The issue estimated `~3.8 GB → <1 GB` assuming chunk text dominates the DB. Measured on the real kernel index, `chunks.text` is ~31% of the DB (not the dominant majority), so blob-level compression lands the full kernel around ~3.6 GB → ~2.8 GB. This PR delivers the compression #77 specifies; the remaining size budget (edge framing, indexes, symbols) is tracked in #224.

## Follow-ups (not in this PR)
- gc sweep of orphan dict versions + an explicit dict-retrain op — only relevant once a retrain creates a second version.

Closes #77.

https://claude.ai/code/session_01Qj5oCLLt7UcjBv2Ckr3gc8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -251,6 +251,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "556e016178bb5662a08681bbe0f00f8e17631781a4dfc8c45e466e4b185ec27f"
 dependencies = [
  "find-msvc-tools",
+ "jobserver",
+ "libc",
  "shlex",
 ]
 
@@ -2487,6 +2489,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "jobserver"
+version = "0.1.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9afb3de4395d6b3e67a780b6de64b51c978ecf11cb9a462c66be7d4ca9039d33"
+dependencies = [
+ "getrandom 0.3.4",
+ "libc",
+]
+
+[[package]]
 name = "js-sys"
 version = "0.3.99"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3190,6 +3202,7 @@ dependencies = [
  "tree-sitter-rust",
  "tree-sitter-typescript",
  "ureq 2.12.1",
+ "zstd",
 ]
 
 [[package]]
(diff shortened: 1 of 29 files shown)


skakri:
Also folds in the fix for the schema-upgrade self-deadlock (#226): the open-time `Older`→migrate now acquires through a reentrant `locks::WriteLock`, so a CLI write command / git-hook `maintenance` / watcher pass that already holds the per-DB write lock can migrate under it instead of timing out 30s. This matters precisely because Phase 2's 23→27 schema bump makes every existing index `Older` on the first open after upgrade. Commit 05f6cf5.

Issue #215, opened 2026-06-18
duplication/centralization: semantic (embedding-cosine) + graph-pattern/callee-multiset signals (structural clones shipped)


skakri:
## Idea
We index everything — use it to find **repeated patterns = candidates for refactoring/centralization**.

## What exists vs the gap
- Present: `chunk_embeddings` (per-chunk vectors, the cosine substrate behind `semantic_search`); the call graph (`edges_data`); `important_symbols` (to rank which duplicates are worth centralizing).
- NOT this: `repo_clusters` is *ownership* clustering (path proximity + graph edges + co-change), not code-similarity.
- Gap: no near-duplicate / clone-detection pass.

## Proposed design — two complementary signals
1. **Copy-paste clones (Type-1/2) — token-shingle MinHash + LSH.** Normalize each symbol body (tree-sitter already parses it: strip comments/whitespace, optionally canonicalize identifiers/literals), k-shingle -> MinHash signature -> LSH buckets -> candidate pairs above a Jaccard threshold. Embedder-independent, deterministic, high-confidence. **Ship first.**
2. **Semantic clones (Type-3/4) — embedding cosine clustering** over `chunk_embeddings` (bucketed/ANN, or all-pairs within MinHash buckets). Catches "same logic, different shape." Quality depends on the embedder — gate on a real one (fastembed/e5); the default hash embedder is weak for semantics.
3. *(optional)* **Graph-pattern duplication** — functions with matching callee multiset/sequence = repeated orchestration worth a shared helper; cheap over `edges_data`.

## Output / ranking
Clone classes ranked by **centralization ROI ~= avg clone size x number of copies x cross-module spread** (same block across 5 modules >> within one file). Each class: member locations, representative snippet, similarity scores, an `important_symbols`-derived load-bearing flag, optional extraction-site hint.

## Build
- CLI `rag-rat duplicates [--min-tokens N] [--min-copies 2] [--semantic]`; MCP `duplication_candidates`.
- Run on-demand (like `oracle`), not every index; cache a `clone_classes` table if it becomes routine.

## Caveats
Exclude generated/test paths (don't flag generated bindings/fixtures); language-aware normalization (start language-agnostic); calibrate Jaccard/cosine thresholds on this repo (expose as flags).

skakri:
## Design expansion — from "similarity score" to "actionable refactor"

[...]

PR #229, opened 2026-06-19
feat(index): clone-detection fingerprint substrate — SourcererCC inverted index (#215)

skakri:
Plan 1 of #215 — the **clone-detection fingerprint substrate**, using a SourcererCC-style normalized-token inverted index. Foundation only; the query surface, SCIP secondary fingerprint, and anti-unification refinement are Plans 2–4.

## What this adds (`index/clones/`)
- **`normalize_baseline`** — per-symbol AST walk → normalized token sequence (identifiers alpha-renamed by first occurrence, literals bucketed by category, structure kept verbatim) so renamed/Type-2 clones collapse to one shape.
- **`tokens`** — FNV-1a **token bag** (`token_hash → freq` multiset) + **`struct_hash`** (the exact Type-1/2 fast path); both deterministic and machine-stable.
- **`store_symbol_fingerprints`** — writes `symbol_fingerprints` + the `symbol_token_postings` inverted index + bumps `clone_token_df`, on both the incremental/heal and full-rebuild paths (the full rebuild refreshes df authoritatively).
- **`candidate_clone_components`** — the SourcererCC candidate read: `struct_hash` fast path, a rarest-token **sub-block filter** over a deterministic `(df, token_hash)` order, and an **exact max-denominator overlap verify** (`overlap / max_len ≥ θ`), then union-find — all over the scoped `files` view.

**Schema** (V029, additive, STRICT): `symbol_fingerprints`, `symbol_token_postings`, `clone_token_df`, `clone_refinements`.

## Invariants
- **Scope-independence** — fingerprints/postings key by `symbol_id` (FK `ON DELETE CASCADE`), no scope columns; clustering is the only scope-dependent step (query-time over the scoped `files` view). Avoids the per-scope cache leak the importance cache had to work around.
- **Baseline-only recall** — only baseline-normalizer postings drive candidate recall, so it never depends on oracle/SCIP coverage; SCIP is a later secondary signal.
- **df is selectivity-only** — admissibility comes from the shared deterministic token order + exact verify, never df accuracy; a missing/stale df can only slow candidate generation, never miss a clone.
- **Max-denominator gate** — `overlap/max` (not `overlap/min`, which is containment) keeps a small helper contained in a large function from being a false whole-symbol clone.

## Tests
655/655 (`cargo nextest run -p rag-rat-core`). Coverage: renamed-clone matching, literal bucketing, deterministic token bags, FK-cascade reindex freshness, scope-correct components, the containment-rejection adversarial case (20-token body inside a 200-token one is **not** a clone), and df-empty recall parity.

## Design lineage
Inspired by SourcererCC (token-bag candidate generation), NiCad (normalized near-miss framing), GumTree (move-aware AST diff, for the Plan-4 refinement), anti-unification / least-general generalization (template extraction), and CP-Miner (planned fragment-level mining + copy-paste-bug heuristics). See the README "Prior art" section.

## Follow-up (non-blocking)
#230 — the full-rebuild path re-reads + re-parses the file to fingerprint; move it into the prepare phase (drops the redundant work + a narrow TOCTOU).

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,3 +202,10 @@ and uses `gh api`; normal query tools read only the local cache.
 
 Releases are automated by [release-plz](https://release-plz.dev) (the three crates ship in lockstep;
 see [`docs/releasing.md`](docs/releasing.md)). `rag-rat` is MIT-licensed — see [LICENSE](LICENSE).
+
+## Prior art
+
+rag-rat's clone-detection design is inspired by SourcererCC's scalable token-bag candidate
+generation, NiCad's normalized near-miss clone-detection framing, GumTree's move-aware AST
+differencing, and anti-unification / least-general generalization for template extraction. Planned
+fragment-level mining and copy-paste bug heuristics are inspired by CP-Miner.
--- a/crates/rag-rat-core/src/index/clones/mod.rs
+++ b/crates/rag-rat-core/src/index/clones/mod.rs
@@ -64,6 +64,32 @@ pub(crate) fn fingerprint_symbol(node: Node<'_>, text: &str) -> Option<SymbolFin
     })
 }
 
+/// Baseline fingerprints for a file's function symbols, walking the SHARED parse tree (no re-parse,
+/// no DB). Returns `(local_symbol_index, fingerprint)` pairs keyed by index into `symbols`, so the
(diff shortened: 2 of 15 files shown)


skakri:
Follow-up tracked in #230 (full-rebuild re-read + TOCTOU → fingerprint in the prepare phase).

PR #244, opened 2026-06-21
Cross-platform support (macOS + Windows)

### LINKED ISSUE #237: Enable rusqlite `bundled` feature — Windows compile/link blocker
**Severity: blocker (Windows).** Single make-or-break item for a Windows build.

## Problem
`rusqlite = "0.40"` (`Cargo.toml:35`) is declared with no features, so `libsqlite3-sys 0.38.1` links the **system** SQLite. `Cargo.lock` confirms its only build deps are `pkg-config` + `vcpkg` (the system-library resolvers) — **not `cc`** — which proves `bundled` is off.

- Linux ✅ (`libsqlite3-dev` present in CI/bench images)
- macOS ✅ (system `libsqlite3` ships in the SDK)
- **Windows ❌** — no system sqlite by default → link failure (`unable to find library -lsqlite3`) unless a vcpkg sqlite3 is pre-installed.

The binary genuinely links sqlite C symbols (`crates/rag-rat-core/src/index/mem_diag.rs`, `crates/rag-rat-core/src/storage.rs::sqlite_version()`), so it is not eliminable dead code.

## Fix
```toml
rusqlite = { version = "0.40", features = ["bundled"] }
```
Compiles SQLite from the vendored amalgamation via `cc` on every OS. No new toolchain needed (a C compiler is already required by zstd-sys / tree-sitter / ring). Also lets `tools/bench.Containerfile` drop its `libsqlite3-dev` apt line.

## Acceptance
- [ ] `rusqlite` declared with `bundled`
- [ ] `tools/bench.Containerfile` no longer installs `libsqlite3-dev` (optional cleanup)
- [ ] Verified green by a Windows CI leg

Cost: added C-compile time on cold builds + larger binary (mind the binary-size budget in `docs/`).

### LINKED ISSUE #238: Add macOS + Windows CI matrix (test / clippy / build-full)
**Severity: major (tooling).** CI is 100% `ubuntu-latest` across all 9 workflows — macOS/Windows are never compiled or tested, so any platform regression ships undetected to `cargo install` users.

## Change
Add an OS matrix to the `test`, `clippy`, and `build-full` jobs in `.github/workflows/ci.yml`:
```yaml
strategy:
  matrix:
    os: [ubuntu-latest, macos-latest, windows-latest]
runs-on: ${{ matrix.os }}
```

Keep **single-OS (Linux)**:
- `fmt` (nightly rustfmt, platform-agnostic)
- `coverage` (llvm-cov — one OS is enough)
- all `bench*` / `oracle*` workflows (iai-callgrind needs valgrind = Linux-only; testbeds are Linux-pinned for comparability)

## Notes
- Depends on #237 (the Windows `test`/`build-full` legs won't link until `rusqlite` is bundled).
- `build-full` is the load-bearing leg — it exercises the default-feature `fastembed`/`ort` (ONNX Runtime) download+link path where a real mac/win break would surface.
- `--no-default-features` keeps the `test`/`clippy` legs network-free (hash embedder, no ORT download).
- Windows runners ship MSVC; macOS runners ship Xcode CLT — the C toolchain (zstd-sys, tree-sitter grammars, ring) is satisfied out of the box.

## Acceptance
- [ ] test/clippy/build-full run on ubuntu + macos + windows
- [ ] fmt/coverage/bench/oracle remain Linux-only
- [ ] All three OSes green

### LINKED ISSUE #239: Resolve home dir portably ($HOME → %USERPROFILE%) for --global hooks + model cache
**Severity: major runtime (Windows).** Builds fine; a real CLI command fails at runtime on Windows.

## Problem
Two sites read `$HOME` directly, which is conventionally unset on Windows (it uses `%USERPROFILE%` / `%LOCALAPPDATA%`):

1. `crates/rag-rat-cli/src/claude_settings.rs:193` — `settings_path(global=true)` does
   `std::env::var_os("HOME").ok_or_else(|| anyhow!("HOME not set"))?`, so
   **`rag-rat hooks install --global`** hard-errors `"HOME not set"` on stock cmd/PowerShell
   (works under Git-Bash, which sets HOME). Project-scoped install is unaffected.
2. `crates/rag-rat-core/src/index/ai/helpers.rs` — `fastembed_cache_dir()` falls back from
   `XDG_CACHE_HOME` → `HOME/.cache` → repo-relative `./.rag-rat/models`. On Windows it lands in a
   per-checkout `.rag-rat/models` (re-downloads per repo, pollutes the tree). Only affects the
   `fastembed` feature.

## Fix
Minimal, zero-dep: `var_os("HOME").or_else(|| var_os("USERPROFILE"))` for the settings path; add a
`LOCALAPPDATA` branch for the cache dir. Or adopt the `dirs` crate (`home_dir()` / `cache_dir()`),
which already does the per-OS right thing and preserves XDG on unix. (`dirs` is present transitively
in `Cargo.lock` but is not a direct dep of `rag-rat-cli`.)

## Acceptance
- [ ] `hooks install --global` works on a stock Windows shell
- [ ] model cache resolves to a per-user dir on Windows, not repo-relative
- [ ] unix XDG/HOME behavior unchanged

### LINKED ISSUE #240: Move CLI `libc` dep under [target.'cfg(unix)'.dependencies] (hygiene)
**Severity: minor (hygiene).** Not a build failure — cleanup for parity.

`crates/rag-rat-cli/Cargo.toml:23` declares `libc.workspace = true` unconditionally, while
`rag-rat-core` and `rag-rat-mcp` correctly gate it under `[target.'cfg(unix)'.dependencies]`. Every
CLI `libc::` use lives in `init/mod.rs` under `#[cfg(unix)]`, so on Windows `libc` is a dead
dependency (it still compiles — thin platform-types crate — so no error, and there is no
`deny(unused_crate_dependencies)` lint).

## Fix
Move the CLI's `libc` line into a `[target.'cfg(unix)'.dependencies]` table to match the other two
crates, dropping an unused dep on Windows.

## Acceptance
- [ ] CLI `libc` is cfg(unix)-gated
- [ ] Workspace builds on all targets

### LINKED ISSUE #241: Harden chunker byte offsets for CRLF line endings (latent)
**Severity: minor (latent data hygiene, Windows).** Builds and works today; worth hardening.

## Problem
`crates/rag-rat-core/src/index/chunker.rs` advances byte offsets with `byte += line.len() + 1`
(lines ~60, 211, 241, 323) while iterating `text.lines()`, which strips both `\n` and `\r\n`. On a
CRLF checkout (git `autocrlf=true`, the Windows default) every terminator is 2 bytes, so stored
`chunks.start_byte`/`end_byte` under-count by 1 per preceding line.

## Why it is latent, not a returned-data bug
Verified: `start_byte`/`end_byte` are **written but never read back for text resolution**.
`read_chunk` and anchor validation are line-number + content-hash based (`query/mod.rs`,
`index/anchors.rs`), and the only `start_byte`/`end_byte` consumers are the SCIP-oracle / clones
paths, which use fresh tree-sitter/SCIP ranges, not the chunker arithmetic. So this does **not**
mislocate returned chunks — it is off-by-N values in an effectively-unused stored column on CRLF
checkouts.

## Fix (future-proofing)
Compute offsets from real terminator lengths (e.g. `split_inclusive('\n')` and use each slice's byte
length) or normalize CRLF→LF on read before chunking. Add a CRLF fixture test asserting chunk byte
ranges match the on-disk bytes.

## Acceptance
- [ ] chunk byte offsets are correct on CRLF input
- [ ] CRLF fixture test added

### LINKED ISSUE #242: Docs: add platform-support note once mac/Windows CI is green
**Severity: minor (docs).** Gated on #237 and the CI-matrix issue in this milestone.

The README install section is OS-agnostic in wording (`cargo install rag-rat`) but states it "links
against the system SQLite via rusqlite" with no Windows caveat, and the watcher note frames
freshness around Linux/WSL mounts. No doc states macOS/Windows is supported or tested. Since CI never
builds mac/win today, support cannot be honestly claimed yet.

## Once the compile fix + CI matrix land
- Add a short **Platform support** note: Linux / macOS / Windows tested in CI.
- Document the C-toolchain prerequisite (macOS: Xcode Command Line Tools; Windows: MSVC Build Tools).
- Drop the "links against the system SQLite" caveat if `bundled` is adopted (#237).
- Tie the `cargo install` promise to the now-green mac/win CI.

## Known runtime gaps to document (not bugs — by design)
- SIGUSR1 hot-upgrade is unavailable on Windows (restart the MCP server to pick up a new binary).
- The fleet auto-upgrade trigger is Linux-only (walks `/proc`) — a no-op on macOS and Windows.
- The grep-augment hook uses a per-call direct DB read on Windows (no warm Unix-socket listener);
  full listener works on macOS.

## Acceptance
- [ ] Platform-support section added after #237 + CI matrix are green

skakri:
Resolves the **Cross-platform support (macOS + Windows)** milestone.

## What changed

| # | Area | Change |
|---|------|--------|
| #237 | deps | `rusqlite` declared with the `bundled` feature — libsqlite3-sys compiles the vendored SQLite amalgamation via `cc` on every OS, so Windows links without a system SQLite. `Cargo.lock` gains the `cc` build edge; `bench.Containerfile` drops `libsqlite3-dev`. |
| #240 | cli | `libc` moved to `[target.'cfg(unix)'.dependencies]` (matches core/mcp) — dead dep dropped on Windows. |
| #239 | runtime | `settings_path` falls back `$HOME → %USERPROFILE%` (`hooks install --global` works on stock Windows shells); `fastembed_cache_dir` gains a `#[cfg(windows)]` `%LOCALAPPDATA%` branch (per-user model cache). Unix XDG/HOME precedence unchanged. |
| #241 | index | Chunker byte offsets are CRLF-correct: `split_inclusive('\n')` + raw-slice advance across all sites, and `line_span` returns the on-disk span so the tree-sitter code path's `end_byte` stays byte-aligned with the symbol offsets `query::graph_meta` joins against. New `chunker_tests` assert exact on-disk boundaries on the multi-chunk path. |
| #238 | ci | New `ci-cross-platform.yml` runs clippy/test/build-full on macOS + Windows, gated to `release: published` (+ `workflow_dispatch`). `ci.yml` stays Linux-only on every PR/push, so per-commit cost is unchanged. |
| #242 | docs | README **Platform support** section: per-OS C-toolchain prereq, dropped stale "system SQLite" caveat, documented the Unix/Linux-only maintenance conveniences. |

## CI cadence (per request)
macOS/Windows are **release-gated**, not run on every PR — they fire on the same `release: published` trigger `oracle.yml`/`bench-release.yml` already use. Linux keeps full per-PR/main coverage in `ci.yml`.

## Verification
`cargo +nightly fmt --check`, `cargo clippy --workspace --all-targets --no-default-features --locked -D warnings`, `cargo build --workspace --no-default-features --locked`, and `cargo nextest run --workspace --no-default-features --locked` all green — **870/870 tests** (6 new CRLF chunker tests). Verified on Linux; the macOS/Windows legs will run on the next release.

> Note: #237 makes two repo memories ("rusqlite is non-bundled" / "must apt-get libsqlite3-dev") stale once this merges — they should be updated/obsoleted post-merge.

Fixes #237
Fixes #238
Fixes #239
Fixes #240
Fixes #241
Fixes #242

### DIFF
--- a/.github/workflows/ci-cross-platform.yml
+++ b/.github/workflows/ci-cross-platform.yml
@@ -1,0 +1,98 @@
+name: CI (cross-platform)
+
+# macOS + Windows coverage for the test/clippy/build-full legs. These platforms are NOT exercised on
+# every PR/push (that would multiply CI minutes for the rare platform regression); the per-PR + main
+# matrix in ci.yml stays Linux-only. Instead this runs only when a release is published — the same
+# `release: published` trigger oracle.yml and bench-release.yml already key off (release-plz cuts the
+# GitHub release, authored with RELEASE_PLZ_TOKEN so downstream workflows fire, once the Release PR's
+# version commit lands on main). workflow_dispatch is kept so the matrix can be validated manually.
+#
+# Jobs mirror ci.yml's Linux legs exactly (toolchain, rust-cache, feature flags) so a real mac/win
+# break surfaces the same way it would on Linux:
+#   - clippy / test run --no-default-features (hash embedder, no ONNX/model download — network-free);
+#   - build-full compiles the default feature set (fastembed + model2vec / ort), the load-bearing leg
+#     where the ONNX Runtime download+link path would actually break on a new platform.
+#
+# NOTE: the Windows legs depend on #237 (rusqlite must be `bundled` so it links without a system
+# SQLite). With that landed, windows-latest links the vendored SQLite amalgamation via cc.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+# Least privilege: CI only reads the repo.
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-cross-platform-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+
+jobs:
+  clippy:
+    name: clippy (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+      - uses: Swatinem/rust-cache@v2
+      # --no-default-features = hash-only build: no ONNX/model downloads, fast and network-free.
+      # The full feature set is compile-checked by the build-full job below.
+      - run: cargo clippy --workspace --all-targets --no-default-features --locked -- -D warnings
+
+  test:
+    name: test (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+      - uses: taiki-e/install-action@nextest
+      # Hash embedder only (no model downloads). Mirrors ci.yml's test leg; the Codecov upload is
+      # intentionally dropped here (coverage / test-analytics stay Linux-only).
+      - name: Run tests (nextest)
+        # --no-fail-fast: this matrix runs only on release (rare), so surface EVERY platform-specific
+        # failure in a single run rather than stopping at the first (the ci profile fails fast).
+        run: cargo nextest run --workspace --no-default-features --locked --profile ci --no-fail-fast
+
+  build-full:
+    name: build (fastembed + model2vec) (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+      # Align the MSVC C runtime so the link succeeds. esaxx-rs (a C++ dep via
+      # tokenizers <- model2vec/fastembed) hardcodes cc's .static_crt(true) => /MT, but the ort_sys
+      # ONNX Runtime prebuilt is /MD; the mismatch is LNK2038/LNK1319 on Windows. cc appends env
+      # CFLAGS/CXXFLAGS AFTER its own /MT and cl.exe is last-flag-wins, so /MD here overrides
+      # esaxx's /MT to match ort. Windows-only: on macOS the esaxx build is clang, where -MD means
+      # "emit make dependencies" and would break the build.
+      - name: Align MSVC CRT to dynamic (/MD) to match the ort prebuilt
+        if: runner.os == 'Windows'
+        shell: bash
+        run: |
+          echo "CFLAGS=-MD" >> "$GITHUB_ENV"
+          echo "CXXFLAGS=-MD" >> "$GITHUB_ENV"
+      # Compile the default feature set (downloads the ONNX runtime once, then cached) so the
+      # released binary's embedding paths never break silently on mac/Windows.
+      - run: cargo build --workspace --locked
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,6 +32,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloca"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5a7d05ea6aea7e9e64d25b9156ba2fee3fdd659e34e41063cd2fc7cd020d7f4"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -178,6 +187,12 @@ checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
 
 [[package]]
 name = "bitflags"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "bitflags"
 version = "2.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
@@ -193,9 +208,9 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
-version = "0.12.0"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cdd35008169921d80bc60d3d0ab416eecb028c4cd653352907921d95084790be"
+checksum = "d2f6c7dbe95a6ed67ad9f18e57daf93a2f034c524b99fd2b76d18fdfeb6660aa"
 dependencies = [
  "hybrid-array",
 ]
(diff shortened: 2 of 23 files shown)


skakri:
### Cross-platform CI validated green ✅

The macOS + Windows matrix was run on this PR (via a temporary `pull_request` trigger, since `workflow_dispatch` can't target a workflow that isn't yet on the default branch). All six legs pass:

| | clippy | test | build (fastembed+model2vec) |
|---|---|---|---|
| **macOS** | ✅ | ✅ | ✅ |
| **Windows** | ✅ | ✅ | ✅ |

The temporary trigger has been removed — the workflow is now `release: published` + `workflow_dispatch` only, as designed, with `--no-fail-fast` so a release run reports every platform failure at once.

#### Latent cross-platform issues the new matrix surfaced (and this PR fixes)
None were caused by the milestone changes — they were pre-existing, invisible while CI was Linux-only (exactly what #238 is for):

- **Windows ML link** — `esaxx-rs` (`/MT`) vs ONNX Runtime prebuilt (`/MD`) CRT mismatch (`LNK2038/LNK1319`); fixed by forcing `CFLAGS/CXXFLAGS=-MD` on the Windows `build-full` leg.
- **Windows verbatim-prefix paths** — `ignore_rules`/`watch` compared a canonicalized `config.root` (`\\?\C:\…`) against gix's plain `C:\…` workdir; fixed with a shared canonicalization-aware `base_under_worktree` helper (real production bug — worktree-root `.gitignore` was dropped for subdir configs).
- **Windows test db-lock** — ~210 tests called `fs::remove_dir_all(root).unwrap()` with the SQLite db still open (Windows can't delete open files); made cleanup best-effort.
- **macOS `/var`→`/private/var`** — temp-dir symlink broke the subdir-overlay + socket-path tests; canonicalize / use a short fixed base.
- **Dead-code on non-unix** — `fleet`/`claude_hook` unix-only items now `cfg`-gated.

All green locally too: `cargo +nightly fmt --check`, `clippy --all-targets --no-default-features -D warnings`, `nextest --workspace` (870/870).

skakri:
### Codex review findings — all addressed

- **P1 — bundled SQLite vs MSRV (`Cargo.toml`)** — ✅ fixed in `2d1ccf8`. Confirmed locally: `cargo +1.89 check` fails with `E0658: use of unstable library feature cfg_select` in `libsqlite3-sys 0.38.1`'s build script. `cfg_select!` stabilized in **Rust 1.95** (verified: 1.94 fails, 1.95 OK), and `rusqlite 0.40` pins `libsqlite3-sys ^0.38.1` so there's no `cfg_select`-free bundled path without downgrading rusqlite. Bumped the workspace `rust-version` to `1.95` (the full default-feature workspace — fastembed + model2vec + ONNX + bundled SQLite — checks clean on 1.95.0) and documented it in the README platform section.

- **P1 — gate Unix-only imports for Windows clippy (`claude_hook/mod.rs`)** — ✅ already fixed in `c8c7375` (the comment was on the earlier `71faca3`). `PathBuf`/`Duration` are now `#[cfg(unix)]`-gated alongside `socket_path`/`SOCKET_BUDGET`; the Windows clippy leg passes.

- **P2 — remove the temporary PR trigger (`ci-cross-platform.yml`)** — ✅ already removed in `d0c7c0c` (the comment was on the earlier `2f9f617`). The workflow is back to `release: published` + `workflow_dispatch`. (It was re-added once more to re-validate a follow-up fix on Windows, then removed again.)

The full mac/Windows matrix was validated green on this branch throughout; this push re-runs the Linux PR CI + benchmarks.

skakri:
### Review comments — status (post-squash, `b912022`)

**Codex findings**

1. **P1 bundled SQLite vs MSRV** (`Cargo.toml`) — resolved. `rust-version = 1.95` (verified: 1.94 fails the `cfg_select` build, 1.95 OK), and bundled is now **Windows-only**, so Linux/macOS don't compile `cfg_select` at all (their dep floor is 1.88).
2. **P2 remove temp PR trigger** (`ci-cross-platform.yml`) — done; the workflow is back to `release: published` + `workflow_dispatch` only (the squash folded out the temporary trigger I used to re-validate the matrix).
3. **P1 gate Unix-only imports for Windows clippy** (`claude_hook/mod.rs`) — done; `PathBuf`/`Duration` are `#[cfg(unix)]`-gated. The Windows clippy leg passed green.
4. **P1 Unix-only benches break Windows clippy** (`ci-cross-platform.yml`) — **not reproduced.** The `clippy (windows-latest)` leg (`cargo clippy --workspace --all-targets --no-default-features --locked -- -D warnings`) **passed green** on this exact code. `iai-callgrind` 0.16 compiles its bench targets on Windows, so `--all-targets` is fine there. No change.

**Bencher alert (`index_time/full_rebuild_cargo` +63%) — false alarm, stale baseline.**
The 4.27s baseline is frozen from before the recent #215 clone-detection indexing work raised main's full-rebuild time. Decisive local A/B (same machine, both **system** SQLite, clean tmpfs):

| | full_rebuild_cargo |
|---|---|
| main `838cd9e` | 7.04s (6.90–7.23) |
| this branch `b912022` | 6.31s (6.18–6.46) |

The branch is **~10% faster than current main** — no regression from this PR. (Bundled vs system SQLite is also a wash, ~6–7s either way.) The Bencher baseline should be refreshed against current main; separately, #215 looks to have slowed main's full-rebuild ~64% vs the frozen baseline, which is worth a look on its own.

Note: the `bench-pr-track` job failed (a force-push race during the squash), which is why the Bencher comment above still shows the pre-squash number.

Issue #248, opened 2026-06-21
edge_oracle is wiped on every reindex (CASCADE FK + edge_id rowid key) and never repopulated (opt-in oracle)


skakri:
## Bug
`edge_oracle` holds the per-edge SCIP resolution (resolved_symbol_id, scip_symbol) written by `oracle run` (stats show `rows_written: 41167` on a real run). But the live index has **0 edge_oracle rows** despite the oracle having run — the per-edge resolution is lost.

## Root cause
`edge_oracle` is keyed by the volatile `edge_id` rowid with `FOREIGN KEY(edge_id) REFERENCES edges_data(id) ON DELETE CASCADE` (migrations.rs:588-604). Every reindex rewrites `edges_data` (full rebuild `graph_index.rs:258`; per-file `file_rows.rs:64`/remove_file_in_scope), so the CASCADE wipes all verdicts. The DDL comment assumes reindex "recompute[s] the oracle" — but the oracle is **opt-in** (`auto_run=false`), so it never repopulates. Result: edge_oracle dies on the first reindex after an oracle run and stays empty.

By contrast `logical_symbol_monikers` was deliberately made to survive reindex: no FK, content-derived `logical_symbols.id`, reads join live symbols so dangling rows never resolve (migrations.rs:615-636). `edge_oracle` never got that treatment — the two oracle outputs have inconsistent lifecycle.

## Impact
Anything reading edge_oracle is dormant on real indexes: `compare_graph_to_scip`, oracle status/eval coverage metrics, and any future SCIP-aware feature (e.g. the deferred #215 Plan 3 clone moniker-collapse).

## Fix direction
Bring edge_oracle to survive-reindex parity with monikers: drop the CASCADE FK, re-anchor verdicts by content (`file_sha` + callee byte range — the `idx_edge_oracle_staleness(file_sha, tool, tool_version)` index already exists), and have reads join live edges so stale/changed-file verdicts don't count. Unchanged files (same file_sha) keep their resolution across reindex; changed files' verdicts are stale and dropped on the next oracle run.

Issue #231, opened 2026-06-19
perf(clones): investigate BLOB-packing the token bag (vs per-token symbol_token_postings rows)


skakri:
## Context
The #215 SourcererCC clone-detection substrate stores each function symbol's normalized token bag as **one `symbol_token_postings` row per distinct token** (`(symbol_id, normalizer_kind, token_hash, freq)`). On a full rebuild that's ~1M+ serial INSERTs on the cargo corpus and is the **dominant residual indexing cost** after the cheap perf fixes in #229 — skip the redundant `clone_token_df` upsert on full rebuild (`a0c6eb9`) and drop the unused `idx_symbol_token_postings_inv` (`dadaa2c`), which brought `index_time/full_rebuild_cargo` from +22.8% to +6.2% over the pre-#215 baseline on the dev box.

## Idea
Store each symbol's token bag as a **single serialized BLOB** on `symbol_fingerprints` (e.g. a `token_bag` column packing the `(token_hash, freq)` multiset) instead of N `symbol_token_postings` rows — collapsing the dominant write volume from ~1M rows to **one row per symbol** (~30k on cargo).

The Plan-1 candidate read (`query_api/clones.rs::load_scoped_baseline_bags` / `candidate_pairs`) already loads bags **by `symbol_id`** and builds the inverted index **in Rust**, so it loses nothing it currently uses — it would just deserialize the BLOB instead of grouping rows.

## Trade-offs to investigate
- **df refresh:** `rebuild.rs::refresh_clone_token_df` currently does a SQL `GROUP BY (normalizer_kind, token_hash)` over postings. With BLOBs it would deserialize + aggregate bags in Rust (or keep a lightweight derived df table). Measure which is faster.
- **Plan 2 `clones_for_symbol` reverse lookup** (token_hash → symbols): a postings table is the natural home for a `token_hash`-keyed lookup; a BLOB is not. Resolve the tension — e.g. keep the BLOB for the bulk bag read **plus** a lean token index built once post-rebuild, or accept a Rust-side scan for the reverse lookup at Plan-1/2 scale.
- **SCIP secondary bag (Plan 3)** would pack the same way (separate `normalizer_kind`).
- BLOB encoding: fixed-width LE `(i64, i64)` pairs, sorted by `token_hash`, with a version/length header; reuse `NORM_VERSION` for invalidation.

## When
[...]

PR #252, opened 2026-06-23
feat(clones): multi-language correctness — comments, literals, TS function-valued declarators, generated-skip (#232)

### LINKED ISSUE #232: clone detection: multi-language correctness hardening (comments, literals, language partition, normalizer_version, TS coverage)
Codex's re-review of the #215 SourcererCC clone-detection substrate (PR #229) surfaced a cluster of **multi-language correctness** gaps. None are live bugs today — the substrate has no query surface yet (Plan 2) and the Plan-1 tests/corpus are Rust-only — but they degrade recall/precision on non-Rust code and one is a latent correctness bug on a `NORM_VERSION` bump. They're coupled (two of them bump `NORM_VERSION`, which is what makes the version filter necessary) and all want the cross-language test scaffolding Plan 2 builds, so tracking them together for **Plan 2**.

(Generated-file exclusion from the candidate read was the one cheap, standalone item and is handled in #229 directly.)

## Normalization quality (`index/clones/normalize.rs`) — affects stored fingerprints (needs a `NORM_VERSION` bump + reindex)
1. **Comments aren't ignored.** Grammars that surface comments as leaf nodes get the comment text emitted as verbatim structural tokens, so two otherwise-identical functions that differ only by comments get different `struct_hash` and enough non-overlap to fall below θ → real clones missed. Skip `node.is_extra()` / comment kinds in `walk` before the verbatim-token branch.
2. **Multi-language literal leaves aren't bucketed.** In TS/Python/C/C++, `true`/`false`/`null`/`undefined`/`none` are leaf kinds with those exact names (not `*_literal`), and TS string contents are `string_fragment` leaves — all fall through verbatim, violating the literal-bucketing invariant off-Rust (functions differing only in those literal values diverge). Add these kinds to the literal bucket — carefully, without reintroducing the Rust double-emission the earlier `true`/`false` removal fixed (in Rust they sit under a `boolean_literal` parent caught by `ends_with("literal")`).

## Candidate-read filters (`index/query_api/clones.rs::load_scoped_baseline_bags`)
3. **Partition candidates by language.** All baseline fingerprints go into one pool and `SymbolBag` carries no language, so the sub-block pass can pair Rust/TS/C functions through shared `IDn`/punctuation/keyword tokens; the refinement table is single-language and the normalizer's structural node names are grammar-specific, so mixed-language components are noise. Load `symbols.language` and only pair within the same language.
4. **Filter by `normalizer_version`.** The read filters `normalizer_kind` but not `normalizer_version`. When `NORM_VERSION` bumps (e.g. fixing #1/#2 above), unchanged files keep old-version rows while reindexed files get the new token space → the read compares mixed-version `struct_hash`/postings and can miss clones or group incompatible fingerprints until a full rebuild. Bind `sf.normalizer_version = clones::NORM_VERSION` on the fingerprint + postings loads (or invalidate old rows on a version change). **This becomes load-bearing the moment #1/#2 land.**

## Coverage (`index/file_index.rs` / `index/clones`)
5. **Fingerprint function-valued declarators.** In TS, `const load = () => { … }` and class field handlers are recorded as `variable_declarator`/`public_field_definition` with kind `const`, so the `kind == "function"` fingerprint guard drops them — those function bodies can never be clone candidates. Either classify function-valued declarators as functions before the guard, or extend the fingerprintable-kind set.

## Also (perf, complementary to #229's generated read-filter)
6. **Skip fingerprinting generated files at index time.** #229 filters generated files out of the *read*; the *index* still computes + writes fingerprints/postings for path-generated files (`file_is_generated` true, kind ≠ `Generated`) that are then never read. Skipping fingerprint computation for generated files in the prepare phase avoids the dead write volume.

## Sequencing
Best done as one batch in Plan 2 alongside the `find_clones`/`clones_for_symbol` surface, with a small multi-language fixture (Rust + TS + Python) so the partition/literal/comment behavior is actually exercised. #1+#2 share one `NORM_VERSION` bump; ship #4 with them.

Refs: #215, #229.

skakri:
Fixes #232. Multi-language correctness hardening for the #215 clone substrate.

## What
Four correctness fixes so clone recall/precision hold beyond Rust, plus a Rust/TS/Python test harness and one `NORM_VERSION` bump:
1. **Ignore comments** — comment leaves were emitted as verbatim structural tokens, so comment-only differences broke `struct_hash`/overlap. `walk_spanned` now skips `node.is_extra() || kind.contains("comment")` (pushing neither token nor span — the seq↔span bijection is preserved).
2. **Bucket strings + value-erase booleans** — TS `string_fragment` bodies join the literal bucket (Python `string_content` was already covered); `true`/`false` collapse to a **single value-erased `LIT_BOOL`** (not per-value `LIT_TRUE`/`LIT_FALSE`), so two bodies differing only in a boolean are clones. The null-family is deferred (single fixed lexemes, ~0 recall value, and a cross-language wrapped-node hazard).
3. **Fingerprint TS function-valued declarators** — `const load = () => {…}` and class-field arrow handlers (parsed as `variable_declarator`/`public_field_definition`, kind `const`) were dropped by the `kind == "function"` guard. The clone fingerprint guard is now function-value-aware (`child_by_field_name("value") ∈ {arrow_function, function_expression}`) — **symbol `kind` is unchanged** (no ripple into chunking/graph/search).
4. **Skip fingerprinting generated files at index time** — path-heuristic generated files (`src/generated/*`, `*.d.ts`) were fingerprinted then never read. Pure write-side storage hygiene (the candidate read already filters `files.generated = 0`).

Items #3 (language partition) and #4 (`normalizer_version` filter) from the issue were **already shipped** by the #231 BLOB-pack — confirmed, not re-done.

## How it stays correct
- **Grammar-verified:** every tree-sitter kind name was checked against a live AST probe of the pinned grammars (tree-sitter 0.26.9 / -rust 0.24.2 / -typescript 0.23.2 / -python 0.25.0 / -c 0.24.2 / -cpp 0.23.4).
- **Type-recovery contract updated (a net improvement):** `literal_bucket_to_type` (signature.rs, the exhaustive-buckets source of truth) + the antiunify `widen_string_content_run`/type-hint paths now map `LIT_BOOL → bool` and `string_fragment → &str`. Rust bool typing was previously *dead* (booleans went verbatim; the old `LIT_BOOLEAN_LITERAL→bool` arm was never emitted), so `LIT_BOOL` both erases the value *and* recovers `bool`.
- **`NORM_VERSION` 1→2** auto-invalidates old fingerprints (the read binds `sf.normalizer_version = NORM_VERSION`) and the 4b refinement cache (folded into the content-addressed key). Existing indexes recompute clone fingerprints on the next index — no schema migration.

## Verification
997 tests green across `rag-rat-core` + `rag-rat` + `rag-rat-mcp`; clippy `-D warnings` + nightly fmt clean. The #231 recall-parity pin + the #248 reindex-survival trip-wire still pass. A multi-language integration test plants within-language clones (comment-only, string-literal, TS function-valued) in a Rust+TS+Python fixture and asserts clones are found within each language with **no cross-language components**.

The change went through plan → 3-explorer **grammar-verified** adversarial review (4 blockers folded: the boolean value-erase semantics, the `literal_bucket_to_type`/antiunify contract audit, the pre-parse early-bail widen for #5, and the MIN_TOKENS fixture sizing) → implementation.

## Deferred / follow-up
- Null-family literal bucketing (`null`/`undefined`/`none`/`nullptr`/C `NULL`) — low recall value + cross-language leaf-kind inconsistency; left verbatim.

### DIFF
--- a/crates/rag-rat-core/src/index/clones/mod.rs
+++ b/crates/rag-rat-core/src/index/clones/mod.rs
@@ -15,7 +15,15 @@ use tree_sitter::Node;
 
 /// Bumped when normalization changes; invalidates fingerprints (and the later refine cache) without
 /// a schema migration.
-pub(crate) const NORM_VERSION: i64 = 1;
+///
+/// `2` (#232): the normalization stream changed in two ways that affect every fingerprint — (1)
+/// comments and other tree-sitter EXTRAs are skipped (`walk_spanned`), and (2) multi-language
+/// literal leaves bucket (TS `string_fragment` → `LIT_STRING_FRAGMENT`; `true`/`false` value-erase
+/// to `LIT_BOOL`). The already-shipped read filter (`sf.normalizer_version = NORM_VERSION` on every
+/// fingerprint read) auto-EXCLUDES the v1 rows, which are recomputed at v2 on the next reindex; the
+/// 4b refinement cache invalidates via `NORM_VERSION` in the content-addressed `refinement_key` +
+/// the freshness predicate (`refine::cache`). No schema migration is needed.
+pub(crate) const NORM_VERSION: i64 = 2;
 /// Bumped when the LCS alignment / refinement algorithm changes; participates in the content-
 /// addressed `refinement_key` and in the `clone_refinements` cache freshness predicate, so a bump
 /// invalidates every cached refinement without a schema migration (the same discipline as
@@ -137,6 +145,15 @@ mod tests {
     /// Rust-only wrapper — the existing Rust fingerprint tests call this unchanged.
     fn fp(src: &str) -> Option<SymbolFingerprint> {
         fp_lang(src, "t.rs", Language::Rust)
+    }
+
+    #[test]
+    fn norm_version_is_2() {
+        // #232: the comment-skip (T2) + multi-language literal bucketing (T3) changed the
+        // normalization stream, so NORM_VERSION must be 2. The read filter + content-addressed
+        // refinement key both key off this constant, so a stream change without the bump silently
+        // serves stale fingerprints/refinements.
+        assert_eq!(NORM_VERSION, 2);
     }
 
     #[test]
--- a/crates/rag-rat-core/src/index/clones/mod.rs
+++ b/crates/rag-rat-core/src/index/clones/mod.rs
@@ -112,12 +112,41 @@ mod tests {
     use crate::index::parser;
     use crate::language::Language;
 
-    fn fp(src: &str) -> Option<SymbolFingerprint> {
-        let parsed = parser::parse_file(Path::new("t.rs"), Language::Rust, src).expect("parse");
-        let func = parsed.symbols.iter().find(|s| s.kind == "function").expect("function");
-        let node =
-            parsed.root().descendant_for_byte_range(func.start_byte, func.end_byte).expect("node");
+    /// Generalized fingerprint harness (#232): parse `src` with `language` (under `path`), select
+    /// the target symbol — the one whose subtree normalizes to the MOST tokens, language-agnostic
+    /// so it works for Rust `function`, TS `const`/function-valued declarators, and Python
+    /// `function` alike — and return its fingerprint (`None` below `MIN_TOKENS`). Test fixtures
+    /// MUST clear `MIN_TOKENS` (20) or `fp_lang` returns `None` for the size gate, not the
+    /// property under test.
+    fn fp_lang(src: &str, path: &str, language: Language) -> Option<SymbolFingerprint> {
+        let parsed = parser::parse_file(Path::new(path), language, src).expect("parse");
+        let node = parsed
+            .symbols
+            .iter()
+            .filter_map(|s| {
+                let node = parsed.root().descendant_for_byte_range(s.start_byte, s.end_byte)?;
+                let len = normalize::normalize_baseline(node, src).len();
+                Some((len, node))
+            })
+            .max_by_key(|(len, _)| *len)
+            .map(|(_, node)| node)
+            .expect("at least one symbol with a normalizable subtree");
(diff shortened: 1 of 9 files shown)


skakri:
## 3-adversary audit: SOUND, merge-safe

Three independent adversaries (normalization-across-grammars / type-recovery-contract / #5+#6+integration), each building live AST probes against the pinned grammars. All **SOUND** — zero correctness blockers.

- **Normalization:** comment-skip + bijection correct on all 6 grammars; LIT_BOOL value-erase correct in 5 langs (no double-emit); TS template literals + Python f-strings *are* bucketed.
- **Type contract:** no boolean regression (zero golden assertions changed — strictly additive); the value-erase does **not** hide the variation (`per_member_values` carry raw `true`/`false`); enumeration-doc exhaustive.
- **#5/#6:** classifier empirically correct on TS+TSX (accepts arrow/func-expr/async/class-field/typed/generic; rejects number/object/destructure/alias/ternary); #6 index-skip == read-filter parity on every write path incl. heal.

Polish folded (`0f23f91`): corrected two factual-error doc comments (Kotlin booleans; post-#6 df scope). Contained recall/cosmetic gaps filed as follow-ups: #253 (Kotlin booleans + C/C++ char), #254 (template-literal widening).

## ⚠️ Operational: NORM_VERSION 1→2 requires a full reindex
Existing indexes keep v1 fingerprints, which the candidate read filters out — so **clone tooling returns zero candidates on any existing index until `rag-rat index --full`** (the migration-free invalidation working as designed; incremental/hook reindex only re-fingerprints *changed* files). After merge + installing the v2 binary, run `rag-rat index --full` on the dogfood index (the same make-it-live step as edge_oracle/#250). No code action; just don't expect clone results on a stale index.

Issue #122, opened 2026-06-14
Research/spike: dream mode — scheduled local-AI memory-maintenance daemon


skakri:
**Research / spike.** A scheduled local-AI ("dream mode") daemon that audits/refreshes the memory layer — freshness audits, issue→memory gap mining, contradiction detection, coverage maps, rejected-alternative mining — emitting **findings, not facts** into a separate `dream_findings` table, promoted only by deterministic verifier / human / strong-agent confirmation.

**Five refinements that decide differentiation vs week-two-disable:**
1. **"Idle" is the WRONG trigger.** rag-rat fought hard to make idle cheap (the self-sustaining watcher loop bug; the no-op-pass / `GC_EVERY_PASSES` / 2-worker-tokio idle-load work). LLM inference on "stopped typing" = fan/battery/heat → disabled. Make it an explicit `rag-rat dream` batch command for CI / cron / server / manual; if local, gate on AC + thermal-headroom + long-idle. (First generative-LLM dependency — keep Ollama out-of-process, never in the binary.)
2. **Deterministic layer GATES the model.** Anchor validity is already self-healed for free; wake the model only on the semantic residue (stale/relocated anchor, touched bound symbols, a fired heuristic). The model never auto-closes — the deterministic checker does.
3. **Contradiction detection needs TWO signals** (model's semantic flag + an effect/contract confirmation), or it floods the scarcest resource — attention.
4. **Identity-key findings** `(kind, subject, claim-hash)` so re-runs supersede not stack; unreviewed low-confidence findings auto-decay. Otherwise the pile rots and erodes trust in fresh memories.
5. **Measure it on commit-replay (#120)** — recall (does it surface rationale a later commit had to rediscover) + finding-acceptance-rate (precision) — or it's faith-based.

**Build first inside dream mode** (not the freshness audit, which is mostly deterministic): (a) **rejected-alternative mining** — highest-value/lowest-supply memory kind; v1 from **PR review threads** (GitHub API already pulled, no new plumbing), v2 from agent sessions (new capture pipeline). (b) **memory-coverage map** — high-risk × low-memory-density (risk score × memory count); no model needed.

**Framing:** a cheap overnight triage producing a ranked, evidence-anchored worklist for the strong agent — NOT an autonomous closer.

Depends on #120, #121. Ref: `docs/plans/2026-06-14-agent-value-strategy.md` §4e.

skakri:
[...]

PR #260, opened 2026-06-23
feat(dream): deterministic memory-maintenance worklist v1 (#122)

skakri:
First shippable slice of dream mode (#122): a **zero-LLM, deterministic** `rag-rat dream` command that surfaces a reviewable worklist of findings **about** memories — and never mutates one.

## What it adds
- **Migration V033 `dream_findings`** — the worklist table. Findings are identity-keyed `(kind, subject, claim_hash)` with a refresh / supersede / resolve lifecycle. Polymorphic `subject` (memory id or symbol/path), so no FK.
- **`rag-rat dream`** command (TOON/JSON, via `output::render`) with two finding kinds:
  - `coverage_gap` — load-bearing symbols (call-graph in-degree, distinct callers) with no memory binding; test-infra filtered.
  - `stale_reference` — memory bodies referencing a `.rs` path that no longer resolves against the index (suffix-aware, so prose shorthand still resolves).

## Safety
`dream_run` writes **only** to `dream_findings` — it never mutates a `repo_memories` row (dream mode proposes; a human/strong-agent confirms). The command takes the `WriteLock` like every other write command. A column-snapshot test guards the no-mutation invariant.

## Why deterministic / what's deferred
The spike (write-up in #122) found that **change-based** staleness signals (churn, struct-drift) are *unmeasurable* on this young repo — base rates saturate. So v1 ships the **rate-independent** signals (`coverage_gap` + reference-resolution) and explicitly defers the semantic-contradiction layer and the staleness-efficacy validation (which needs an accumulated review ledger over time). Details + the four commit-replay evals are in the #122 write-up.

## Quality
Hardened by three internal adversarial reviews, which caught and fixed: a status-blind lifecycle bug (flip-back / resolved-reappears), `coverage_gap` scan-budget starvation + caller over-count + an N+1, the missing `WriteLock`, a `stale_reference` false-positive source, and non-deterministic tie ordering. Full `rag-rat-core` suite green (865/865), clippy clean, fmt clean. Rebased on current `main`.

Refs #122. Does not close it — the deferred semantic layer + staleness-efficacy validation remain.

### DIFF
--- a/crates/rag-rat-cli/src/cli.rs
+++ b/crates/rag-rat-cli/src/cli.rs
@@ -69,6 +69,10 @@ pub(crate) enum Command {
     /// Inspect and re-anchor source-anchored repo memories.
     Memory(MemoryArgs),
 
+    /// Dream-mode memory-maintenance worklist (#122): deterministic coverage-gap + stale-reference
+    /// findings written to `dream_findings`. Surfaces findings ABOUT memories; never mutates them.
+    Dream(DreamArgs),
+
     /// GitHub papertrail sync.
     Github(GithubArgs),
 
@@ -175,6 +179,13 @@ pub(crate) struct ImportantSymbolsArgs {
     /// global-by-default — it never auto-seeds from the git diff).
     #[arg(long, value_delimiter = ',')]
     pub personalize: Vec<String>,
+}
+
+#[derive(Debug, Args)]
+pub(crate) struct DreamArgs {
+    /// Max coverage-gap findings to surface (stale-reference findings are always all reported).
+    #[arg(long)]
+    pub limit: Option<u32>,
 }
 
 #[derive(Debug, Args)]
--- a/crates/rag-rat-cli/src/commands.rs
+++ b/crates/rag-rat-cli/src/commands.rs
@@ -132,6 +132,9 @@ pub(crate) fn important_symbols(
 /// stale references), sync it into `dream_findings`, and render the open worklist. Writes ONLY to
 /// `dream_findings` — never mutates a memory.
 pub(crate) fn dream(config: &Config, args: &DreamArgs) -> anyhow::Result<()> {
+    // `dream` WRITES dream_findings — serialize with the watcher/index like every other write
+    // command (index/maintenance/oracle); WriteLock is reentrant so the open-time migrate is safe.
+    let _lock = rag_rat_core::locks::WriteLock::acquire_blocking(&config.database)?;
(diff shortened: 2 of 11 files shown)


skakri:
Thanks — fair review. Addressed in `e3f4101` (866/866 core suite green, clippy + fmt clean).

**Must-fix (both fixed):**
1. **`coverage_gap` logical_symbol coverage** — now mirrors the canonical coverage in `query/memory/api.rs`: excludes callees covered via a `logical_symbol` binding too, mapped through `logical_symbol_members` (a logical binding covers *all* member variants, not just the one stored `symbol_id`). (Minor note for the record: on the current index all 13 logical_symbol bindings happen to also carry a non-null `symbol_id`, so the `symbol_id=NULL` path isn't triggered today — but mirroring the canonical definition is the right robustness fix regardless.) Also swapped the unanchored `path LIKE '%test%'` for `has_test_code = 0`.
2. **`stale_reference` status** — now `status IN ('active','stale')`, matching the memory layer.

**Should-fix:**
- **Age-decay** — implemented (`effective_rank`: base_rank halved every 2 weeks since `first_seen_at_ms`), now drives the worklist ordering, with a unit test. The schema comment no longer overclaims.
- **STRICT** — `dream_findings` is now `STRICT` (greenfield; aligns with the recent `edge_oracle`/`symbol_token_postings` tables).

**Nit:**
- **Atomicity** — `sync` runs in one transaction; no torn worklist on a mid-run failure.

**Deferred (follow-ups, noted in code comments):**
- **Caller in-degree checkout-scoping** — the `files` join already resolves to the `temp.files` view so *callees* are checkout-scoped, but the caller count isn't. The right fix is to reuse the already-scoped `important_symbols` PageRank instead of this raw in-degree proxy; filed as follow-up.
- **`accepted`/`dismissed`/`reviewed_at_ms` wiring** — those columns are the lifecycle target of the review command, which is the next slice.
- **MCP `dream` tool** — agreed the pull-based strong-agent consumer wants it; CLI-only is the v1 surface, MCP tool to follow.

**On the #122 scope-inversion:** that's a genuine product-priority call. The spike's finding (in the #122 write-up) was that the deterministic freshness/coverage signals are the *validatable, safe, zero-LLM* slice, whereas rejected-alternative mining is generative + LLM-heavy (the v4 in the design). I shipped the safe deterministic core first deliberately — but if you'd rather mining lead, I'm happy to resequence; it's a sequencing decision, not a defect.

Issue #292, opened 2026-06-24
Write-time clone check (#287): near tier too noisy on boilerplate / test code


skakri:
## Observed
While editing rag-rat itself with the #287 write-time clone-check hook live, writing a single test function (`no_op_on_unparseable_text`) triggered a warning listing it as **~88% similar to 50+ other test functions** across the repo. A second new test flagged ~25 matches. None are real clones — test functions all share the "build fixture → call → assert" shape, and boilerplate-heavy Rust (`let`/`if`/`for`/`match`/`.unwrap()`) pushes token-bag overlap above θ=0.7 between unrelated functions.

The **exact** tier (struct_hash) stays precise and high-value. The **near** tier (≥0.7 overlap) is the noise source, and it's worst exactly where it's least useful: test modules.

## Why it matters
At this noise level an agent learns to ignore the hook, which kills the feature. The signal (real exact/near duplicates) drowns in boilerplate matches.

## Options (pick a combination)
1. **Exclude `#[test]` / `#[cfg(test)]` functions** from the write-time check — `find_clones` already has a `tests_excluded` notion; reuse it. Highest-leverage, removes the worst noise.
2. **Cap the match list** in the hook output (e.g. top 5 by similarity) so a single function can't print 50 refs.
3. **Raise the hook's near threshold** (e.g. 0.85) while leaving `find_clones`/`clones_for_symbol` at 0.7 — the write-time surface wants higher precision than the explicit query.
4. **Exact-only mode** for the hook, with near as opt-in.

Likely answer: (1) + (2) + a higher default near threshold. Threshold/exclusions should be configurable.

## Acceptance
Editing a typical test or boilerplate function produces zero or a small, genuinely-similar set — not a 50-item wall. Exact duplicates still always surface.

PR #295, opened 2026-06-24
chore(clones): consolidate the three test-path detectors into one canonical helper (#294)

### LINKED ISSUE #294: Consolidate the three overlapping test-path detectors into one canonical helper
Three functions independently answer "is this a test path?", with slightly different conventions — surfaced while building the #292 clone-check (whose own `is_test_file_path` is the third). This is *semantic* duplication (same purpose, different impl), so the structural clone check doesn't flag it; `semantic_search` did.

| function | module (layer) | input | conventions |
|---|---|---|---|
| `is_test_file_path` | `index/parser.rs` (index) | `&Path` | test dirs + `conftest.py` + `test_*`/`_test`/`_tests`/`*Test`/`*Tests`/`*TestCase` + `.test.`/`.spec.` (cross-language) |
| `is_test_like_path` | `query/text_compare.rs` (query) | `&str` | `/test/`, `/tests/`, `/__tests__/`, `_test.rs`, `.test/.spec.ts(x)` |
| `is_test_or_mock_path` | `query/repo_brief.rs` (query) | `&str` | the above + `/__mocks__/`, `/tests.rs`, `_tests.rs` |

## Proposal
One canonical `is_test_path` helper, the UNION of the conventions above (don't lose `/__mocks__/` from repo_brief, or the cross-language filename forms from parser).

## Care points
- **Layering:** `query/` depends on `index/`, not vice versa, so the canonical helper must live in a layer both can reach (the `index`/parser side, or a lower shared util) — `parser.rs`'s version is already the most complete.
- **`&Path` vs `&str`:** take `impl AsRef<Path>` (or a `&str` form + convert at the parser call site).
- **Behavior re-validation:** the existing callers change behavior when broadened — `index/staleness.rs` (test-file skip), `index/query_api/graph.rs:274` (test-callsite edge filter), `query/repo_brief.rs` (support-path down-weight). Broadening is probably correct (they ARE test paths) but each needs a check + the union must keep `/__mocks__/`.
- Keep `is_test_scaffolding_text` separate — it's CONTENT detection (`.mock`, `jest.`, `expect(`), a different concern that the clone graph happened to group with `is_test_like_path` at 86% structural similarity.

Note: `symbols.is_test` (the #292 index-time flag) is the durable per-symbol signal now; the query-layer callers could eventually read it instead of re-deriving from the path, but that's a larger change — this issue is just the path-helper dedup.

skakri:
Closes #294. Three functions independently answered "is this a test path?" — `is_test_like_path` (query/text_compare), `is_test_or_mock_path` (query/repo_brief), and the #292 `is_test_file_path` (index/parser) — each with slightly different conventions. Semantic duplication the structural clone check doesn't flag; `semantic_search` surfaced it while reviewing #293.

## Change
One canonical **`parser::is_test_path(impl AsRef<Path>)`** — the **union** of all three:
- dir segments `tests`/`test`/`__tests__`/`__mocks__`/`spec` (case-insensitive — `__mocks__` was repo_brief-only)
- `conftest.py`, `*.test.*`, `*.spec.*`
- stems `test`/`tests`/`test_*`/`*_test`/`*_tests`/`*Test`/`*Tests`/`*TestCase` (cross-language — the `*Test`/`*TestCase` forms were parser-only)

`impl AsRef<Path>` lets the `&str` query callers and the `&Path` parser caller share it. The two query copies are **deleted**; all four call sites (`text_compare`, `repo_brief`, `staleness`, `graph`) repoint to the canonical — which also **drops `staleness`'s index→query dependency** (bonus layering cleanup).

`is_test_scaffolding_text` is intentionally left alone — it's CONTENT detection (`.mock`/`jest.`/`expect(`), a different concern that the clone graph had grouped with `is_test_like_path` at 86% structural similarity.

## Behavior
The query callers (`staleness` test-skip, `graph` test-callsite filter, `repo_brief` support-path down-weight) now use the broader union, so they catch more genuine test paths (e.g. root-level `tests/`, `*Test.kt`, `conftest.py`) — and `repo_brief` keeps `__mocks__`. No path that matched before stops matching.

## Tests
A direct union-coverage test (every convention + near-miss negatives `latest.rs`/`contest.rs`/`manifest.rs`) plus the existing cross-language integration tests. 1045 workspace tests, clippy `--all-targets`, nightly fmt.

### DIFF
--- a/crates/rag-rat-core/src/index/parser.rs
+++ b/crates/rag-rat-core/src/index/parser.rs
@@ -651,7 +651,7 @@ fn detect_is_test(
     scope_path: &str,
     name: &str,
 ) -> bool {
-    if is_test_file_path(path) {
+    if is_test_path(path) {
         return true;
     }
     match language {
@@ -675,11 +675,20 @@ fn detect_is_test(
     }
 }
 
-/// Cross-language test-FILE conventions: a `tests`/`spec` directory, or a filename like
-/// `test_*` / `*_test` / `*_tests` / `*Test` / `*Tests` / `*.test.*` / `*.spec.*` / `conftest.py`.
-fn is_test_file_path(path: &Path) -> bool {
+/// The CANONICAL cross-language test-path detector (#294) — the one reused by the indexer (the
+/// `is_test` computation) AND the query layer (`staleness` test-file skip, `graph` test-callsite
+/// filter, `repo_brief` support-path down-weight). A test path is any of: a test directory segment
+/// (`tests`/`test`/`__tests__`/`__mocks__`/`spec`, case-insensitive), `conftest.py`, a `*.test.*` /
+/// `*.spec.*` filename, or a stem like `test`/`tests` / `test_*` / `*_test` / `*_tests` / `*Test` /
+/// `*Tests` / `*TestCase`. Takes `impl AsRef<Path>` so both `&Path` (parser) and `&str` (the query
+/// callers' stored path strings) pass directly.
+pub(crate) fn is_test_path(path: impl AsRef<Path>) -> bool {
+    let path = path.as_ref();
     if path.components().filter_map(|component| component.as_os_str().to_str()).any(|segment| {
-        matches!(segment, "tests" | "test" | "__tests__" | "__test__" | "spec" | "specs")
+        matches!(
+            segment.to_ascii_lowercase().as_str(),
+            "tests" | "test" | "__tests__" | "__test__" | "__mocks__" | "spec" | "specs"
+        )
     }) {
         return true;
     }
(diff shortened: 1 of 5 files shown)


skakri:
Folded in a second small dedup from the post-#294 clone sweep: `enrich_symbol_kinds` / `enrich_memory_counts` in `repo_brief.rs` were 99.5% identical (same per-path enrich loop, differing only in the lookup fn + target field) → extracted a generic `enrich_rows<T>(counts_by_path, assign)`, kept the two as named `pub(crate)` wrappers.

The rest of the ≥0.9 sweep was either false-positive structural similarity (`parse_error`/`parse_symbols` return different types; `first_/last_descendant_node` are different traversals) or idiomatic API mirrors (`find_callers`/`trace_callees` differ only in a direction bool — left as-is, the symmetry reads better than a shared flag helper). No other real duplicates.

Issue #120, opened 2026-06-14
Spike: commit-replay retrieval eval harness (measure-first foundation)


skakri:
**Spike — measure-first foundation.** Build a commit-replay retrieval eval harness. rag-rat has NO task-shaped retrieval eval today — only the SCIP oracle (edge-resolution precision) and a hand-authored static fixture (`eval.rs`). Replay gives free, auto-growing ground truth: the diff IS the gold.

**Mechanism:** for a past PR/commit, check out the parent, index the pre-state, use the issue/commit text as the query, and score whether retrieval / `edit_brief` surfaces the files/symbols/tests the merge actually touched.

**Caveats to design in from the start:**
- **Label noise** — a PR diff bundles refactors/formatting; weight by changed-*symbol*, exclude bulk/mechanical commits, treat file-recall as a loose upper bound.
- **Leakage / time-split** — PR/issue bodies are often written post-fix and name the solution → prefer the issue body *at open time* as the query; if you ever TRAIN on replay, split by time (old→train, recent→eval) or you measure memorization.
- **Volume** — this repo is ~222 commits: enough for an eval harness + few-shot calibration + hard-negative mining, NOT to fine-tune a retriever from scratch.

**Why first:** you can't improve (or safely turn on fact generation / dream mode) what you can't measure. Everything downstream rides on this.

Ref: `docs/plans/2026-06-14-agent-value-strategy.md` §4d.

Issue #112, opened 2026-06-14
Research: stronger code embedder + vector quantization (ANN deferred)


skakri:
**Research.** Evaluate offering an optional stronger code-trained embedder tier (CodeRankEmbed 137M / jina-code-0.5B) alongside the current model2vec (order-insensitive) and fastembed backends, plus vector quantization (int8 / binary / Matryoshka truncation) to shrink the on-disk index.

**Why it helps an agent:** better semantic recall on code queries; a smaller index keeps the brute-force flat scan fast.

**Note — ANN explicitly deferred:** the brute-force `vector_candidates` scan is sub-50ms and *exact* at repo scale; HNSW/IVF/DiskANN only pay off at ~1M+ vectors. Do NOT add ANN until a measured latency problem appears on a large corpus; then prefer `sqlite-vec` (exact, in-SQLite) or `usearch` (mmap on-disk) over an in-memory faiss server. (Captured as a repo memory.)

Ref: `docs/plans/2026-06-14-agent-value-strategy.md` §4.6–4.7.

skakri:
## Resolved — the embedder + quantization research is complete (measured).

### Stronger code embedder: no off-the-shelf win (measured on the #120 commit-replay eval)
A/B'd against all-MiniLM-L6-v2 (200 cases, at-head):
- **bge-small-en-v1.5** (general, 384d): ties MiniLM — recall@10 0.612 vs 0.611 (within noise).
- **jina-v2-base-code** (code-specific, 768d): **loses** — recall@10 0.592 vs 0.611, at 2× the vector size.
- A general **cross-encoder** (bge-reranker-base) over the top-50: **hurts** — −0.075 recall@3, +24s/query.

Swapping the embedder does not move retrieval on this corpus. The limiter is **candidate-generation recall** (~35% of gold never retrieved — commit messages are adversarial queries that describe the *change*, not the file's content) plus the eval's query distribution — not the embedder. The CodeRankEmbed self-export path is documented (rag-rat memory) but gated on a measured win that nothing has cleared.

### Shipped
- **Centralized embedding-model registry** (#309) — adding/comparing a model is one table row. BGE + jina are selectable via `[local_ai.embedding] model = "..."`; **MiniLM stays the default** (measured).
- **int8 scalar quantization** for `chunk_embeddings` (#311) — ~4× smaller vectors, neutral quality.
[...]

PR #308, opened 2026-06-24
feat(embed): add BGE-small-en-v1.5 as a selectable embedder (#112)

skakri:
## What

Adds **BGE-small-en-v1.5** as a selectable embedder alongside the default all-MiniLM-L6-v2. BGE-small is a stronger general-retrieval model at the **same 384-dim**, so switching is a re-embed — no schema/dim change.

- Parameterizes `FastEmbedEmbedder` (`new` / `new_bge_small` / `with_model`).
- Registers the model in the manifest + the manifest-current check.
- Wires the factory / model-version / expected-dim / install paths; `install_fastembed_model` now inits the model matching the requested id (BGE pulls its own weights).
- Builds with and without the `fastembed` feature.

Activate: `rag-rat models install fastembed-bge-small-en-v1.5` + reconcile.

## Measurement

recall@10 / symbol-recall vs all-MiniLM on the commit-replay eval (#120) — running in an isolated bench checkout; delta to follow here. (Hash baseline: recall@10 0.420 / symbol 0.276 — the question is how much a real embedder moves it.)

## Follow-ups
- `fastembed_cache_ready` status display is still MiniLM-shaped (cosmetic; the embedder downloads via fastembed regardless).
- CodeRankEmbed (code-specific, 768-dim) is the bigger-win tier (T16) once this drop-in is validated.

### DIFF
--- a/crates/rag-rat-core/src/index/ai/helpers.rs
+++ b/crates/rag-rat-core/src/index/ai/helpers.rs
@@ -19,11 +19,7 @@ pub(crate) fn embed_query_with(
     embedder: &dyn Embedder,
     query: &str,
 ) -> anyhow::Result<QueryEmbedding> {
-    // Apply the model's query-side instruction prefix (BGE asymmetric retrieval); empty for
-    // symmetric models, so this is a no-op for hash / all-MiniLM / model2vec (#112 review).
-    let prefix = embedder.query_prefix();
-    let text = if prefix.is_empty() { query.to_string() } else { format!("{prefix}{query}") };
-    let texts = vec![text];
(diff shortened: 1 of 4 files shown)


skakri:
Superseded by #309. The registry refactor in #309 rewrites BGE's wiring (BGE is now one row in the `EMBEDDING_MODELS` table), so BGE ships there as a selectable option alongside jina-v2-base-code rather than as its own change. #309 also carries the measured A/B (BGE ties MiniLM; jina doesn't beat it → MiniLM stays default) and all of this branch's commits. Closing in favor of #309.

PR #311, opened 2026-06-25
feat(embed): int8 scalar quantization for chunk_embeddings — ~4x smaller, neutral quality (#112)

skakri:
int8 scalar quantization for `chunk_embeddings` (#112) — ~4x smaller embeddings, zero quality cost.

## Measured (200-case commit-replay, MiniLM, at-head, same cases)
| metric | f32 | int8 | Δ |
|---|---|---|---|
| **bytes/chunk** | 1536 | **388** | **3.96× smaller** |
| recall@3 | 0.315 | 0.318 | +0.003 (neutral) |
| recall@10 | 0.424 | 0.424 | flat |
| mrr@10 | 0.651 | 0.656 | +0.005 (neutral) |

The vector blobs shrink ~4× on disk and in the brute-force vector scan's working set; quality is neutral (the tiny + deltas are noise — nothing regressed). On a large index (e.g. a kernel-scale 4M-chunk repo at 384-dim) that's ~6.5 GB → ~1.6 GB of vector storage.

## How
Symmetric int8 scalar quantization: one f32 `scale` (= max|v| / 127) + one signed byte per dim. `decode_vector` distinguishes the two on-disk formats purely by **length** (`4*dim` f32 vs `dim+4` int8 — they never collide for any dim ≥ 1), so legacy f32 blobs stay readable and an index can be mixed-format during migration. New writes are int8; the vector search path is unchanged (it already decodes via `decode_vector`). Unit-tested: round-trip within one quant step, cosine preservation < 1%, legacy-f32 read, all-zero vector.

## Migration
Existing f32 rows stay readable and re-encode to int8 when their chunk is next re-embedded. A one-time re-encode pass (to realize the disk win on *stable* code without a full re-index) is filed as a follow-up.

Refs #112.

### DIFF
--- a/crates/rag-rat-core/src/index/ai/helpers.rs
+++ b/crates/rag-rat-core/src/index/ai/helpers.rs
@@ -192,21 +192,55 @@ pub(crate) fn fastembed_cache_dir() -> PathBuf {
     PathBuf::from(".rag-rat").join("models")
 }
 
+/// Symmetric int8 scalar-quantization range: codes span [-127, 127] (not 128) so a value and its
+/// negation map to opposite codes. The largest-magnitude component maps to ±127; the rest round
+/// proportionally.
+const INT8_QUANT_LEVELS: f32 = 127.0;
+
+/// Decode a stored embedding blob back to f32. TWO on-disk formats coexist, distinguished purely by
+/// length (they never collide: `4*dim != dim+4` for any `dim >= 1`):
+/// - **legacy f32**: `dim` little-endian f32 (`4*dim` bytes).
+/// - **int8 (#112)**: a leading f32 `scale` then `dim` signed-int8 codes (`4 + dim` bytes),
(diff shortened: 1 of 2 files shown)


skakri:
Thanks — that P1 (the stale `length(vector_blob) == HASH_EMBEDDING_DIM * 4` assertion in `reconcile_requires_explicit_model_install_and_ignores_stale_artifacts`) was already fixed in d4210f2, pushed after the reviewed commit 9eaf0ff. It now asserts `4 + HASH_EMBEDDING_DIM` (the int8 layout: a 4-byte f32 scale + one byte per dim). The CI `test`/`coverage` jobs flagged the same failure; current HEAD is green across build/clippy/rustfmt/test/coverage/eval. It was the only f32-blob-length assumption in the tree (grep-confirmed; the search path is format-agnostic via `decode_vector`).

PR #313, opened 2026-06-25
feat(embed): re-encode existing f32 embedding blobs to int8 (#312)

### LINKED ISSUE #312: int8 quant: re-encode existing f32 embedding blobs to int8 for stable code
Follow-up to the int8 quantization PR. New/re-embedded vectors are int8, but existing f32 `chunk_embeddings.vector_blob` rows for UNCHANGED files stay f32 (they're Current, never re-embedded), so an existing index keeps ~4x-larger vectors for stable code until wiped + re-indexed. `decode_vector` reads both, so it's correct — just misses the disk win. Add a one-time format-only re-encode (decode f32 → encode int8, no model inference) — e.g. a maintenance/reconcile pass or a `rag-rat reconcile --reencode-vectors` — so existing indexes realize the ~4x storage reduction without a full re-index.

skakri:
Follow-up to int8 quantization (#311): re-encode **existing** f32 vector blobs to int8.

## Why
New writes are int8 (#311), but existing f32 `chunk_embeddings.vector_blob` rows for unchanged files stay f32 forever (Current, never re-embedded), so an existing index keeps ~4× larger vectors for stable code. `decode_vector` reads both formats so it's correct — it just misses the disk win.

## What
A **format-only** re-encode (decode f32 → encode int8, **no model inference**, quality-identical to a fresh int8 reconcile):
- `reencode_legacy_f32_blobs` detects f32 rows by length (`length(vector_blob) = 4*embedding_dim`; int8 is `dim+4`, disjoint for dim≥1), converts in bounded per-batch transactions (4000/txn), idempotent + resumable + provably terminating.
- **Automatic, run-once**: a meta-gated wrapper (`vector_int8_reencode_done`) wired into `run_maintenance_pass` on the base index — the disk win applies after upgrade with no user action, and no full-table scan on later passes (the gate short-circuits; safe because new writes are always int8).
- **Explicit**: `rag-rat reconcile --reencode-vectors` forces the conversion one-shot (for a huge index now).

Tested: f32→int8 within one quant step, idempotent, the meta gate. Full suite green (924 default / 927 `--features eval`), clippy + fmt clean both configs.

Closes #312.

### DIFF
--- a/crates/rag-rat-cli/src/cli.rs
+++ b/crates/rag-rat-cli/src/cli.rs
@@ -313,6 +313,11 @@ pub(crate) struct ReconcileArgs {
     /// Truncate chunk text to this many chars before embedding.
     #[arg(long)]
     pub max_embedding_chars: Option<usize>,
+    /// Force the legacy-f32 → int8 vector re-encode now (#312), ignoring the run-once meta gate.
+    /// A format-only conversion (no model inference); idempotent — converts only rows still in
+    /// f32.
+    #[arg(long)]
+    pub reencode_vectors: bool,
 }
 
 #[cfg(feature = "eval")]
--- a/crates/rag-rat-cli/src/cli.rs
+++ b/crates/rag-rat-cli/src/cli.rs
@@ -315,7 +315,9 @@ pub(crate) struct ReconcileArgs {
     pub max_embedding_chars: Option<usize>,
     /// Force the legacy-f32 → int8 vector re-encode now (#312), ignoring the run-once meta gate.
     /// A format-only conversion (no model inference); idempotent — converts only rows still in
-    /// f32.
+    /// f32. SHORT-CIRCUITS: re-encodes and exits, ignoring the other reconcile flags (no
+    /// embeddings are computed). Honors `--max-seconds` (the conversion is bounded and resumes
+    /// on a later run).
     #[arg(long)]
     pub reencode_vectors: bool,
 }
--- a/crates/rag-rat-cli/src/commands/mod.rs
+++ b/crates/rag-rat-cli/src/commands/mod.rs
@@ -480,12 +480,8 @@ pub(crate) fn models(config: &Config, args: &ModelsArgs) -> anyhow::Result<()> {
 }
 pub(crate) fn reconcile(config: &Config, args: &ReconcileArgs) -> anyhow::Result<()> {
     let db = open_index(config)?;
-    // Force the legacy-f32 → int8 vector re-encode (#312) when asked, ignoring the run-once gate —
-    // for users who want it now on a huge index. Format-only, idempotent.
-    if args.reencode_vectors {
-        let converted = db.reencode_legacy_vectors_now()?;
-        eprintln!("rag-rat: re-encoded {converted} legacy f32 vector blobs to int8");
-    }
+    // `--plan` is a READ-ONLY dry run: it must return before any mutation, so it stays read-only
+    // even when combined with a mutating flag like `--reencode-vectors` (#312).
     if args.plan {
         let plan = db.reconcile_plan()?;
         // `--plan` prints a human summary by default; the global `--json` switches to the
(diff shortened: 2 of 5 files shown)


skakri:
All three review points addressed (latest commit):

- **P2 (concurrent rewrite)** — the UPDATE is now guarded on the original blob: `SET vector_blob = ?1 WHERE chunk_id = ?2 AND model_id = ?3 AND vector_blob = ?4`. If a reconcile re-embeds the row between the read and the write it no longer matches (new writes are int8), so the UPDATE is a no-op and the fresh embedding is preserved instead of being clobbered with the stale f32 vector. Only rows where the UPDATE affected exactly one row count as converted.
- **P2 (rescan)** — replaced the bare `LIMIT` with a `(chunk_id, model_id)` keyset cursor (`AND (chunk_id, model_id) > (?1, ?2) ORDER BY chunk_id, model_id`), so each batch resumes via the PK index instead of rescanning the table head. Termination is now cursor-driven; the old `converted == 0` break is removed.
- **P2 (forced gate)** — `reconcile --reencode-vectors` now sets `vector_int8_reencode_done` after a successful run, so the next maintenance pass doesn't full-scan to find nothing.

+3 tests: cursor walk across multiple chunk_ids × model_ids with a tiny batch size, concurrent-rewrite skip, and forced-path-sets-gate. Full suite green (927 / 930 `--features eval`), clippy + fmt clean both configs.

skakri:
Round-2 review points addressed (latest commit):

- **`--plan` read-only** — the `if args.plan { … return }` early-return now comes *before* the `--reencode-vectors` block in `reconcile()`, so `reconcile --plan --reencode-vectors` reports the plan without mutating the index.
- **maintenance budget** — the automatic re-encode in `run_maintenance_pass` is now skipped when `max_seconds == 0` and otherwise bounded by the same shared deadline (`started + max_seconds`) as the embedding reconcile. To stay resumable without re-introducing the rescan, the `(chunk_id, model_id)` keyset cursor is **persisted to reconcile_meta after each committed batch** and reloaded on the next call; a deadline stop returns `completed=false` and leaves the run-once gate UNSET, so the next pass resumes from the persisted cursor. The done-gate is set + the cursor cleared only on full completion. The deadline is checked after each committed batch (never mid-batch), and the cursor is saved post-commit (so a crash leaves it lagging, never ahead). Explicit `--reencode-vectors` runs unbounded.

+2 tests (deadline-stop → gate unset + cursor persisted → follow-up resumes to completion; budget-fits happy path sets the gate). Full suite green (929 / 932 `--features eval`), clippy + fmt clean both configs.

skakri:
Both round-3 points are fixed — and rather than patch them in isolation, I ran an exhaustive multi-lens review of the whole diff to catch everything at once. The two flagged here plus six more are addressed in the latest commit:

- **`--reencode-vectors` fall-through** → the flag now short-circuits (re-encodes and `return`s before any embedding reconcile), matching its 'no model inference' help, and emits `{"reencoded_vectors": n}` under `--json`.
- **Already-expired deadline** → the deadline is now checked at the TOP of the loop, so an exhausted budget does zero write work (not one 4000-row batch).
- **(new, data-integrity)** the detector `length(vector_blob) = 4*embedding_dim` collided with `embedding_dim = 0` (a pre-migration-002 `Failed` row's empty `x''` blob) and would have rewritten `x''` into a bogus 4-byte blob — both detector sites now carry `AND embedding_dim > 0` (documented + tested).
- the auto maintenance re-encode now takes only `max_seconds/2` so fresh-content embedding isn't starved during a multi-pass conversion; the forced path honors `--max-seconds`; maintenance errors are logged + the count is in the report JSON; the test harness uses the prod `UNIQUE(chunk_id, model_id)` schema and the 'PK' doc comments are corrected.

Full suite green (930 core / 933 `--features eval` / 109 cli), clippy + fmt clean both configs.

skakri:
@codex review

PR #314, opened 2026-06-25
chore(ci): target workflow triggers to the crates they exercise

skakri:
## What

Make the PR-triggered CI workflows **crate-targeted** so an irrelevant change stops fanning out the whole matrix. The motivating case: a `rag-rat-cli`-only PR was triggering the full **oracle** resolution gate, which only tests the core engine's parser/graph.

## The change (pure trigger filters — no job logic touched)

| Workflow | Before | After |
|---|---|---|
| `oracle.yml` | `crates/**` (matches cli/mcp) | `crates/rag-rat-core/**` + Cargo + oracle tooling |
| `bench-pr-run.yml` | *(no filter)* | `crates/rag-rat-core/**` + Cargo (benches are core-only) |
| `eval.yml` | *(no filter)* | core **+ cli** + `evals/` + Cargo + nextest |
| `ci.yml` | *(no filter)* | `paths-ignore` `**/*.md`, `docs/**` (stays broad for code) |

**Net effect:** mcp-only PR → `ci` only · docs-only PR → nothing · cli-only PR → `ci` + `eval`, no oracle/bench.

## Why it's safe

`main` has **no branch protection / no required status checks** (verified), so a path-skipped run can't deadlock a merge — these filters only save runner minutes and review noise. Every filter still carries the shared triggers (`Cargo.toml`, `Cargo.lock`, `.config/nextest.toml`) and its own workflow file, so a dep bump or workflow edit re-triggers everything. The `workflow_run` followups (`oracle-pr-comment`, `bench-pr-track`) auto-skip when their upstream skips.

## Two judgment calls

- **`eval` includes `cli`** — the `--features eval` clippy runs *only* in `eval.yml`, so a cli eval-surface break would otherwise slip through.
- **`ci.yml` stays broad for all code** — it's the only `--workspace` lint+test gate; trimmed only for pure-docs.

Untouched: `bench.yml` (main-only), `release-plz`, release/dispatch workflows.

### DIFF
--- a/.github/workflows/bench-pr-run.yml
+++ b/.github/workflows/bench-pr-run.yml
@@ -6,8 +6,16 @@ name: bench-pr-run
 # Pattern mirrors Bencher's "Track Benchmarks with GitHub Actions" (fork PRs) docs.
 
 on:
+  # The criterion + iai benches live in crates/rag-rat-core/benches and exercise only core's
+  # index/query/clone paths — cli/mcp/docs changes can't move the numbers, so they skip (and the
+  # paired bench-pr-track tracker, gated on this workflow_run, then skips too).
   pull_request:
     types: [opened, reopened, edited, synchronize]
+    paths:
+      - 'crates/rag-rat-core/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - '.github/workflows/bench-pr-run.yml'
 
 # Fork-safe by construction: no secrets are used here, and the token is read-only.
 permissions:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,17 @@
 name: CI
 
 on:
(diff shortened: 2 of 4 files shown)


skakri:
Both fixed (latest commit `a651454`):

- **eval fixtures** — added `tests/fixtures/**`. You're right: the `eval_suite` loads them via `fixture_root()`, and main CI runs core `--no-default-features` so it never compiles the eval module — eval.yml is the only gate on those fixtures.
- **oracle cli** — re-added `crates/rag-rat-cli/**`. Confirmed `tools/oracle-run.sh` runs `rag-rat index --full` + `rag-rat oracle report`, so the gate genuinely exercises the cli entry points, not just core. The `index` handler lives in the shared `commands/mod.rs`, so there's no clean surgical subset — the honest scope is core + cli (the net narrowing is now just 'drop the mcp crate'). Rationale comments corrected on both.

PR #315, opened 2026-06-25
feat(eval): track commit-replay search recall in Bencher on main

skakri:
## What

Wire the commit-replay search-quality eval (#120) into the existing Bencher pipeline as a **main-only tracked signal** — the search-**quality** counterpart to the iai/criterion **performance** signals. Completes the "wire eval into CI" work.

## Why main-only + tracking (no `--err`)

Measured: a 200-case at-head replay is **~8–10 min** and the metric is **noisy** — the replay gold is each commit's changed paths, so a single PR's recall delta is usually below the per-case noise floor (≥200 cases needed for signal). A per-PR hard gate would be both too slow for the critical path and mostly false-alarm.

Instead `recall@3` carries a statistical **lower**-boundary `t_test` threshold (a recall regression is a *drop*) that **alerts** on a real regression in Bencher but never hard-fails the run — the same "headline signal, not a gate" stance the repo already documents for the iai/criterion series and oracle's heavy push.

## The change

- **`tools/eval-report-bmf.py`** — `EvalReport` JSON → Bencher Metric Format, mirroring `tools/oracle-report-bmf.py`. One `search_replay` benchmark, `recall_at_3` / `recall_at_10` / `mrr_at_10` as fractions in [0,1] (the same numbers `eval` prints).
- **`bench.yml`** — new `eval_recall` job (main / `workflow_dispatch`): builds an embedded index at HEAD (`index --discover` → `models install fastembed-all-minilm-l6-v2` → `reconcile`), runs `rag-rat --json eval --replay --replay-max-cases 200`, converts to BMF, pushes via `bencher run --adapter json`. recall@3 gets the lower-boundary threshold; recall@10 + mrr@10 are tracked. Two CI footguns handled: **`fetch-depth: 0`** (the replay walks git history; a shallow checkout → zero cases) and the cached MiniLM model (`RAG_RAT_MODEL_CACHE`, mirrors `eval.yml`).
- **`docs/bencher.md`** — documents the signal + both footguns.

**No Rust change** — `eval --json` already emits the structured `EvalReport` via `print_output`.

## Validation

- BMF converter tested on a real report + malformed input (clean error).
- **End-to-end from a fresh clone** (no `.rag-rat/` index, exactly what CI gets): `index → reconcile → replay → BMF` ran clean with non-zero recall (proves at-HEAD + history present + embeddings working). The `eval_recall` job is main-only (`if: refs/heads/main`), so it first executes on merge to main, not on this PR. Only `bencher run` itself isn't locally testable (needs the token) — it mirrors the working oracle invocation exactly.

Independent of #314 (CI targeting) — different files, no conflict.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -86,3 +86,76 @@ jobs:
             --threshold-upper-boundary 0.95 \
             --thresholds-reset \
             "cargo bench --no-default-features --bench index_time"
+
+  # Search-quality headline (#19): commit-replay recall over the self-index at HEAD, tracked in
+  # Bencher. recall@3 is the metric an agent actually feels — "did the right chunk land in the first
+  # 3 reads". The 200-case replay is ~8-10 min, so it lives HERE on main (post-merge), never on the
+  # PR path: it would be both too slow to gate and too noisy to gate (the replay gold is each
+  # commit's changed paths, so a single PR's recall delta is usually below the per-case noise floor).
+  # Tracking only, NO `--err` — a real regression ALERTS via the statistical lower boundary on
+  # recall@3; a noisy dip doesn't hard-fail the run. Same stance as the iai/criterion series above.
+  eval_recall:
+    name: search recall (commit-replay)
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    env:
+      # Stable, cacheable location for the MiniLM ONNX download (fastembed honours this env var;
+      # mirrors eval.yml). The model is immutable, so the cache key is static.
+      RAG_RAT_MODEL_CACHE: ${{ github.workspace }}/.rag-rat-models
+    steps:
+      - uses: actions/checkout@v5
+        # FULL history: the commit-replay derives its cases from git history (commit message = query,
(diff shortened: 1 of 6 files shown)


skakri:
Both fixed (latest commit):

- **replay gold → indexed paths** — `generate_replay_suite` now filters each case's path gold to `db.indexed_path_set()` (live `files` rows, `kind != 'deleted'`) and drops cases left with no indexed gold. So a commit touching `.github/**`, `tools/**`, or a root manifest no longer adds unretrievable gold to the denominator — recall measures search quality, not the file mix. (New `git_history::indexed_path_set` free fn + an `IndexDatabase` wrapper; parent-state replay already restricts gold to parent-existing paths, left as-is.) +1 unit test.
- **don't reset peer thresholds** — dropped `--thresholds-reset` from the recall `bencher run`. The `bench` job publishes its instruction/latency thresholds to the same `main`/`ubuntu-latest` branch+testbed, and `--thresholds-reset` removes every threshold not named in the run — so each job now manages only its own measure's threshold and resets nothing; `recall_at_3` is created/updated in place. Still no `--err`.

Build + eval-feature replay tests green; clippy + fmt clean both configs.

Issue #317, opened 2026-06-25
Native HTTP/Ollama embedding provider (phase 1)


skakri:
A new `ollama` (later generic `http`) backend on the embedder registry (`embedding_models.rs`), alongside `fastembed`. The only thing in the embedding hot path; native Rust HTTP. **Ships first** — unlocks local Ollama + any user-supplied URL (connect-mode) with no cookbook, no cloud.

Depends on the local-Ollama spike for the `/api/embed` contract.

## Scope
- [ ] `ollama` backend: `POST /api/embed`, batched.
- [ ] Config: `endpoint`, `model`, `dim`, `auth_env`, `batch_size`, `request_timeout_s`.
- [ ] **Dim validation** — refuse when the first response's length ≠ configured `dim`.
- [ ] **Freshness identity** — fold model id + endpoint (+ mode) into the reconcile freshness key.
- [ ] **Query-side embedding** — search embeds the *query* at search time, so it must use the SAME model+runtime as the chunks (one vector space). Split chunk-embed backend (can be remote) from query-embed backend (must be local + same model). For a remote chunk-embed, the query path uses a *local* same-model backend (local Ollama / fastembed); validate model+runtime consistency so offline search stays in-space.
- [ ] Reconcile integration — reuse the existing deadline/budget/keyset-cursor loop for timeout + partial-batch resume.
- [ ] Optional `remote-embed` cargo feature (keep the HTTP-client dep off a minimal build — dependency surface; rag-rat is native throughout).

## Acceptance
- `reconcile` against a local Ollama embeds the index; results retrievable via search.
- Dim mismatch is rejected with a clear error.
- Changing model/endpoint forces a clean re-embed (no silent mixing of vector spaces).
- **Offline semantic search works after a remote bulk-embed**: queries embed locally with the same model, in the same vector space (no live endpoint needed at query time).

PR #320, opened 2026-06-25
refactor(embed): extract index/ai/providers/ — single resolution chokepoint (#317 task 1)

skakri:
**Task 1 of 7** toward the native HTTP/Ollama embedding provider (#317) — the foundation refactor everything else lands on.

## What
Extract a new `index/ai/providers/` directory that becomes the home of **all** embedding-provider code, so the centralization is structural ("no provider code outside this dir" is true by construction):

- `providers/mod.rs` — the resolution **chokepoint**: the `Embedder` trait + `embedder_for_spec` dispatch + `active_embedder` resolution (**signatures unchanged**), the feature-missing consts, and `MockEmbedder`.
- `providers/{fastembed,model2vec,hash}.rs` — the existing impls, one backend per file.
- `ai/mod.rs` curates the surface via `pub use providers::{…}` (mod.rs as the index, `rust-modern-style`) — **no transition shim**.
- `helpers.rs` keeps `embed_query`/`embed_query_with`, now resolving `active_embedder` from its new home.

## Why
The plan splits chunk-embed (can be remote) from query-embed (must be local) and adds an Ollama backend. Doing that safely needs a single place that constructs embedders / selects a backend. This PR makes that place exist first, as a pure no-behavior move, so the Ollama work (later tasks) lands on a clean seam.

## Verification — byte-green
- `cargo nextest run -p rag-rat-core`: **930 → 930** passed (identical), 0 skipped; **934** `--features eval`.
- `cargo build -p rag-rat-core --no-default-features`: clean.
- `cargo clippy --all-targets -- -D warnings` **and** `--no-default-features`: clean.
- `cargo +nightly fmt`; `cargo build --workspace`: clean.

Pure mechanical move — no behavior change.

### DIFF
--- a/crates/rag-rat-core/src/index/ai/helpers.rs
+++ b/crates/rag-rat-core/src/index/ai/helpers.rs
@@ -1,5 +1,5 @@
 use super::*;
-use crate::embedding_models::{Backend, EmbeddingModelSpec, spec};
+use crate::embedding_models::{HASH_MODEL_ID, spec};
 
 pub(crate) fn embed_query(
     conn: &Connection,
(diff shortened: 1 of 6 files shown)


skakri:
Fixed in #323 — `index::ai::MODEL2VEC_HF_REPO` is restored. The const stays in `providers/model2vec.rs` (where it belongs, a `Model2VecEmbedder` construction detail) and is added to the existing feature-gated re-exports (`providers/mod.rs` → `ai/mod.rs`). It's re-exported **feature-gated** rather than the pre-move unconditional: a model2vec const is only meaningful with the `model2vec` feature, so it now exists only in builds that use it.

PR #323, opened 2026-06-25
fix(embed): restore public index::ai::MODEL2VEC_HF_REPO (#320 review)

skakri:
Addresses the Codex P2 on the merged #320 (providers/ extraction).

**The regression:** the move put `MODEL2VEC_HF_REPO` into the now-private `providers::model2vec`, and `ai/mod.rs` re-exported only `Model2VecEmbedder` — so the public `index::ai::MODEL2VEC_HF_REPO` const (published-crate API) vanished. Silent public-API change in a 'no-behavior' refactor. (Practical impact ~nil — its sole user is `Model2VecEmbedder` — but worth fixing to keep the refactor honest.)

**Fix:** keep the const **in `model2vec.rs`** where it belongs, and add it to the existing feature-gated re-exports (`providers/mod.rs` → `ai/mod.rs`), restoring `index::ai::MODEL2VEC_HF_REPO`. Re-exported **feature-gated** (not the pre-move unconditional): a model2vec const is only meaningful with the `model2vec` feature, so it now exists only in builds that use it — no dead-code-exemption gymnastics, and arguably more correct than the pre-move unconditional wart.

**Verified:** build default + `--features model2vec` + `--no-default-features`; clippy `--all-targets` clean both configs; full core suite; fmt.

### DIFF
--- a/crates/rag-rat-core/src/index/ai/mod.rs
+++ b/crates/rag-rat-core/src/index/ai/mod.rs
@@ -15,8 +15,6 @@ pub(crate) use policy::*;
 pub use providers::FastEmbedEmbedder;
 #[cfg(test)]
 pub use providers::MockEmbedder;
-#[cfg(feature = "model2vec")]
-pub use providers::Model2VecEmbedder;
 pub(crate) use providers::active_embedder;
 // Curate the provider surface onto the `index::ai` path so existing `super::*` callers
 // (`reconcile`, `status`, `helpers`) and the external `ai::Embedder` /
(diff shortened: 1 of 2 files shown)


skakri:
Addressed in #324 (folded into the in-flight Task-3 PR rather than a 4th standalone). You're right that the gated re-export still dropped `index::ai::MODEL2VEC_HF_REPO` in `--no-default-features` builds. Fixed by gating at the **dependency boundary**: the `model2vec` module + the dep-free `MODEL2VEC_HF_REPO` const are now unconditional, and only the parts that need `model2vec-rs` (the imports + `Model2VecEmbedder` struct/impls) are `#[cfg(feature = "model2vec")]`. The public path is restored in all builds, and the const stays in its own file.

PR #342, opened 2026-06-30
feat(embed): auto-tune ephemeral remote embedding concurrency

### LINKED ISSUE #328: feat(cookbook): auto-tune embedding throughput params at provision time (GPU→params)
When the cookbook provisions a remote embedding box (#318), it should emit the throughput params rag-rat uses against that box — not leave the user to guess concurrency / batch / `OLLAMA_NUM_PARALLEL`.

## It's `(GPU, model) → params`, not `GPU → params`

The optimal params depend on the model's per-text cost as much as the GPU. Measured: all-minilm never exceeded ~40% GPU util and capped at ~250 texts/s on a T4 — Ollama-request-bound, so for the model shipping today the params are nearly GPU-**independent** (~conc 16–32, `NUM_PARALLEL` 16–32, batch 256). The mapping only starts to bite once we add a heavier embedding model where the GPU is the real constraint. Key it on both.

## The params span server AND client

The cookbook sets `OLLAMA_NUM_PARALLEL` on the box it provisions **and** emits the matching client concurrency + batch into the `[local_ai.embedding.remote]` config rag-rat reads. One mapping, two consumers.

## Approach: auto-tune > static table

A hand-calibrated `(GPU,model)→params` table goes stale and needs per-hardware calibration. Instead:

1. After the box boots, run a ~30s throughput **micro-sweep** (ramp client concurrency, find the throughput knee — same shape as the saturation sweep run during #317 validation).
2. Pick the params at the knee, emit them into the remote config.
3. **Cache per `(GPU, model)`** so ephemeral re-provisions skip re-tuning.
4. A small static seed table is the fallback / sweep starting point.

## Depends on

- #327 — the client-side concurrency mechanism + `concurrency` config field. The params are meaningless until rag-rat can act on them.

Part of #318 / #319.

skakri:
## Summary

Closes #328.

For ephemeral cookbook-provisioned remote Ollama boxes (Modal / RunPod), rag-rat now **auto-tunes how many `/api/embed` requests it keeps in flight** during the bulk reconcile, instead of running the whole thing at a fixed guess. After the box verifies it's serving, rag-rat runs a short client-concurrency micro-sweep against it, picks the throughput knee, and uses that for the live reconcile fan-out.

### Cap model

`[remote] concurrency` is the user's **hard cap**, not a target. The tuner only ever optimizes *within* it:

- the tuned knee is clamped to the cap,
- the box's `OLLAMA_NUM_PARALLEL` is set to the cap, and
- the persisted config is never overwritten.

Lower `concurrency` to tighten the cap; the tuner re-tunes under the new limit.

### Tuning lives in rag-rat, not the cookbook recipe

The sweep runs the **real `OllamaEmbedder`** ([`throughput_tune.rs`](crates/rag-rat-core/src/index/ai/throughput_tune.rs)), so the measured request shape (num_ctx, batch size, char budget) matches what reconcile actually sends. The previous approach re-implemented that request inside the TS recipe and drifted from the real one on every knob — the root cause of a long tail of review churn.

- The sweep engine is **backend-agnostic**: a generic `run_sweep` over the `Embedder` trait plus a per-backend adapter, so the planned Infinity / vLLM backends reuse it.
- The chosen knee is cached in the index's `index_meta`, keyed by `runtime + provider + GPU + model + request-shape`, with a TTL.
- Env knobs: `RAG_RAT_TUNE_MS` (sweep budget), `RAG_RAT_TUNE_TTL_MS` (cache TTL; `0` = re-sweep every time), `RAG_RAT_DISABLE_TUNING=1` (skip the sweep, use the cap).

### Cookbook is now a dumb provisioner (`@rag-rat/cookbook` 0.4.0)

The recipe sets `OLLAMA_NUM_PARALLEL` straight from the cap (`ollama_num_parallel`) and emits only the endpoint (+ optional auth) on `ready`. It no longer sweeps or carries request-shape knobs. A legacy recipe's `throughput` field on `ready` is a tolerated unknown key on the Rust side, so an older published recipe still works.

## Validation

- `cargo nextest run -p rag-rat-core -p rag-rat` — 1273 pass
- `cargo clippy -p rag-rat-core --all-targets` — clean
- `cargo +nightly fmt`
- `npm run typecheck` + `npm run build` in `cookbook`
- Live end-to-end: Modal (L4 / A100) and RunPod (A4000) with all-minilm and nomic-embed-text; every box torn down and verified leak-free. Empirically, ollama embedding over the cookbook proxy is overhead/server-bound (not GPU-bound), so a bigger GPU doesn't move the knee — the tuner converges to a low value and stops.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@ target/
 
 # Plans/specs are kept on disk for review, never committed (see dont-commit-plans).
 docs/plans/
+throughput-tune-rust-refactor-plan.md
+
+# RunPod e2e creds (never commit)
+.runpod-e2e.env
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -26,7 +26,7 @@ Rust parser is built against the exact event shapes below; implement them precis
 | spawn | runs `node recipe.mjs` (or `npx tsx recipe.mts`), with `RAG_RAT_COOKBOOK_INPUT` set in env | starts up |
 | input | sets `RAG_RAT_COOKBOOK_INPUT` = JSON `CookbookInput`; provider creds already in env / `~/.modal.toml` | reads + parses that env var |
 | progress | reads `status` / `log` events from **stdout** | emits `status` (provisioning → pulling → verifying) and `log` events as it works |
-| ready | reads the `ready` event → uses `endpoint` | once the box is up **and serving**, emits one `ready` event |
+| ready | reads the `ready` event → uses `endpoint` (+ optional `auth_token`) | once the box is up **and serving**, emits one `ready` event |
 | liveness | uses the endpoint | stays running, holding the box |
 | teardown | sends `SIGTERM` (or `SIGINT`) | emits a `tearing_down` status, destroys the box, exits `0` |
 | failure | sees an `error` event + non-zero exit, no `ready` | on provisioning failure: emits an `error` event, exits non-zero, **before** any `ready` |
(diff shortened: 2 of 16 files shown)


skakri:
@codex review

PR #433, opened 2026-07-04
test: multi-repo global-db integration matrix

### LINKED ISSUE #403: consolidation A8: multi-repo isolation integration matrix
End-to-end acceptance gate for the consolidation: two real git fixture repos indexed into one global database via the CLI binary.

Matrix: search/symbol/impact return only the querying repo's rows; memories isolated (create/search/doctor); clones per repo; gc of repo A leaves repo B byte-identical (row-count parity snapshot); `consolidate` of a third legacy repo lands its memories; concurrent index of A + memory writes on B both complete. Green on CI closes phase A.

skakri:
The phase-A exit gate: an end-to-end acceptance matrix for the global-DB
consolidation. Two or three real git fixture repos share **one** consolidated
global database, exercised through the built `rag-rat` binary (the assert_cmd
subprocess pattern the other CLI integration tests use), so the whole
`repo_id`-scoping surface goes red if any of it regresses.

Every scoping assertion pins **positive active-repo identity** — the returned
row's owning `repo_id`, a repo-distinct token, or the expected id — never a bare
count or the sibling's absence (which would still pass if an unscoped `LIMIT 1`
returned the sibling's row).

`crates/rag-rat-cli/tests/multi_repo_e2e.rs`, nine cases:

0. **Fixture identity** — two byte-identical repos still register as two distinct `repo_id`s (deterministic per-commit dates via the shared `common::git_commit` helper), the guarantee the rest of the matrix rests on.
1. **search / symbol / impact** repo-scoped — a same-named symbol resolves to the row *owned by the active repo*; impact for A's symbol is empty when scoped to B.
2. **Memories isolated** — identical titles coexist; `memory_search` / `memory list` / `memory doctor` key on the active repo's *own* memory id; A's `gone` anchor is never flagged by doctor B.
3. **Clones per repo** — each repo's `--recall-symbols` listing contains its *own* pair (present) and never the sibling's (absent); a precompute of B leaves A's clone generation byte-identical.
4. **gc parity** — gc of A leaves B's row counts identical across parents **and** child tables (`repo_memory_bindings`, `clone_edges`), and the reported `files_remaining` counts A alone.
5. **consolidate** — a third legacy per-repo index's memory lands in the global store, FTS-searchable immediately, and hidden from both pre-existing repos.
6. **Lock disjointness** — repo B's `WriteLock` is held explicitly while a full `index` of A runs; A acquires its own flock and completes, proving the per-repo locks are disjoint (a shared lock would hang A).
7. **Linked-worktree journey** — with no branch-local config, the linked checkout resolves through main's config and lists a *main-seeded* memory; a divergent branch config is ignored with a loud warning; `init` refuses and names main.
8. **dream** worklist repo-scoped — each repo references a path that exists only in the sibling, so a union-of-files stale-reference leak would suppress the finding.

Test hygiene: every case pins `RAG_RAT_DATA_DIR` (and `RAG_RAT_MODEL_CACHE`) to
per-test temp dirs so a keyless config never resolves to a real machine store;
the query surfaces the CLI does not expose are opened in-process against the same
global DB via an explicit `database` pin, race-free under the threaded runner;
all fixture commits route through the shared deterministic-identity helper.

Dream/reconcile scoping: settled — both `dream_findings` and `reconcile_attempts`
query predicates are already fully repo-scoped in production, so the matrix pins
the contract rather than fixing a leak. No production changes.

Gates: `cargo nextest run --workspace` 1598 passed and `--no-default-features`
1596 passed; `cargo clippy --all-targets` 0 warnings; two `taskset -c 0,1`
libtest rounds 9/9; `cargo +nightly fmt --check` clean.

Fixes #403

### DIFF
--- a/crates/rag-rat-cli/tests/multi_repo_e2e.rs
+++ b/crates/rag-rat-cli/tests/multi_repo_e2e.rs
@@ -1,0 +1,1055 @@
+//! Multi-repo global-DB integration matrix (memory-sync phase A8, issue #403 — the phase-A exit
+//! gate). Two (or three) REAL git fixture repos share ONE consolidated global database, driven
+//! end-to-end through the built `rag-rat` binary (the `assert_cmd`-style subprocess pattern the
+//! other CLI integration tests use). Every contract the A1–A7 workstreams landed — `repo_id`
+//! scoping on reads/sweeps, per-repo write locks, the generation axis, the consolidate importer,
+//! the governing-config seam + linked-worktree discovery, the identity gate — is exercised across
+//! the shared file, so the matrix goes RED if any of them regresses.
+//!
+//! ASSERTION-STRENGTH INVARIANT: every scoping assertion pins POSITIVE ACTIVE-REPO IDENTITY — the
+//! returned row's owning `repo_id`, a repo-distinct token, or the expected id — never merely a
+//! cardinality (`len() == 1`) or the sibling's absence. A count or an emptiness check would still
+//! pass if an unscoped `LIMIT 1` returned the SIBLING's row, or if a scoped read leaked and then
+//! happened to return the wrong single row; pinning identity is what closes that hole.
+//!
+//! FIXTURE IDENTITY: commits route through `common::git_commit` (shared helper, #435), which pins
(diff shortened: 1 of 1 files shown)


skakri:
@codex review

PR #470, opened 2026-07-07
fix(watch): gate the linux-only drain_until_quiet test helper behind cfg

### LINKED ISSUE #467: CI (cross-platform) red on main: dead-code `drain_until_quiet` on windows/macos
The **CI (cross-platform)** workflow is failing on `main` (first seen on the v0.14.0 release run, [28879193645](https://github.com/cq27-dev/rag-rat/actions/runs/28879193645)). Builds and tests pass on both windows-latest and macos-latest; only the two `clippy` jobs fail.

## Failure

```
cargo clippy --workspace --all-targets --no-default-features --locked -- -D warnings
error: function `drain_until_quiet` is never used
  --> crates/rag-rat-core/src/watch.rs:3182:8
  = note: `-D dead-code` implied by `-D warnings`   (lib test target)
```

## Root cause

`drain_until_quiet` is a `#[cfg(test)]` helper in `watch.rs`. Its only callers (lines ~3237, ~3574, ~3587) are the watch-**placement** tests, which are gated `#[cfg(target_os = "linux")]` (inotify-specific; cfg gates at watch.rs:3208 and :3546). On windows/macos those tests are compiled out, so the helper has zero callers → dead-code → `-D warnings`. Linux CI passes because there the callers compile and the helper is used. The sibling helper `drain_until_path_under` is unaffected — it also has callers in non-Linux-gated tests.

## Fix

Gate the helper with the same `#[cfg(target_os = "linux")]` as its exclusively-Linux callers (not a blanket `#[allow(dead_code)]`, which would hide genuinely-dead code later).

## Verification note

This does **not** reproduce by running clippy on Linux (there the callers compile, so the helper is live). Verify by dispatching the cross-platform workflow on the fix branch (`gh workflow run "CI (cross-platform)" --ref <branch>`) and confirming windows + macos clippy go green.

skakri:
## Root cause

The cross-platform CI (windows-latest / macos-latest) clippy job was red on `main`:

```
error: function `drain_until_quiet` is never used
  --> crates/rag-rat-core/src/watch.rs:3182:8
  = note: `-D dead-code` implied by `-D warnings`
```

`drain_until_quiet` is a `#[cfg(test)]` helper in `crates/rag-rat-core/src/watch.rs`. Its only callers live in watch-**placement** tests gated `#[cfg(target_os = "linux")]` (inotify-specific). On windows/macos those tests compile out, so the helper has no callers → dead code → the `--no-default-features --all-targets -- -D warnings` clippy leg fails. Linux CI stayed green because there the callers compile and the helper is live. The sibling `drain_until_path_under` was unaffected — it also has callers in non-Linux-gated tests.

## Fix

Gate the helper with `#[cfg(target_os = "linux")]` so it is present on exactly the platforms that use it, matching its exclusively-Linux callers. A targeted cfg rather than a blanket `#[allow(dead_code)]` keeps genuinely dead code detectable in the future. One-line change.

## Verification

- `cargo +nightly fmt` — clean.
- `cargo clippy --workspace --all-targets --no-default-features --locked -- -D warnings` on Linux — passes (unchanged; Linux was already green).
- Cross-platform CI (`CI (cross-platform)`, run 28882264798) — green: `clippy (windows-latest)` and `clippy (macos-latest)` both succeeded.

Closes #467

### DIFF
--- a/crates/rag-rat-core/src/watch.rs
+++ b/crates/rag-rat-core/src/watch.rs
@@ -3179,6 +3179,7 @@ mod tests {
 
     /// Drain real-watcher setup noise until the channel stays quiet for `quiet_ms`, capped by
     /// `max_ms`, so negative placement probes only observe events from the mutation under test.
+    #[cfg(target_os = "linux")]
     fn drain_until_quiet(
         rx: &std::sync::mpsc::Receiver<notify::Result<Event>>,
         quiet_ms: u64,


skakri:
Closing as redundant: the identical one-line fix already landed on main via #468 (squash-merged as a6df7aa). No diff remains against main.

Issue #491, opened 2026-07-08
memory validate: qualified-name relocation can rebind to the wrong same-name twin (struct vs impl)


skakri:
The gone-binding relocation arm in `validate_logical_symbol_binding` (`query/memory/validate.rs`) resolves by qualified name alone:

```sql
SELECT id, path FROM logical_symbols
WHERE qualified_name_id = (SELECT id FROM name_strings WHERE value = ?1)
  AND repo_id = ?2
LIMIT 1
```

No kind/signature discrimination and no ORDER BY — when one qualified name has same-name twins (a `struct` and its `impl` block; overloads with distinct signatures), which row wins is plan-order luck. A struct-bound memory can silently relocate onto the impl row (observed on the dogfood DB: a `FileLock`-bound memory whose stored `symbol_kind = struct` ended up pointing at the impl logical id and later validated gone).

Fix: the binding stores `symbol_kind` and `signature_hash` — use them. Prefer a kind-matching candidate (and signature-hash agreement via a members join where it discriminates further), with a deterministic ORDER BY tiebreak, so relocation is stable and lands on the twin the memory was actually bound to.

Test shape: struct+impl twins at one qualified name, impl inserted first (so the current unordered pick deterministically grabs the wrong row), binding stored with `symbol_kind = struct` and a dead id → validate must report `relocated` onto the struct row.

PR #525, opened 2026-07-08
feat(index): realign logical-symbol references on key-derivation drift

### LINKED ISSUE #493: logical-symbol ids: realign bindings on key-derivation drift, not just repo_id changes
A logical symbol's stable id hashes `(repo_id, language, path, name, qualified_name, kind, signature)`. Three inputs are parser/normalizer artifacts — when signature capture, kind classification, or qualified-name derivation changes (a grammar bump, a normalizer fix), every logical id in the repo can churn at once, stranding all logical-symbol memory bindings simultaneously. The relocation ladder then heals them one validate at a time (or leaks them to doctor).

`realign_logical_symbol_ids` (`index/graph_index.rs`) already contains the complete remap machinery — old→new id rewrite across every referencing table (`repo_memory_bindings`, `logical_symbol_members`, `logical_symbol_monikers`, call paths), two-phase through negative temp ids, idempotent — but it only runs on `repo_id` changes (V040 migration, adoption).

Proposal:
- Stamp a `LOGICAL_KEY_VERSION` in meta (the `GENERATED_FLAGS_VERSION` pattern): bump it whenever `LogicalSymbolKey`'s derivation inputs change semantics.
- On a version mismatch at rebuild, run a generalized realign: diff vanished vs appeared ids, match on `(path, name, kind)` + member content agreement, and reuse the existing two-phase remap — converting whole-repo stranding into an invisible in-place migration.
- Unmatched vanished ids fall through to the validate-time relocation ladder as today.

skakri:
A logical symbol's stable id hashes `(repo_id, language, path, name, qualified_name, kind, signature)`. Three of those inputs are parser/normalizer artifacts: a grammar bump or a signature-capture / kind / qualified-name derivation change churns every logical id in a repo at its next rebuild, stranding all logical-symbol memory bindings, call-path endpoints, and oracle monikers at once. `realign_logical_symbol_ids` can't help — it recomputes ids from the *stored* (old-derivation) field values, so it reproduces the old ids.

## Design

- **`LOGICAL_KEY_VERSION`**, stamped per-repo in `repo_meta` (the `graph_index_version` pattern): bump it whenever `LogicalSymbolKey`'s derivation changes semantics. Only a whole-corpus pass stamps — the full rebuild (when it carried no overlay rows; carried symbols are not reparsed) and the fresh-index standalone pass. Partial passes (incremental sweep, single-file heal, worktree overlay refresh) heal what is visible but defer the stamp: untouched files' drift is still in the future, and an early stamp would strand it permanently.
- **Heal inside `rebuild_logical_symbols`**: on a stamp mismatch, snapshot the *referenced* old rows before the clear (bounded by durable references, never the whole table), then realign each drifted id onto the new row matching `(path, name, kind)` with qualified-name OR signature agreement — the same discriminator evidence as the #491/#494 relocation twins. Group signatures count only when every member carries the same non-null capture. Multiple eligibles narrow to both-axes agreement; a kind-relaxed pass covers kind-classification drift; a strict winner carried by a single axis is vetoed when a cross-kind twin agrees on the axis it lacks (the decoy shape). Ambiguity and target collisions match nothing and fall to the validate-time relocation ladder — a confidently mis-anchored memory is worse than a flagged one.
- **Evidence survives partial passes**: the snapshot is memoized by a pass's first `remove_file_in_scope` — a partial pass deletes an edited file's old symbols (the signature evidence) before the logical rebuild runs, and one unhealed rebuild destroys the old rows forever. Idle passes remove nothing and never pay the scan.
- **Occupied ids**: an old id occupied by a different symbol's re-derived row realigns by evidence like a vanished one; without a winner its references are vacated to a sentinel so the ladder relocates them visibly instead of validating a live wrong id as healthy. In-place winners keep their references.
- **Discriminator refresh**: realigned references get their bind-time `binding_id` / `symbol_kind` / `signature_hash` rewritten from the row they now point at, so a later churn or relocation searches with current evidence. A moniker moving onto an id already occupied by a stale dangling row for the same tool displaces it instead of aborting the rebuild on the moniker PK.
- **Scale**: the candidate probe runs behind a transient `(repo_id, path, logical_name)` index — built for the heal, dropped before it returns — so a moniker-bearing repo (snapshot close to the whole table) doesn't go quadratic, and ordinary rebuilds pay no index maintenance.

## Tests

18 new tests covering: end-to-end drift heal (full and partial passes), the version gate, overload-twin ambiguity, qualified-name drift via signature agreement, kind drift via the relaxed pass, the cross-kind decoy veto, mixed-NULL and disagreeing member signatures, occupied-id realign/vacate/bijectivity (incl. moniker PK collisions and the dangling-occupant displacement), in-place winners surviving the vacate sweep, discriminator refresh, partial-pass stamp deferral (incremental + single-file heal), carried-overlay stamp deferral, and transient-index cleanup.

Fixes #493

### DIFF
--- a/crates/rag-rat-core/src/index/file_index.rs
+++ b/crates/rag-rat-core/src/index/file_index.rs
@@ -61,7 +61,9 @@ impl IndexDatabase {
             &text,
             &scope,
         )?;
-        self.rebuild_logical_symbols()?;
+        // Defer: a single-file heal must not stamp the logical-key version — every other file's
+        // drift is still in the future (#493).
+        self.rebuild_logical_symbols(graph_index::KeyVersionStamp::Defer)?;
         self.resolve_edges()
     }
 
--- a/crates/rag-rat-core/src/index/file_rows.rs
+++ b/crates/rag-rat-core/src/index/file_rows.rs
@@ -45,6 +45,11 @@ impl IndexDatabase {
         commit_sha: &str,
         worktree_id: &str,
     ) -> anyhow::Result<()> {
+        // #493: every live-scope file replacement funnels through here, so this is the seam that
+        // memoizes the drift-heal snapshot BEFORE the symbol deletes below destroy its
+        // member-signature evidence. A no-op unless the key-version stamp is stale, and paid at
+        // most once per pass.
+        self.capture_drift_snapshot_before_removal()?;
         let path = path_string(path);
         // Direct edges_data writes (#79): these statements touch up to every in-edge of a file's
         // symbols, so they must not pay the view triggers' per-row dictionary probes.
--- a/crates/rag-rat-core/src/index/graph_index.rs
+++ b/crates/rag-rat-core/src/index/graph_index.rs
@@ -141,12 +141,23 @@ pub(crate) fn realign_logical_symbol_ids(conn: &rusqlite::Connection) -> rusqlit
         }
     }
 
-    // Two-phase through negative temp ids: `stable_id` values are all >= 0 (`>> 1`), so a `-(i+1)`
-    // temp can never collide with a real id or with another temp. Phase 1 vacates every OLD id in
-    // the remap set; phase 2 lands the finals — so a new id equal to another remapped row's old
-    // id never trips the PK mid-pass. (A new id equal to an already-aligned row's id is a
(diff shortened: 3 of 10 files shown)


skakri:
@codex review

PR #526, opened 2026-07-08
Op-authoring helper + full backfill of existing memories (phase B, unwired)

### LINKED ISSUE #524: memory op-log: op-authoring helper + full backfill of existing memories (phase B, unwired)
Seventh phase-B op-log increment, after #513/#523 (persisted local device identity). With the op model, the signed stream-scoped store, and a stable local signing key all in place, this freezes the machinery that turns a memory into signed log entries — WITHOUT yet touching the live memory write path.

## Scope

- A lamport-allocating authoring helper: given a stream, the local device, and a `MemoryOp`, read the store's chain tail for that (stream, device), allocate the next lamport (genesis when empty), sign the entry under the local device, and append. The first time the store is driven by an allocator rather than hand-fed signed bytes.
- The memory→op translation: turn a persisted memory row into its `NodeCreate` (content), a `NodeStatus` when it is not active, and an `EdgeAdd` per typed node-edge. Code-anchor bindings are per-device derived state and stay OUT of the log.
- The one-time **full backfill**: for the active repo's owner stream, deterministically order the existing memories and author the above for each — idempotent (a second run adds nothing), gated so it runs at most once. This is the chosen posture: the log becomes the complete signed history from genesis.

All isolation-tested against a fixture database. Still **not** wired into `create_memory` / `update_memory` — that wire-up (and its transactional + failure-posture decisions) is the immediate follow-up, kept separate because it touches load-bearing code with a large test blast radius.

## Out of scope

- Wiring the helper into the live mutation entry points (the next increment).
- Multi-repo backfill beyond the active scope; key rotation; roster (#405); transport (#406).

Part of #404.

skakri:
Seventh phase-B op-log increment, after #480/#489/#495/#497/#500/#503/#504/#509/#511/#513/#523. Freezes the machinery that turns memories into signed log entries — the minting primitive, the row→op translation, and the one-time full backfill — WITHOUT touching the live memory write path.

## What landed

- **`store::author_op` / `store::author_batch`** — mint entries by allocating the next Lamport from the `(stream, device)` chain tail, signing, and inserting. `author_op` is one op in its own `IMMEDIATE` txn; `author_batch` is a whole sequence in ONE atomic txn (an interrupted backfill rolls back and re-runs clean) with a single projection re-fold at the end. `append`'s reproject branch is factored into a shared `reproject_after_write` helper. The primitives take a `&LocalDevice`, so `DeviceSecret` stays oplog-private.
- **oplog's first crate-internal API surface** — a curated `pub(crate)` re-export in `oplog/mod.rs` (the minting fns + the op vocabulary). Every submodule is otherwise private, so this is the ONE seam `query::memory` reaches through — and the only direction of the dependency (`oplog` never depends back on `query::memory`, which the next increment's reverse call would otherwise cycle).
- **`query::memory::authoring`** — translates a memory into `NodeCreate` [+ `NodeStatus` when non-active] [+ `EdgeAdd` per outgoing typed node-edge, in `edge_key` order]. Code-anchor **bindings** stay out of the log: they are per-device derived resolution state, re-validated locally, never part of the shared node graph. `backfill_memory_oplog` authors every one of the active repo's memories into its owner stream — **idempotent** (a completed atomic batch leaves a non-empty chain, so a re-run is a no-op), **scope-gated** (a no-op on an unscoped DB), deterministically ordered `(created_at_ms, id)`. The log becomes the complete signed history from genesis.

Still **not** wired to the live write path — `mod oplog` stays dead-code-allowed. The next increment calls `backfill_memory_oplog` + `author_op` inside each mutation's transaction (making `create_memory`/`update_memory` transactional), and reuses `memory_to_ops` for the mutations themselves.

## Tests / gates

- store: `author_op` mints a genesis→1→2 chain and projects; `author_batch` writes N ops in one txn with sequential lamports (empty batch is a no-op; continues a non-empty chain).
- authoring: `memory_to_ops` translates content + non-active status + edges in order; end-to-end backfill over a 3-memory fixture (one obsolete, one edge) yields 3 projected nodes + the right status + the edge, and a second backfill is a no-op; unscoped and empty-repo backfills are no-ops.
- Full workspace suite green (2040 tests), clippy clean on default and `--no-default-features`, nightly fmt clean.

Fixes #524

### DIFF
--- a/crates/rag-rat-core/src/oplog/identity.rs
+++ b/crates/rag-rat-core/src/oplog/identity.rs
@@ -29,17 +29,22 @@ pub(crate) struct LocalDevice {
 }
 
 impl LocalDevice {
-    /// The signing capability — the authoring seam signs entry bodies with this.
-    pub(crate) fn secret(&self) -> &DeviceSecret {
+    /// The signing capability — the authoring seam signs entry bodies with this. `pub(super)`: the
+    /// secret (and `DeviceSecret` itself) stay oplog-internal; the store's authoring path is the
+    /// only caller. Cross-module callers author under a whole `&LocalDevice`, never the raw secret.
+    pub(super) fn secret(&self) -> &DeviceSecret {
         &self.secret
     }
 
-    /// The verifying key (for a local self-verify or to hand a peer).
-    pub(crate) fn public(&self) -> DevicePublic {
+    /// The verifying key (for a local self-verify). `pub(super)` to match `DevicePublic`'s
+    /// oplog-internal visibility.
+    pub(super) fn public(&self) -> DevicePublic {
         self.public
     }
 
-    /// The opaque device fingerprint the op model + signed entry body carry.
+    /// The opaque device fingerprint the op model + signed entry body carry. `pub(crate)` — the
+    /// authoring seam in `query::memory` passes it to `chain_tail`; `DeviceFingerprint` is likewise
+    /// crate-visible.
     pub(crate) fn fingerprint(&self) -> DeviceFingerprint {
         self.fingerprint
     }
--- a/crates/rag-rat-core/src/oplog/mod.rs
+++ b/crates/rag-rat-core/src/oplog/mod.rs
@@ -40,3 +40,19 @@ mod op;
 mod project;
 mod store;
 mod stream;
+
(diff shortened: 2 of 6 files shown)


skakri:
@codex review

skakri:
@codex review

Issue #522, opened 2026-07-08
perf(reconcile): policy re-derivation parses every chunk per pass — trust the persisted embedding_policy


skakri:
The reconcile/maintenance paths re-derive the embedding policy from chunk **text** for large chunk populations, and every such derivation that reaches the low-signal gate re-parses the chunk with tree-sitter (`is_low_signal_chunk` → `parse_within_budget`: thread spawn + text copy + global memo probe per call):

- `embed_loop.rs::embedding_policy_skip_summary` streams **every chunk in the index** through `embedding_policy_for_chunk` — and per its own comment it "runs on every reconcile (and `reconcile_plan`)". On the kernel-sized example from that comment (~4.2M chunks) that is millions of parses and thread spawns per pass.
- `store.rs::estimated_reconcile_jobs` runs `policy_for_job` per candidate on the watcher/maintenance gate (`pending_embedding_jobs`).
- `store.rs::select_reconcile_batch` runs `policy_for_job` per batch candidate (bounded by batch size — least concerning).

The index already **persists** the decision: `chunks.embedding_policy` / `chunks.embedding_priority` are written at index time by `prepare_chunks` → `insert_chunks`. Two directions, not mutually exclusive:

1. **Read the persisted decision** where the caller only needs counts or a cheap eligibility gate (`skip_summary` could be a `GROUP BY chunks.embedding_policy`; the estimator could trust the column). Needs a staleness story: the persisted decision reflects the policy code at index time — a policy-version stamp (bump → re-derive lazily) would make trusting it sound.
2. **Make re-derivation cheap**: once the low-signal classification is span-based at index time (#516), the text-based fallback still parses. A policy-version-gated trust of the persisted flag avoids the parse entirely on the hot paths.

Either way, the goal: reconcile passes should not pay O(index chunks) tree-sitter parses to answer questions the indexer already answered and stored.

Related: #516 (removes the per-chunk parse at index time).

Issue #518, opened 2026-07-08
perf(index): heal path parses the same file up to 5× — share one parse like the prepare phase


skakri:
`file_index.rs::index_file` — reached via `heal_file`, the query-time lazy heal of a stale file — parses the same file text up to **five times**:

1. `parser::parse_error` (parse-failure flag),
2. `chunker::chunks_for_file` → `parse_symbols`,
3. `symbols::symbols_for_file` → `parse_symbols`,
4. `store_symbol_fingerprints` → `parser::parse_file`,
5. `edges::index_file_edges` → `syntactic_edges` → `parse_within_budget`.

Each parse also pays the worker-thread spawn + full-text copy in `parse_within_budget`. The full-rebuild and changed-mode paths already solved this: `prep.rs::prepare_index_content` parses once and feeds the shared tree to all five consumers.

Fix: split a text-based core out of `prepare_index_content` (it currently reads the file itself; `heal_file` already holds the text) and route `index_file` through it. Beyond the 5×→1× parse win, this removes real logic-drift risk — the generated-file fingerprint gate and the parser-failure message shape are currently duplicated between the two paths and must be kept in sync by hand.

Volume is low (per-file, triggered when a query sees a stale row), so this is more a structural unification than a hot-path fix — but it makes the heal path exactly as cheap and exactly as consistent as the bulk path.

PR #527, opened 2026-07-08
perf(index): shared-parse low-signal classification + O(1) chunker line spans

### LINKED ISSUE #516: perf(index): embedding policy re-parses every chunk — classify low-signal from the shared file parse
`ai/policy.rs::is_low_signal_chunk` re-parses **every chunk's text** with tree-sitter to decide the embedding policy. It runs inside `prepare_chunks`, i.e. for every chunk on every indexing path (full rebuild, changed-mode incremental, heal).

Each call goes through `parse_within_budget`, so per chunk we pay:

- a fresh OS thread spawn (the pathological-parse abandon path needs a `'static` worker),
- a full copy of the chunk text (`text.to_string()`),
- a SipHash of the text + a probe of the global `TIMED_OUT_PARSES` mutex (contended across the rayon prepare threads),
- a new `Parser` + grammar handle.

A repo producing ~100k chunks pays ~100k thread spawns per full rebuild, on top of the one legitimate parse per file that `prepare_index_content` already does and shares with chunks/symbols/edges/fingerprints.

The fix is the one already sketched in the function's own doc comment: **precompute the low-signal flag from the file's shared parse**. Chunk spans are known at prepare time — classify the named nodes overlapping each chunk's byte span (descend through container nodes that extend beyond the span, so a `use`-only chunk inside a `mod` body still classifies as plumbing) instead of parsing the chunk text against a `chunk.<ext>` pseudo-path.

Notes for the implementation:

- Keep the standalone text-based classification as the fallback for chunks with no shared tree (generated files, oversized files past `MAX_STRUCTURAL_PARSE_BYTES`, hard parse failures) so behavior there is unchanged.
- `embedding_policy_for_chunk` checks low-signal **last** (after SkipTooLarge / SkipGenerated / SkipTestFixture / language / SkipTooSmall), so the precomputed flag only needs to be consulted at that point.
- Existing `low_signal_tests` in policy.rs pin the text-based semantics; add span-based tests asserting parity on the same shapes (imports-only chunk, symbol chunk, mixed chunk, docstring-only chunk).
- Measure with `benches/rag_pipeline.rs` (iai instruction counts) and `benches/index_time.rs`.

### LINKED ISSUE #517: perf(chunker): line_span rescans the file from byte 0 per symbol — O(symbols × file bytes)
`chunker.rs::line_span(text, start_line, end_line)` converts a line range to a byte span by scanning the file **from byte 0** on every call. `code_chunks_for_symbols` calls it once per symbol, and `push_uncovered_chunk` once per uncovered gap — making chunking O(symbols × file bytes).

A 512 KB file (the structural cap) with ~2k symbols rescans on the order of a gigabyte of text to produce its chunks. The irony: every `ParsedSymbol` already carries exact `start_byte`/`end_byte` from tree-sitter; the scan exists only to snap spans to whole lines.

Fix: build **one line-offset table per file** (byte offset of each line start) and derive every span from it in O(1). `prepare_chunks` already splits the file's lines once for the anchor computation — same lesson, applied to the chunker. Per the no-shim convention, migrate `line_span`'s call sites to the table rather than keeping the scanning function as a wrapper.

Semantics to preserve exactly (pin with tests before refactoring):

- raw on-disk bytes in the span (CRLF preserved) so chunk byte offsets stay aligned with the tree-sitter symbol offsets `query::graph_meta` joins against;
- `None` for `start_line == 0`, `end_line < start_line`, or `start_line` past EOF;
- the whitespace-only span → `None` gate;
- `end_line` past the last line clamps to EOF.

Measure with `benches/rag_pipeline.rs` / `benches/index_time.rs`; the win concentrates in symbol-dense files.

skakri:
Two prepare-phase wins from the parse-pipeline efficiency milestone.

Fixes #516. Fixes #517.

## Low-signal embedding gate: one parse per file, not per chunk (#516)

`is_low_signal_chunk` re-parsed **every chunk's text** with tree-sitter to decide the embedding policy — inside `prepare_chunks`, on every indexing path. Each re-parse went through `parse_within_budget`, paying a worker-thread spawn, a full text copy, a hash + global-mutex probe of the timeout memo, and a fresh `Parser` per chunk.

Now the prepare phase classifies each chunk's **byte span against the file's shared parse** (`is_low_signal_span`): low-signal iff every named node the span covers is plumbing, descending into container nodes that extend beyond the span — so a `use`-only slice of a `mod` body classifies the way its standalone parse used to. The text-based path stays as the documented fallback for callers with no tree (generated / oversized / markdown / parse-failure files, the heal path pending #518, and the reconcile paths pending #522).

- `embedding_policy_for_chunk` takes a `LowSignalCheck`: `FromSpan` carries the shared root + chunk span, `FromText` is the fallback. The check is evaluated only at the final gate, so the cheaper skips (`SkipTooSmall` etc.) still short-circuit before any tree walk or re-parse happens.
- `prepare_chunks` threads the shared root through; `prepare_index_content` passes it, heal/reconcile pass `FromText`.
- Tests: span-twin coverage of the existing `low_signal_tests` shapes (imports-only, definition, docstring, `#include` vs `#define`, mixed, whitespace) plus the container-descent case, and a prep-level parity test asserting span- and text-classification produce identical policy decisions on the same chunks (with assertions that both outcomes occur, so parity can't pass vacuously).

## Chunker: O(1) line spans via a per-file line-offset table (#517)

`line_span` rescanned the file **from byte 0** to convert a line range to a byte span, and was called once per symbol plus once per uncovered gap — O(symbols × file bytes); a symbol-dense 512 KB file rescanned ~1 GB of text during chunking.

`LineOffsets` is built once per file (one pass over the bytes) and answers every span in O(1). Call sites migrated, `line_span` deleted (no wrapper shim). Exact semantics pinned by tests before the swap: raw on-disk bytes (CRLF preserved, so chunk offsets stay aligned with the tree-sitter symbol offsets `query::graph_meta` joins against), `None` for 0/inverted/past-EOF starts, EOF clamping, `split_inclusive` line-count semantics (no phantom trailing line). The existing CRLF byte-offset pins (#241) all pass unchanged.

## Benchmarks

iai-callgrind (`benches/rag_pipeline.rs`, cargo-resolver corpus subtree), branch vs an `origin/main` baseline on the same machine — deterministic instruction counts, 0 regressions:

| bench | instructions | vs main |
|---|---|---|
| `index` (full rebuild) | 1,551,624,214 | **−5.52%** |
| `query_cold` | 146,843,690 | **−35.11%** |
| `query_warm` | 178,568,196 | **−18.04%** |

The query-side win is the chunker fix surfacing in hit validation, which re-derives a result file's chunks and previously paid the per-symbol rescan there too.

### DIFF
--- a/crates/rag-rat-core/src/index/ai/policy.rs
+++ b/crates/rag-rat-core/src/index/ai/policy.rs
@@ -17,6 +17,35 @@ pub(crate) fn needs_embedding(
         || chunk.embedding_text_version.as_deref() != Some(EMBEDDING_TEXT_VERSION)
 }
 
+/// How the `SkipLowSignal` gate classifies a chunk. Evaluated ONLY if the cheaper gates
+/// (`SkipTooLarge`/`SkipGenerated`/`SkipTestFixture`/language/`SkipTooSmall`) don't short-circuit
+/// first — a sub-80-char chunk must pay neither a re-parse nor a tree walk.
+pub(crate) enum LowSignalCheck<'a> {
+    /// Re-parse the chunk text (`is_low_signal_chunk`) — for callers with no tree at hand:
+    /// generated / oversized / markdown / parse-failure files, the heal path, the reconcile paths.
+    FromText,
+    /// Classify the chunk's byte span against the file's shared parse (`is_low_signal_span`,
+    /// #516) — one tree-sitter parse per file instead of one per chunk.
+    FromSpan { language: Language, root: tree_sitter::Node<'a>, start_byte: usize, end_byte: usize },
+}
+
+impl LowSignalCheck<'_> {
+    fn is_low_signal(
+        &self,
+        language: &str,
+        chunk_kind: &str,
+        symbol_path: Option<&str>,
+        trimmed: &str,
+    ) -> bool {
+        match self {
+            Self::FromText => is_low_signal_chunk(language, chunk_kind, symbol_path, trimmed),
+            Self::FromSpan { language, root, start_byte, end_byte } =>
+                is_low_signal_span(*language, *root, *start_byte, *end_byte),
+        }
+    }
+}
+
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn embedding_policy_for_chunk(
     path: &Path,
     language: &str,
@@ -25,6 +54,7 @@ pub(crate) fn embedding_policy_for_chunk(
     symbol_path: Option<&str>,
     text: &str,
     max_embedding_chars: usize,
+    low_signal: LowSignalCheck<'_>,
 ) -> EmbeddingPolicyDecision {
     let path_text = path.to_string_lossy();
     let trimmed = text.trim();
@@ -48,7 +78,7 @@ pub(crate) fn embedding_policy_for_chunk(
     if trimmed.chars().count() < MIN_EMBEDDING_CHARS {
         return policy("SkipTooSmall", 9, false);
     }
-    if is_low_signal_chunk(language, chunk_kind, symbol_path, trimmed) {
+    if low_signal.is_low_signal(language, chunk_kind, symbol_path, trimmed) {
         return policy("SkipLowSignal", 9, false);
     }
     policy("Embed", embedding_priority(&path_text, language, chunk_kind, symbol_path), true)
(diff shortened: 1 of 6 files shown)


skakri:
@codex review

Issue #530, opened 2026-07-09
perf(reconcile): fast-path embedding_policy_skip_summary via a full-rebuild version stamp


skakri:
Follow-up from #522 / PR #529. `embedding_policy_skip_summary` recomputes each chunk's policy from text (a tree-sitter parse for the low-signal gate) on every reconcile and plan. It cannot simply aggregate the persisted `chunks.embedding_policy`: that column is stamped at index time with `DEFAULT_MAX_EMBEDDING_CHARS`, and (for chunks predating migration 006, which had no backfill) defaults to `Embed`, so a `GROUP BY` would under-report the skips the authoritative `select_reconcile_batch` actually makes on a non-default `max_embedding_chars` or an older index. (An earlier cut of #522 shipped that `GROUP BY`; two review passes surfaced exactly those divergences and it was reverted.)

A correct fast path needs a **per-repo version stamp set only by a full rebuild** (mirror `graph_index_version` / `GENERATED_FLAGS_VERSION`): an incremental/changed pass re-inserts only changed files, so it cannot certify unchanged chunks are current-code-fresh — stamping there would reintroduce the bug. The summary would then read the aggregate only when `max_embedding_chars == DEFAULT_MAX_EMBEDDING_CHARS` AND the stamp matches the current policy version; otherwise recompute. Bump the version const whenever `embedding_policy_for_chunk` changes.

Alternative: an on-open `ensure_embedding_policy_current` backfill (re-derive over all chunks once per version bump) — but that pays the O(chunks) parse it removes, once. The stamp-on-rebuild approach avoids that at the cost of only activating after a `--full`.

Low priority: the summary is diagnostic-only, and #522 already removed the parse from the hotter watcher backlog gate (`estimated_reconcile_jobs`).

skakri:
Concrete finding from a parse-path audit: this summary is the **one remaining per-chunk re-parse** in the whole index/reconcile path. Everywhere else parses each file once and shares the tree — `prep::prepare_index_content_from_text` classifies low-signal via `LowSignalCheck::FromSpan { root }` off the single shared parse, and symbols/chunks/edges/clone-fingerprints all reuse that same tree.

[...]

PR #572, opened 2026-07-09
perf(reconcile): classify skip-summary low-signal from one shared parse per file

skakri:
## What

`embedding_policy_skip_summary` (the diagnostic explaining why chunks didn't embed) re-parsed **each chunk's text** with tree-sitter to derive its low-signal policy — O(chunks) parses on every reconcile, and chunks overlap so overlapped regions were parsed multiple times over.

It now classifies low-signal from **one shared parse per file** (`FromSpan`, the same #516 index-time path): group chunks by file, reconstruct the file text from the chunks' verbatim substrings, and — **only when that reconstruction hashes to the stored `files.sha256`** — parse each file once and classify every chunk's span. O(files) parses instead of O(chunks).

## Correctness

`chunk_text` is line-ending-**normalized** (the chunker strips `\r` and re-adds `\n` while advancing byte offsets over raw lengths), so `raw_len == end − start` is an LF-repo coincidence, not an invariant. The `files.sha256` verification is what makes reconstruction safe: any file whose rebuilt text doesn't hash-match (CRLF / no-trailing-newline / older-chunker rows) falls back to per-chunk `FromText`. Generated / markdown / oversized files also fall back; memory stays bounded to one structural file's chunks (an oversized file flips to streaming per-chunk classification immediately).

## Tradeoff (deliberate)

The summary now mirrors prep's #516 low-signal classification rather than the embed path's `FromText`. The two agree except for a chunk that slices into a long comment/string (where `FromSpan` treats the sliced leaf as plumbing). The summary is diagnostic-only — nothing gates on it — so this is acceptable; documented inline.

## Tests

- `reconstruct_file_text` unit tests: overlap, coverage gap, multibyte boundary.
- End-to-end: the shared-parse summary equals the per-chunk `FromText` reference on a real index — including a **CRLF file**, which exercises the sha-verify fallback (without it the reconstruction would silently corrupt the diagnostic).

Refs #530 — complementary to its version-stamp caching (this makes the recompute cheaper; the stamp skips it when the index is certified unchanged).

### DIFF
--- a/crates/rag-rat-core/src/index/ai/policy.rs
+++ b/crates/rag-rat-core/src/index/ai/policy.rs
@@ -57,6 +57,43 @@ impl LowSignalCheck<'_> {
     }
 }
 
+/// The policy from the cheap, PARSE-FREE gates that precede the low-signal check (`trimmed` is the
+/// pre-trimmed chunk text), or `None` when the chunk REACHES the low-signal gate — the only gate
+/// that needs a tree-sitter parse. A caller that shares one parse across a file's chunks uses this
+/// to skip the parse entirely when no chunk reaches the low-signal gate.
+pub(crate) fn cheap_skip_policy(
+    path: &Path,
+    language: &str,
+    file_kind: &str,
+    chunk_kind: &str,
+    symbol_path: Option<&str>,
+    trimmed: &str,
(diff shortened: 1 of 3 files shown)


skakri:
@codex review

Issue #502, opened 2026-07-08
watcher: a HEAD move (pull / branch switch) triggers a full-repo re-derive pass — ~15 min at 100% of one core


skakri:
Observed live on the dogfood DB while working in the repo: the elected watcher's maintenance thread (`rag-rat-watch`) pinned one core at ~100% for tens of minutes, twice in one evening, each burst starting right after ordinary git activity (`git pull --ff-only`, branch checkouts). Not a hang — each pass completed on its own; but the accumulated cost is real: the MCP server's process had ~6.3 CPU-hours over ~26 h of uptime (~24% of a core, continuously).

Evidence pinned so far:

- `reconcile_attempts` shows the passes re-deriving the WHOLE repo's chunk set on every HEAD move: full re-chunks of ~6.8–7.5k chunks at 14:55, 15:07, 19:15, and 19:23, the last two bracketing a `pull --ff-only` and a branch checkout minutes apart. Interleaved passes for real edits are the expected ~10–150 chunks.
- The embedding stage is NOT the burn (8–80 s per pass, vectors served from the content-addressed `embedding_cache`). The CPU goes to the unlogged stages of a full-scope pass: tree-sitter parse + edge extraction + `resolve_all_edges` (~48k symbols / ~173k edges here), re-chunking, and blame — single-threaded on the watcher thread, hence exactly one core.
- The clone graph is NOT the culprit: one Complete generation (20 s build), healthy delta counter — the #479/#490 delta machinery is doing its job.

Suspected root cause: `files` rows are keyed by `(path, commit_sha, worktree_id)`, so a HEAD move re-keys the base scope and the pass re-derives the new commit scope wholesale, even though nearly all content is byte-identical with the previous scope. A branch switch or pull on a mid-size repo should cost roughly its diff, not a full reindex.

Angles worth exploring:

- Content-hash reuse across commit scopes: when a path's sha256 matches the previous base scope's row, carry the derived rows (chunks/symbols/edges) into the new scope instead of re-deriving them.
- If wholesale carry is too invasive, at least skip re-parsing/re-resolving unchanged files (the per-file sha check exists; it is the scope re-key that defeats it today).
[...]

PR #579, opened 2026-07-10
perf(watch): event-scoped worktree-overlay refresh with a per-worktree diff basis

### LINKED ISSUE #577: perf(watch): maintenance pass sweeps every linked worktree overlay every pass — CPU/write treadmill with concurrent agent worktrees
## Symptom

On a dev box running several concurrent coding agents in `.claude/worktrees/` checkouts of this repo, the elected watcher (`rag-rat mcp`) is a permanent CPU + disk treadmill:

- samply on the live process: 96.6% of on-CPU time in `rag-rat-watch-p` — `watch::maintenance_pass → run_pass` 54.4% inclusive, `sqlite3_step` 31.4%, `ts_parser_parse` 30.8%, `refresh_worktree_overlays` 26.5%, `apply_clone_graph_delta` 17%, embedding reconcile 15%.
- `/proc/<pid>/io`: **34.5 GiB cumulative writes** to `rag-rat.sqlite` + `-wal` — the database itself is 632 MB. The same data is being rewritten over and over (WAL appends + checkpoints), not growth.
- From one watcher's log over ~16 h: **274 maintenance passes, 15,022 s (4.2 h) of pass wall-clock** — a 26% duty cycle.

The repo has 8 live linked worktrees (6 agent worktrees under `.claude/worktrees/`, 2 external), diverging from the base checkout by 12–260 files each (543 overlay file rows in total).

## Measured pass anatomy (from the #502 stage timings)

Fully idle, 3 AM, nothing running anywhere, passes every 300 s (`periodic_sweep_secs` backstop):

```
maintenance pass (tail skipped) total_ms=3119 content_changed=false overlays_changed=false stages=discover=72ms overlays=3047ms wal=0ms
```

**98% of an idle pass is the overlay sweep.** Interleaved with each idle sweep: 7–8 single-text `embed request … texts=1` HTTP round-trips against the local query endpoint — one per linked worktree, every pass, all night.

A pass where only overlays changed (base untouched):

```
maintenance pass total_ms=18317 content_changed=false overlays_changed=true stages=discover=69ms overlays=8671ms reconcile=6940ms clone_delta=2603ms memory_validate=29ms wal=0ms
```

Per-hour histogram (one watcher process): idle hours run 11–12 passes/hour at ~40 s/hour; active hours run 17–41 passes/hour at **880–2,772 s/hour** of pass time, with `overlays_changed=true` on most passes (24 of 27 in the worst hour).

## Root cause — three stacked factors

**1. Every pass sweeps every linked worktree (`watch.rs::refresh_worktree_overlays`).**
The per-worktree "delta-only, idle-safe" refresh is only write-idle. Per worktree per pass it still pays:
- a gix tree diff, base HEAD ↔ linked HEAD (`worktree_overlay.rs::compute_linked_worktree_delta`),
- a full working-tree status walk,
- `IgnoreMatcher::compile` over the linked checkout ("recompiled each call"),
- `overlay_needs_embed` → `pending_embedding_jobs_with_available_incremental_embedder`, which **acquires a live embedder (an actual embed HTTP request) and then runs the O(scope) candidate scan** — per worktree, per pass, even when `this_changed=false` and there is no backlog (`ai/reconcile/status.rs:67`).

That is ~400–450 ms per worktree on this repo → the 3.0 s idle pass, 12×/hour, 24/7.

**2. Base↔branch diff as the overlay basis couples every checkout's churn to every other.**
Overlay candidates are the *committed branch diff against the base checkout's current HEAD* ∪ working-tree status. So a commit on the **base** checkout changes the diff basis for **all 8 worktrees at once**: the next pass re-reads and re-indexes the branch-side content of every file that commit touched, in every overlay (tree-sitter parse + row writes ×8). Each changed overlay then runs `finalize_overlay_refresh` inside its transaction — including **a global `rebuild_logical_symbols` (26 k rows dropped + rewritten) and `sync_fts`, once per changed worktree per pass**. Six agents committing/editing concurrently means `overlays_changed=true` on nearly every pass, in both directions (base→overlays, agent→own overlay).

**3. Any overlay change forces the full tail.**
`should_run_pass_tail` treats `overlays_changed` like `content_changed`: base embedding reconcile (up to `PASS_RECONCILE_MAX_SECONDS=60`), `apply_clone_graph_delta` (2.6 s here), `memory_validate` — so the tail runs on nearly every pass during working hours. Combined with triggers from three sides (periodic sweep every 300 s; debounced inotify events across all 9 checkouts; the git hooks firing `rag-rat maintenance` on every commit/checkout in any of them), the write treadmill follows: passes 10–500+ s, dozens per hour, ~2 GiB/h of SQLite rewrite at peak.

## Proposed fix

Keep the overlay feature (worktree-scoped queries are exactly what agent worktrees use); make the per-pass work proportional to what actually changed:

1. **Event-scoped overlay refresh.** The event loop already classifies events per worktree (`event_requests_maintenance` + `LinkedWorktreeWatches`); accumulate the set of event-origin roots while the debounce is armed and carry it in the `PassRequest`. `refresh_worktree_overlays` refreshes only: (a) worktrees in that set, (b) worktrees whose recorded refresh basis (base HEAD, linked HEAD) no longer matches — which is what makes a base commit correctly refresh all overlays — persisting the basis per `worktree_id` after each successful refresh. The **periodic sweep keeps refreshing everything** (it exists to catch missed events; liveness is unchanged — anything missed heals within `periodic_sweep_secs`). The hook-driven `rag-rat maintenance` scopes to the worktree it fired from plus basis-invalidated ones.
2. **Stop probing for embed backlogs on unchanged worktrees.** Replace the per-pass per-worktree embedder probe + candidate scan with a persisted "overlay embed backlog owed" marker (same pattern as `watch_shutdown_reconcile_pending`), set when an inline overlay reconcile returns `Partial`/blocked and cleared when it drains. Unchanged worktree + no marker → zero embed/scan work.

Expected effect on this workload: idle passes drop from ~3 s to ~discover-only (~70 ms) with zero HTTP; active passes refresh 1 worktree instead of 8; the global `logical_symbols`/FTS rebuild runs for actually-changed overlays only. The base-commit case (all overlays legitimately need refresh) is unchanged.

Not in scope, considered and set aside:
- Excluding `.claude/worktrees/**` from overlay tracking (config knob): breaks worktree-scoped queries for exactly the checkouts that use them most; scoping the refresh removes the waste without losing the feature.
- WAL/checkpoint tuning: #573/#560 territory; shaves the margin, doesn't touch the re-fire economics.

Related: #219 (overlay feature), #506 (pass worker), #472 (clone quiet gate — same "defer under churn" shape), #482/#573 (WAL hygiene), #63 (idle backstop), #502 (the stage timings used here).

skakri:
Fixes #577.

## What

`refresh_worktree_overlays` refreshed **every** live linked worktree on **every** maintenance pass. The per-worktree refresh is write-idle when nothing changed, but never free: each one pays a base↔linked tree diff, a full working-tree status walk, an `IgnoreMatcher` compile, and — via `overlay_needs_embed` — an embedder acquisition (a live embed HTTP request) plus an O(scope) pending-jobs scan. On a machine running several concurrent agent worktrees that made every pass cost scale with the fleet instead of with what changed: ~3 s of overlay sweep per otherwise-idle pass, ~12×/hour, 24/7, with one embed request per worktree per pass (#577 has the full profile/log analysis).

Three changes, one seam:

1. **Event-scoped overlay refresh.** The event loop attributes each firing event to its checkout (`WorktreeEventHint`) and unions the contributions while the debounce is armed (`OverlayScope`); the accumulated scope rides the `PassRequest`. An event-scoped pass refreshes only the implicated worktrees. Unattributable events (backend rescan, worktree-registry changes) widen the pass to `All`, as do startup catch-up, the periodic sweep, gc-cadence passes, and the CLI/hook `maintenance` command — the missed-event backstop is unchanged.
2. **Per-worktree refresh basis.** After each successful, **complete** refresh the pass records `(base HEAD, linked HEAD)` as a `repo_meta` marker (`worktree_overlay_basis:<worktree_id>`, same pattern as `watch_shutdown_reconcile_pending`). A scoped pass skips an unlisted worktree only while that pair still matches — so a **base commit** (which moves the diff basis for every overlay at once) and a **linked commit** that arrives with no file event are both still refreshed, from two cheap head reads instead of a delta computation. Heads are read before the refresh and the basis is recorded after it, so races re-refresh rather than skip. A **partial** refresh (the gix status read failed midway) or a failed one *clears* any recorded basis — neither HEAD moves on a dirty edit, so a stale basis would otherwise stay skip-proof. A linked checkout's `rag-rat.toml` edit (which re-targets its branch overlay without moving HEAD) now fires and attributes to its checkout, like a `.gitignore` edit. gc prunes the marker with the dead worktree's overlay rows.
3. **No more per-pass embed probes.** An unchanged worktree checks for an embedding backlog only on `All` sweeps, via the SQL-only `pending_embedding_jobs_with_options` — no embedder acquisition, so an idle pass makes zero embed requests. A `Partial` inline drain heals within one `periodic_sweep_secs` instead of being re-probed on every pass; whether a backlogged reconcile can actually drain is decided inside the reconcile itself (`provision_remote=false`, unchanged).

## Measured

Two linked worktrees (110/130 files diverged), release binaries, steady state, watcher driven end-to-end over MCP; stage timings from the #502 pass summary:

| pass | before | after |
|---|---|---|
| base-only edit (both worktrees unchanged) — `overlays` stage | 123 ms | **4 ms** |
| idle periodic sweep — `overlays` stage | 85–164 ms | 94–99 ms (sweeps still refresh everything, by design) |
| worktree edit — refresh work | touched worktree + sweep of the sibling | touched worktree only |

The skip cost is two head reads + one kv lookup per worktree, so the base-only case scales with fleet size: the production machine that motivated this (#577) runs 8 linked worktrees and pays the sweep on every debounced pass during active hours. The embed-probe elimination (8 HTTP round-trips per idle pass there) is covered by unit tests; the bench env has no embedding model, so both runs show zero requests.

## Tests

- Scoped skip proven by behavior: a dirty edit in an unlisted, basis-unchanged worktree stays out of its overlay on a scoped pass and is picked up by the next `All` sweep.
- Basis invalidation: a linked commit and a base commit each force the refresh under an empty event scope.
- Event attribution: per-checkout `Roots`, registry/rescan → `AllWorktrees`, base-only events → empty scope; placement reporting stays non-short-circuiting across states.
- Scheduler: scope rides the request, coalesces while in flight, gc-cadence passes widen to `All`.
- `overlay_needs_embed` gate matrix (changed / sweep / backlog), including "scoped pass never probes".
- gc prunes a removed worktree's basis marker, keeps a live sibling's.

### DIFF
--- a/crates/rag-rat-cli/src/commands/index_ops.rs
+++ b/crates/rag-rat-cli/src/commands/index_ops.rs
@@ -432,7 +432,12 @@ fn run_maintenance_pass(
     // CHANGED overlay's embeddings are reconciled INLINE (while scoped to it) so worktree queries
     // aren't BM25-only for branch content. It restores the base scope afterward so the base
     // reconcile/gc/memory-validate below run unscoped.
-    rag_rat_core::watch::refresh_worktree_overlays(&mut db, config, budget.as_ref());
+    rag_rat_core::watch::refresh_worktree_overlays(
+        &mut db,
+        config,
+        budget.as_ref(),
+        &rag_rat_core::watch::OverlayScope::All,
+    );
     // The base reconcile gets whatever budget the overlays left; `None` → exhausted (or no cap left
     // at all), so skip it rather than start a fresh full-budget reconcile.
     let reconcile_report = match budget
--- a/crates/rag-rat-core/src/index/mod.rs
+++ b/crates/rag-rat-core/src/index/mod.rs
@@ -68,6 +68,7 @@ pub use query_api::{
     WalCheckpointReport,
 };
 pub(crate) use util::*;
+pub use worktree_overlay::WorktreeOverlayReport;
 
 #[cfg(test)]
 mod anchor_tests;
--- a/crates/rag-rat-core/src/index/query_api/ai_lifecycle.rs
+++ b/crates/rag-rat-core/src/index/query_api/ai_lifecycle.rs
@@ -79,6 +79,17 @@ impl IndexDatabase {
         ai::pending_embedding_jobs(self.storage.connection())
     }
 
+    /// [`Self::pending_embedding_jobs`] with the caller's candidate sizing — SQL-only, no embedder
+    /// acquisition (no probe request), so the watcher's `All`-sweep overlay backlog check is free
+    /// of network work on an idle pass (#577). Whether the backlog can actually drain is decided
+    /// inside the reconcile itself.
+    pub(crate) fn pending_embedding_jobs_with_options(
+        &self,
+        options: &ai::ReconcileOptions,
+    ) -> anyhow::Result<u64> {
+        ai::pending_embedding_jobs_with_options(self.storage.connection(), options)
+    }
+
     pub(crate) fn pending_embedding_jobs_with_available_incremental_embedder(
         &self,
         options: &ai::ReconcileOptions,
--- a/crates/rag-rat-core/src/index/query_api/gc.rs
+++ b/crates/rag-rat-core/src/index/query_api/gc.rs
@@ -160,6 +160,8 @@ impl IndexDatabase {
         // dead-GENERATION sweep already ran above, unconditionally.
         self.delete_staged_files_cascade(StagedSweep::DeadContext)?;
         conn.execute_batch("DELETE FROM temp.staged_file_ids;")?;
+        // A removed worktree's overlay refresh basis (#577) follows its overlay rows out.
+        self.prune_worktree_overlay_basis_outside(live_worktrees)?;
         // Since #248 `edge_oracle` is content-keyed with NO `edges_data` FK (it survives reindex,
         // the moniker model), so the file/edge prune above no longer cascades verdicts
         // away. Dangling verdicts are harmless for correctness — every read joins live
--- a/crates/rag-rat-core/src/index/schema_bootstrap_tests/worktree_overlay.rs
+++ b/crates/rag-rat-core/src/index/schema_bootstrap_tests/worktree_overlay.rs
@@ -188,7 +188,12 @@ fn refresh_worktree_overlays_reconciles_overlay_scope_embeddings() {
     let options = ai::ReconcileOptions { batch_size: Some(8), ..Default::default() };
     let budget = crate::watch::ReconcileBudget::new(options, std::time::Instant::now());
     // The pass refreshes the overlay AND reconciles its embeddings inline.
-    let changed = crate::watch::refresh_worktree_overlays(&mut db, &config, Some(&budget));
+    let changed = crate::watch::refresh_worktree_overlays(
+        &mut db,
+        &config,
+        Some(&budget),
+        &crate::watch::OverlayScope::All,
+    );
     assert!(changed, "the overlay changed (a new branch file was indexed)");
 
     // In the overlay scope, the new branch file's chunk must carry a Current embedding — not be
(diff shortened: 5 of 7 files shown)


skakri:
@codex review

Issue #606, opened 2026-07-10
phase C — C3: /3 acceptance wiring + /1→/2 adoption (#405)


skakri:
Wire the /3 acceptance predicate (§13: citation authority + registers + branch + freshness) into the memory write/read path, and the /1 → /2 adoption migration (§17) that re-authors the live local projection as a /3 genesis batch and retires the /1 stream. Depends on C1, C2. Spec: §13 / §17. Part of #405.

skakri:
**C3.4 (/1→/2 adoption + live-path switch) broken out into milestone #19:**
- #662 — C3.4a: local account bootstrap + account-op author seam
- #663 — C3.4b-i: /3 local-authoring + accepted→projection substrate
- #664 — C3.4b-ii: retarget the live authoring path + §17 adoption
- #665 — C3.4c: read-path confirmation + spec-conformance callout

Key finding from planning: local /3 authoring must use in-tx author seams per §16.2, NOT the self-transacting/quota-capped `content_ingest` (the remote-input path). Acceptance is correct-but-inert locally; load-bearing at phase D. C3.1–C3.3 (acceptance predicate + refold) are already merged.

skakri:
C3 is complete. C3.1–C3.3 (the `/3` acceptance predicate + refold wiring) merged earlier; the C3.4 breakout — #662 (local account bootstrap + account-op author seam), #663 (`/3` local-authoring + accepted→projection substrate), #664 (retarget the live authoring path + §17 adoption), #665 (read-path confirmation + spec-conformance callout) — is all closed.

Follow-ups that came out of this work are tracked separately and are not C3 scope: #691 (project accepted foreign `/3` into `repo_memories`, gated on transport) and #764 (chunk StreamKeyWrap fan-out for large rosters).

Remaining phase C work is #609 (C6: snapshots + coverage manifest + tombstone horizon).

PR #647, opened 2026-07-14
feat(account): add /3 candidate ancestry and branch selection

skakri:
## Summary

Second C3 slice, on top of the merged acceptance predicate (#645). It supplies the two inputs that predicate takes from the content DAG rather than the account log:

- **`ancestry`** — walks `prev_hash` back from a cut's watermark and reports `OnBranch` / `OffBranch`, or UNDECIDED *with its cause*: a withheld watermark parks as `unknown_cut_target`, a missing mid-chain link as `incomplete_cut_ancestry`. Neither ever flips a verdict (I11). A signed header pins only `prev_hash` NULLITY, never that `prev` is a valid contiguous parent — so the walk re-derives that, and refuses to follow a link that jumps chain coordinate or skips a seq slot.
- **`validate_cut_target`** — binds a cut's watermark to the exact `(stream, author_account, device, seq)` its register bounds (§11.3). A cut naming a foreign coordinate is structurally invalid and may neither condemn nor pin.
- **`select_accepted_branch`** — takes one contiguous chain per coordinate from the eligible candidates (§16.2). A register watermark pins the branch it names (promoting a currently-`forked` branch — what makes an off-branch condemnation enforceable); an unforced fork resolves to the lexicographically smaller `entry_hash`.

Mirrors `account::candidate` deliberately, so both logs enforce the same rules.

## The three states, and why two of them are not a partition

`accepted` and `forked` do **not** partition the eligible candidates, and that is the point:

| state | meaning |
|---|---|
| `accepted` | on the chain's selected branch |
| `forked` | reached its chain root through held entries and still lost a slot — a real equivocation loser, terminal unless a later watermark selects it |
| neither | stranded above a gap in the dense chain: it lost nothing, it is merely early |

Calling a stranded entry `forked` would discard valid work for arriving out of order, since the missing predecessor can still turn up and make it contiguous. The acceptance predicate parks it as `missing_predecessor` instead.

`eligible` is the caller's authority verdict rather than something this derives: a condemned entry must never compete for a slot, or an attacker mines a small-hash entry beyond a cut, wins the unforced tiebreak, and forks an honest sibling off the branch. Unlike the account fold this needs no fixpoint loop — content entries confer no authority on other content entries, so one pass suffices.

## Convergence detail worth a second look

One chain can be bounded by **two** registers — the author's roster cut and the owner's grant cut — so an equivocating device can leave them naming different watermarks at the *same* seq. Pins are therefore ordered by `(seq, watermark)`, a total order: `max_by_key` on `seq` alone hands a tie to whichever pin happens to come last, and since pins are assembled from unordered projection reads, two peers holding identical entries would derive different accepted branches.

## Validation

- `cargo +nightly fmt --all --check`; clippy with CI's exact invocations, on **1.96** (the toolchain CI actually runs — 1.97 no longer fires `type_complexity`, which is how the first round went red)
- `cargo nextest run -p rag-rat-core` — 2,380 passed, 0 failed
- Mutation-checked every load-bearing rule (pin promotion, the eligibility guard, the pin total-order, the gapped-vs-forked split, the `u64::MAX` contiguity check): each mutation fails exactly the test that claims to cover it, so none pass vacuously. The `u64::MAX` mutation panics with `attempt to add with overflow`, which is the bug it guards.
- Selection is asserted independent of row order and of pin order; a 2,000-deep chain selects without a stack overflow

Refs #606

### DIFF
--- a/crates/rag-rat-core/src/oplog/account/content/candidate.rs
+++ b/crates/rag-rat-core/src/oplog/account/content/candidate.rs
@@ -132,6 +132,10 @@ pub(super) struct BranchPin {
     pub(super) watermark: EntryHash,
 }
 
+/// Eligible entries indexed by the `(chain, prev_hash)` parent slot they extend; a chain root keys
+/// on `None`. Several entries under one key are an equivocation — the slot selection resolves them.
+type BranchChildren = HashMap<(ChainCoordinate, Option<EntryHash>), Vec<(u64, EntryHash)>>;
+
 /// The branch-selection verdict for one refold: the entries on each chain's accepted branch, and
 /// the equivocation losers (`forked` — terminal unless a later watermark selects them).
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


skakri:
@codex review

PR #670, opened 2026-07-15
fix(watch): surface silently-dropped filesystem watch placements

### LINKED ISSUE #658: fix(watch): surface silently-dropped inotify watch placements
## Problem

`watch_tree_pruned` (`crates/rag-rat-core/src/watch/placement.rs:343`) swallows watch-placement failures:

```rust
if watcher.watch(&d, RecursiveMode::NonRecursive).is_err() {
    continue;
}
```

On Linux, `watcher.watch()` returns `ENOSPC` when `fs.inotify.max_user_watches` is exhausted. The `continue` not only drops the watch on that directory — it also skips the `read_dir` below it, so the **entire subtree** silently loses coverage. The directory then falls back to the periodic sweep (bounded staleness), but there is **no signal** that the real-time path degraded. You cannot tell whether the watcher is actually watching or silently riding the 300s sweep.

## Scope

- Count (and log, coalesced) watch-placement failures in `place_watches` / `watch_tree_pruned` / `watch_created_dirs`, distinguishing "watch dropped" from "subtree skipped".
- Surface the count via a queryable status field (e.g. `index_status` / a watcher-health field), so degradation is observable without log scraping.
- No change to placement behavior; a failed watch still falls back to the sweep exactly as today.

## Acceptance criteria

- [ ] A watch-placement failure increments an observable counter and emits one coalesced log line (not one per directory).
- [ ] The dropped-watch count is visible via a status surface.
- [ ] No behavioral change to which directories get watched; the sweep fallback is unchanged.
- [ ] Cross-OS: the counter is meaningful on every platform (`notify` can fail to place a watch anywhere); Linux `ENOSPC` is the motivating case but the handling is not Linux-specific.

## Out of scope

- Adaptive/hot-cold watch retirement (measured non-bottleneck at current scale; deferred).
- Raising `max_user_watches` or any placement-strategy change.

Independent of the rest of the milestone — can land first.

skakri:
Fixes #658.

## Problem

A directory whose `notify::Watcher::watch` call fails — on Linux, `ENOSPC` once `fs.inotify.max_user_watches` is exhausted; on any OS, a transient failure — silently falls back to the periodic mtime sweep. Until now that degradation was invisible: graph/symbol queries could go stale between sweeps with no signal that real-time watching was partially down.

## Change

Count every watch-placement outcome through the single `place_watch` choke point into per-watcher `WatchPlacementCounters` (interior atomics, shared via `Arc` between the event-loop thread that records placements and the pass worker that reads them). Each pass persists the failure count into `repo_meta`, and `index_status` surfaces it as `watch_placement_failures`. A coalesced `warn!` fires once per new batch of drops (never per directory), correct across a restart via an instance-local last-warned mark.

Key properties:

- **Owned by the watcher instance, not a process global.** One repo config's placement outcomes can never be attributed to another's, and the counts don't outlive the watcher's shutdown. Hook/CLI passes own no watcher and record nothing (`None`) — they place no watches and must not touch the resident watcher's mark.
- **High-water mark, never lowered.** A healthy or freshly-restarted watcher sharing the DB can't erase a concurrently degraded watcher's record; the persist is an atomic max-upsert (single statement), so a concurrent same-repo writer can't regress it. It is a degradation *signal* (per-lifetime maximum), documented as such — not a live "currently degraded" gauge.
- **Absent directories are not counted.** A watch that failed because a configured target doesn't exist yet — a not-yet-created target dir, a branch-specific subdir, a worktree registry that appears later — is an expected miss already covered by the ancestor/bootstrap/created-dir watches, so it does not peg the mark or emit a spurious ENOSPC warning. Only a watch that failed on a directory that *exists* is a real drop.
- **Shutdown flush.** The post-pass linked-worktree sync places watches after the pass already persisted the counter, so the watcher flushes the final mark on the way out — otherwise a drop introduced there, on a periodic-sweep-disabled watcher with no further events, would never be recorded. Best-effort and bounded; a no-op (no DB open) when nothing ever failed.

A failed watch on a directory skips its whole unwalked subtree exactly as before — only the visibility changes.

## Tests

- Failed placements are counted and the subtree is still skipped (per-watcher instances, no cross-test global state).
- A watch failure on an *absent* directory is not counted; the same failing watcher on an existing dir is.
- The shutdown flush persists the mark into `index_status`.
- `index_status` round-trips the value through `repo_meta` with high-water semantics.

Full workspace suite green (`cargo nextest run --workspace --no-default-features`); all three CI clippy gates pass.

### DIFF
--- a/crates/rag-rat-core/src/index/meta.rs
+++ b/crates/rag-rat-core/src/index/meta.rs
@@ -1,9 +1,14 @@
 //! Index key-value meta (the `index_meta` table) and the content-revision digest.
 
 use super::*;
-use crate::config::ResolvedTarget;
+use crate::config::{Config, ResolvedTarget};
+use crate::storage::IndexConnection;
 
 const WATCH_SHUTDOWN_RECONCILE_PENDING_META: &str = "watch_shutdown_reconcile_pending";
+/// Watch-placement failure HIGH-WATER MARK the resident watcher has seen (see `watch::placement`).
+/// Persisted per pass, never lowered, so `index_status` can surface silent inotify degradation
+/// without one watcher process masking another's (see `record_watch_placement_failures`).
+const WATCH_PLACEMENT_FAILURES_META: &str = "watch_placement_failures";
 const BASE_SCOPE_DISCOVERED_META: &str = "files_base_scope_discovered";
 
 impl IndexDatabase {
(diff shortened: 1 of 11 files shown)


skakri:
@codex review

PR #696, opened 2026-07-16
perf(symbol): trigram-index name_strings so bare-name fuzzy lookup stops full-scanning (#685)

### LINKED ISSUE #685: symbol_lookup bare-name fuzzy full-scans name_strings (leading-wildcard LIKE) — index with FTS5 trigram, preserve substring semantics
## Summary

`symbol_lookup` bare-name fuzzy resolution full-**SCAN**s the entire `name_strings` pool (~116k rows on a large index) via a leading-wildcard `LIKE '%name%'`, costing ~40 ms per call. This is the substring-match arm only; exact whole-word lookup is already fast and is not affected.

## Measurement

On a real index (`name_strings` ≈ 115,604 rows), warm cache, sqlite 3.46.1:

| Path | mean | plan |
|---|---|---|
| `symbol_lookup` fuzzy bare-name (`lookup_name`) | ~39.6 ms | `LIST SUBQUERY → SCAN name_strings` |
| `symbol_lookup` exact qualified-name (`lookup_symbol_path`) | ~0.02 ms | UNIQUE `value` index + `idx_symbols_qualified_name_id` |

## Root cause

`lookup_name` (`crates/rag-rat-core/src/query/symbol.rs`) matches qualified names with:

```sql
WHERE (symbols.name = ?1
       OR symbols.qualified_name_id IN (SELECT id FROM name_strings WHERE value LIKE ?2))   -- ?2 = '%name%'
```

The first arm (`symbols.name = ?1`) is an **exact whole-word** match on the indexed `symbols.name` — fast, and it already covers the common "search by the exact identifier" case. The second arm is the substring/fuzzy match: the leading `%` makes the `LIKE` non-sargable, so the subquery scans the whole `name_strings` pool on every bare-name lookup. Cost is essentially independent of the chosen name and grows linearly with the string pool, so it gets worse as the consolidated store grows.

## Constraint (why this needs its own decision)

The fuzzy arm's value is **arbitrary infix substring** matching (`str` finds `as_str`, `to_string`, `str_len`). A naive fix that swaps it for a porter/word-tokenized FTS would silently change the semantics to **whole-token / prefix** matching and lose infix substrings — a behavior regression for agents that rely on partial-name search. The fix must preserve substring semantics.

## Proposed direction

Use an **FTS5 `trigram`** index over `name_strings.value` (or the symbol name set), which supports arbitrary-substring `LIKE` / `MATCH` (min 3 chars) with an index instead of a full scan. Keep the exact `symbols.name = ?1` arm as-is (it's already indexed and is the whole-word path). Short (<3 char) queries fall back to the current scan or the exact arm. Bench the before/after on a large index to confirm the scan is gone and substring results are unchanged.

## Acceptance criteria

- `EXPLAIN QUERY PLAN` for the bare-name fuzzy lookup no longer shows `SCAN name_strings`.
- Substring matches returned are unchanged (a corpus test: a set of `%frag%` queries returns the same symbol set as today).
- Exact whole-word lookup is unchanged and still uses `idx_symbols_name`.

Context: split out of #682 (the graph-traversal full-scan fix), which deliberately left this to a separate decision.

skakri:
## Problem

`symbol_lookup`'s bare-name **fuzzy** arm matches `name_strings.value LIKE '%name%'` — a leading-wildcard LIKE that cannot use the UNIQUE `value` index, so it full-**SCAN**s the ~116k-row interned-name pool on every bare-name lookup (~40 ms measured). The exact whole-word arm (`symbols.name = ?1`) is separate, already indexed, and unaffected.

## Fix

**V72 migration** adds `name_strings_trgm`, an FTS5 **`trigram`** index over `name_strings(value)`:
- External content (`content='name_strings'`, `content_rowid='id'`) — the FTS stores only the index and reads values from the base pool.
- Kept in sync by AFTER INSERT/DELETE/UPDATE triggers (the documented FTS5 external-content pattern); each interned name is indexed as the pool fills.
- One-time backfill of existing names. DROP-and-recreate applier (like `ensure_edges_view`) → fully idempotent and torn-replay safe.

`lookup_name`'s fuzzy subquery changes from `SELECT id FROM name_strings WHERE value LIKE ?2` to `SELECT rowid FROM name_strings_trgm WHERE value LIKE ?2` — the **same `LIKE`**, just resolved through the trigram index. The trigram tokenizer accelerates arbitrary-substring LIKE (≥3 literal chars; shorter patterns still resolve correctly), preserving the exact match set — so substring search (e.g. `str` → `as_str`, `to_string`) is unchanged, not narrowed to whole-token.

## Result

Measured on a real ~116k-name index (identical result rows in every case):

| bare name | old | new | speedup |
|---|---|---|---|
| `reconcile` | 42.3 ms | 5.9 ms | ~7× |
| `embed` | 45.4 ms | 8.8 ms | ~5× |
| `new` | 39.1 ms | 7.8 ms | ~5× |
| `as_str` | 36.6 ms | 20.1 ms | ~2× |
| `str` (9.5k matches) | 51.1 ms | 31.0 ms | ~2× |

(Common substrings retain a floor cost from the outer symbol join + rank `ORDER BY` over the larger match set, which the index can't reduce.) Trigram index overhead ≈ 19 MB for 116k names (~1% of a 1.6 GB store).

## Tests

- `migration_072_is_the_tip_and_trigram_indexes_name_strings` — asserts the FTS table + 3 triggers exist, the AFTER INSERT trigger indexes newly interned names, the trigram `LIKE` returns the **same ids as the base pool `LIKE`**, backfill re-indexes every existing name exactly once, replay is idempotent, and forward-migrate records V72.
- Existing `symbol_lookup` behavior tests pass unchanged (they exercise `lookup_name` end-to-end through the new trigram path).

Fixes #685

### DIFF
--- a/crates/rag-rat-core/src/index/schema/migrations.rs
+++ b/crates/rag-rat-core/src/index/schema/migrations.rs
@@ -784,6 +784,52 @@ pub(crate) fn apply_edge_target_qname_index(conn: &Connection) -> rusqlite::Resu
     )
 }
 
+/// V072 (#685): a trigram FTS index over the interned-name pool. `symbol_lookup`'s bare-name fuzzy
+/// arm matches `name_strings.value LIKE '%name%'` — a LEADING-wildcard LIKE that cannot use the
+/// UNIQUE value index, so it full-scans the whole pool on every bare-name lookup. An FTS5 `trigram`
+/// index accelerates that exact `LIKE` (identical substring-match set for patterns ≥3 chars;
+/// shorter patterns still resolve correctly via a scan). External content
+/// (`content='name_strings'`, `content_rowid='id'`) means the FTS stores only the index and reads
+/// values from name_strings; the insert/delete/update triggers keep it in sync as names are
+/// interned.
+///
+/// DROP-and-recreate (like `ensure_edges_view`) makes it fully idempotent and torn-replay safe: the
+/// backfill always reflects the current pool. It runs once per DB (recorded in schema_version); on
+/// a fresh DB name_strings is still empty here, so the backfill inserts nothing and the AFTER
+/// INSERT trigger indexes each interned name as indexing fills the pool.
+pub(crate) fn apply_name_strings_trigram_index(conn: &Connection) -> rusqlite::Result<()> {
+    conn.execute_batch(
+        "
+        DROP TRIGGER IF EXISTS name_strings_trgm_ai;
+        DROP TRIGGER IF EXISTS name_strings_trgm_ad;
+        DROP TRIGGER IF EXISTS name_strings_trgm_au;
+        DROP TABLE IF EXISTS name_strings_trgm;
+
+        CREATE VIRTUAL TABLE name_strings_trgm USING fts5(
+            value,
+            content='name_strings',
+            content_rowid='id',
+            tokenize='trigram'
+        );
+
+        CREATE TRIGGER name_strings_trgm_ai AFTER INSERT ON name_strings BEGIN
+            INSERT INTO name_strings_trgm(rowid, value) VALUES (new.id, new.value);
+        END;
+        CREATE TRIGGER name_strings_trgm_ad AFTER DELETE ON name_strings BEGIN
+            INSERT INTO name_strings_trgm(name_strings_trgm, rowid, value)
+                VALUES ('delete', old.id, old.value);
+        END;
+        CREATE TRIGGER name_strings_trgm_au AFTER UPDATE ON name_strings BEGIN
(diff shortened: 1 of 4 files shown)


skakri:
Closing unmerged — the trigram approach caused a hard bencher regression (index_time/full_rebuild_cargo +78%, over threshold) from per-insert FTS maintenance on the hot indexing path, and the correct regression-free fix is disproportionate FTS-lifecycle surgery for a secondary read path. Analysis captured on #685; keeping that issue open as won't-do-for-now.

Issue #607, opened 2026-07-10
phase C — C4: account.secrets + content keys + X25519 wrap (#405)


skakri:
The account.secrets log (§8) + per-stream XChaCha20-Poly1305 content keys + X25519 sealed-box wrap per granted device (§15), non-circular `key_id = HKDF(content_key, ...)`, key-wrap fold + lazy rotation on removal. Adds `chacha20poly1305` + `x25519-dalek`. Depends on C1, C3. Spec: §8 / §15. Part of #405.

skakri:
C4 complete — all sub-slices merged: content-key primitives (#709), secrets-chain cut registers (#737), owner-gated secrets acceptance + StreamKeyWrap (#751), minting + wrap authoring (#754), sealing-key selection + key_id adoption cross-check (#760), and lazy content-key rotation on device removal (#762). Deferred follow-ups tracked in #763 (new-device catch-up) and #764 (wrap fan-out chunking). The consuming seal/decrypt path is C5 (#608).

PR #709, opened 2026-07-16
feat(oplog): content-key crypto primitives — content keys, key_id, X25519 sealed-box wrap (C4.1)

skakri:
First slice of the content-key-management phase (#607). The crypto core only — the `account.secrets` log, `key_wrap` authoring, and the actual payload sealing land in later slices (C4.2/C4.3 and C5). Design was adversarially crypto-reviewed before build; the review's three blockers are why `key_id` is 32 bytes, why the ephemeral key is validated + the shared secret asserted contributory, and why `key_wrap` authoring authority is pinned (enforced in C4.3).

## What lands (`oplog::account::keywrap`)
- **`ContentKey`** — a stream's XChaCha20-Poly1305 key (`Zeroizing`, OS-CSPRNG `generate` + seed-deterministic `from_seed`).
- **`KeyId`** = `HKDF-SHA256(content_key, info="rag-rat/content-key-id/1")`, 32 bytes (the frozen `/3` header pins `key_id` at exactly 32). One-way/non-circular — safe in a clear header, and it's what lets a device cross-check its peers' key at an epoch before adopting a key for sealing.
- **`WrapContext { account_id, stream_id, key_epoch, recipient_pub }`** — one canonical fixed-width encoding bound as the AEAD AAD, so a wrap can't be transplanted across account, stream, epoch, or device.
- **`SealedKeyWrap { ephemeral_pubkey, ciphertext[48] }`** + canonical CBOR (domain `rag-rat/key-wrap/1`).
- **`seal_content_key` / `unwrap_content_key`** — hand-rolled HPKE base mode over DHKEM(X25519, HKDF-SHA256) + XChaCha20-Poly1305, mapped field-by-field to RFC 9180 in the module docs. Ephemeral sender key; `HKDF-Extract(salt="", ikm=dh)` then Expand binding `epk || recipient_pub`; wrap key **and** a derived nonce from one PRK (no nonce on the wire).
- **`DeviceX25519Secret::diffie_hellman`** (in `device.rs`) — the ECDH the C1 design reserved for C4; returns the raw `SharedSecret` so the caller owns the contributory policy.

## Security hardening (from the crypto review + Codex)
- Ephemeral pubkey validated through the small-order/identity blocklist **before any DH**; shared secret asserted `was_contributory()` on **both** seal and unwrap — closes the world-computable all-zero-shared-secret substitution.
- Recipient key **canonicalized** before the KDF (X25519 ignores bit 255; seal/unwrap would otherwise diverge for a non-canonical encoding).
- `WrapContext` must name the **actual** recipient (rejects an internally-inconsistent wrap).
- Recovered key `decrypt_in_place` into a `Zeroizing` buffer, and chacha20poly1305's `zeroize` feature enabled so the cipher's own key copy is scrubbed.

## Deps & supply chain
Adds `chacha20poly1305 0.11` (`default-features=false`, `alloc`+`zeroize`) + `hkdf 0.13`, riding the existing `sha2 0.11` / `digest 0.11` line — **no second hash stack**. `cargo deny check`: advisories/bans/licenses/sources all ok.

## Tests
Golden vectors freeze `key_id`, the `WrapContext` bytes, and a full sealed wrap. Adversarial: wrong-recipient, per-field transplant, tamper (epk/ct/tag), small-order epk rejected at validation, non-canonical recipient roundtrips, context/recipient mismatch rejected — the two review-driven fixes are mutation-verified. fmt + both CI clippy invocations + `cargo nextest run -p rag-rat-core` (2609, minus a known dev-box-cage flake in an unrelated `embed_loop` test that passes on CI) all green. Codex-reviewed clean.

## Note (spec owner)
The crypto review flagged that `key_wrap` authoring authority is absent from the sync spec's authorization table. I've pinned it **owner-only** in the C4 plan (enforced in C4.2/C4.3, not this slice); you may want to add the row to the spec doc.

Refs #607

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "aead"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1973cfbc1a2daf9cf550e74e1f088c28e7f7d8c1e1418fb6c9dc5184b7e84c99"
+dependencies = [
+ "crypto-common 0.2.2",
+ "inout",
+]
+
+[[package]]
 name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -320,8 +330,23 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
 dependencies = [
  "cfg-if",
+ "cipher",
  "cpufeatures 0.3.0",
  "rand_core 0.10.1",
+ "zeroize",
+]
+
+[[package]]
+name = "chacha20poly1305"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b89e1c441e926b9c82a8d023f6e1b7ae0adcfaa7d621814e4d60789bac751cb"
+dependencies = [
+ "aead",
+ "chacha20",
+ "cipher",
+ "poly1305",
+ "zeroize",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


skakri:
@codex review

PR #723, opened 2026-07-17
test(mcp): harden the listener-takeover test with explicit readiness signaling (fixes #537)

### LINKED ISSUE #537: flaky(mcp): second_listener_takes_over_when_winner_dies flakes under llvm-cov coverage
Observed on PR #533 (an unrelated edge-extraction change): the coverage CI job failed on `rag_rat_mcp::claude_hook::listener_tests::second_listener_takes_over_when_winner_dies`, then passed on rerun. Locally the test passes but with wide timing variance (5 runs: 6s, 6s, 6s, 6s, and one 42s), which points at a process/listener-failover race that the slower llvm-cov instrumentation exposes. Not a product bug and not related to the PR that surfaced it. Fix: make the failover handoff deterministic (poll/await the takeover with a bounded retry instead of a fixed timing assumption), or mark the test to tolerate the instrumented-run slowness.

Nitjsefnie:
Fixes #537 — same family as the just-merged #721: replace timing-window assumptions with explicit signaling.

## What

`second_listener_takes_over_when_winner_dies` no longer polls the socket with a fixed 100 ms sleep hoping the sequence lands — it waits on explicit `tokio::sync::watch` readiness events: winner bound, loser waiting, loser rebound. The events come from a `ListenerHooks` seam that is **entirely `#[cfg(test)]`-gated**: `spawn_listener` remains the sole non-test entry point, the shared listener body lives once in a `spawn_listener_task!` macro whose hooks argument compiles away in non-test builds, and the takeover/election logic itself is untouched. Verified: a plain `cargo build -p rag-rat-mcp --no-default-features` is clean with zero seam surface.

**Diff-size honesty:** the stat reads +148/−90, but nearly all of it is the existing listener body re-wrapped by the macro (indentation churn) — the alternative, duplicating the body into gated/ungated variants, seemed strictly worse. Happy to restructure if you'd rather see a different gating shape.

**Repro honesty (same story as #721):** the flake did not fire here — 20/20 pinned runs (`taskset -c 0`) passed on unmodified main both before and after. The value is removing the timing dependence entirely. What proves the test still bites: breaking the takeover (loser exits instead of retrying the election) fails it deterministically — `loser must wait, not exit` — twice verified (implementer + an independent re-run before opening).

## Gates

fmt (nightly), clippy (default features — clippy doesn't link, so the ONNX chain isn't in play), and `cargo test --workspace --no-default-features --locked` all green; crate suite 58/58.

---
🤖 This PR was written with AI assistance (implementation: Kimi K2.7 Code; verification and review: Claude), directed by a human-supervised workflow.

### DIFF
--- a/crates/rag-rat-mcp/src/agent_hook.rs
+++ b/crates/rag-rat-mcp/src/agent_hook.rs
@@ -64,6 +64,41 @@ mod listener {
         locks::hook_socket_lock_path_for(config)
     }
 
+    /// Test-only instrumentation hooks for the listener election loop. Tests can wait on the
+    /// receivers to observe readiness instead of polling or sleeping through scheduling races.
+    #[derive(Clone)]
+    #[cfg(test)]
+    pub struct ListenerHooks {
+        /// Set to true when this listener is about to sleep because the election lock is held by
+        /// another process.
+        pub waiting: tokio::sync::watch::Receiver<bool>,
+        waiting_tx: tokio::sync::watch::Sender<bool>,
+        /// Set to true when this listener has won the election and bound the socket, just before
+        /// entering the accept loop.
+        pub bound: tokio::sync::watch::Receiver<bool>,
+        bound_tx: tokio::sync::watch::Sender<bool>,
+    }
+
+    #[cfg(test)]
+    impl Default for ListenerHooks {
+        fn default() -> Self {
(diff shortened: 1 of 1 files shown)


skakri:
Thanks! 🚀

Issue #78, opened 2026-06-12
Benchmark honesty: report hash-embedding bytes separately (model="none") — structure+FTS+graph+git vs +vectors


skakri:
Surfaced from the ~10 GB kernel-index DB analysis.

The kernel bench uses the **hash embedder** (`--no-default-features`). The resulting **~1.6M deterministic baseline vectors are plumbing-validation, not retrieval value** — nobody does vector recall over the kernel with a hash embedder. If they're f32 at a few hundred dims, that's potentially **2+ GB of the ~10**.

## Proposal
- For the **published kernel number**, `model = "none"` is the honest config.
- Restructure the size reporting into the demonstration we actually want:
  > **kernel, full structure + FTS + graph + git: X GB; add vectors: +Y GB.**
- The degradability claim ("base install ships zero AI; add what you use") becomes a **measured number** instead of an assertion.

## Implementation
- `tools/bench-kernel.sh`: headline DB size = structure-only index (no embedding reconcile); optionally a second pass *with* embeddings to report the `+Y` delta.
- Emit both as BMF measures and put the split in `docs/benchmarks.md` (which currently has peak RSS + throughput but no DB size).

skakri:
Done — shipped in #724. `tools/bench-kernel.sh` sets `model="none"` for the headline config and measures the structure-only DB size (`db_size_bytes()` after `wal_checkpoint(TRUNCATE)`); the opt-in `RAG_RAT_KERNEL_VECTORS=1` pass reports `db_size_with_vectors` / `db_size_vectors_delta`. `docs/benchmarks.md` documents the split as a table (structure vs +vectors). Docs condensed in #728. Closing.

PR #724, opened 2026-07-17
bench: report the kernel DB size as a split — structure+FTS+graph+git vs +vectors (#78)

Nitjsefnie:
## Benchmark honesty: report the kernel DB size as a split (#78)

"Indexing the kernel costs N GB" is only honest if it says what is in the N. This reports it as `structure + FTS + graph + git: X; add vectors: +Y`, which turns the degradability claim ("base install ships zero AI; add what you use") into a measured number — `db_size_vectors_delta` is exactly the disk you avoid by not installing a model.

- `model = "none"` stated explicitly in the headline config rather than left to the default.
- Opt-in `RAG_RAT_KERNEL_VECTORS=1` second pass (installs the hash embedder, reconciles, re-measures). Off by default; exposed as a `vectors` dispatch input on `bench-release.yml` so +Y is producible from CI.
- BMF: `db_size` keeps its name and meaning (structure-only) so the Bencher series keeps its history; adds `db_size_with_vectors` + `db_size_vectors_delta`, **omitted** (not zeroed) when the pass does not run.
- Each size is taken after a `wal_checkpoint(TRUNCATE)`, so it is the durable on-disk footprint rather than a pre-checkpoint undercount.

## One correction to the issue premise — please read this before merging

The issue (mine) says the bench produces "~1.6M deterministic baseline vectors… potentially 2+ GB of the ~10". **It produces zero vectors.** `index --full` never runs an embedding reconcile:

- CLI `index` stops at `rebuild_with_progress` (`crates/rag-rat-cli/src/commands/index_ops.rs:89`); `reconcile` is a separate command
- neither it nor `crates/rag-rat-core/src/index/rebuild.rs` reaches an embedder — a full rebuild in fact *deletes* from `chunk_embeddings`
- `reconcile` refuses to embed under a never-installed model, and the bench never installs one

Verified empirically (indexed rag-rat itself twice → `chunk_embeddings = 0`) and across **every tag v0.4.0 → HEAD**. So past `db_size` figures were **already** structure-only: the X half needed *stating*, not changing, and only the +Y half was genuinely missing. The PR is correspondingly smaller than the issue implies.

## A pre-existing docs bug this surfaced

Because that pass never runs, `docs/benchmarks.md` attributed the post-COMMIT RSS plateau to "the embedding reconcile… embeddings are actually computed" — **a mechanism that does not exist**. I marked the cause **open** rather than inventing a replacement.

Corroborating datum: a **56-file** index peaked at 3429 MiB and a **673-file** index at 3495 MiB — near-identical across 12× the files, so the peak is not chunk-driven. The kernel full-history `git log --numstat` walk is my suspect, but I did **not** verify it and have not claimed it. Happy to split this hunk into its own PR or file it as an issue if you would rather keep this one narrow.

## Measurement honesty — what was NOT run

**I could not run the kernel at scale here, so no kernel X/Y GB number is quoted anywhere** — not in the docs, not in this description. The docs say explicitly that kernel-scale +Y is unmeasured and lands from your next `bench-release` dispatch.

What I did run: the **real** `bench-kernel.sh` against the **real** pinned kernel commit (`028ef9c…`), scope-bounded with the script own `RAG_RAT_KERNEL_SUBDIRS=lib` (673 files):

| | Bytes | |
|---|---|---|
| structure + FTS + graph + git | 64,815,104 | 62 MiB |
| add vectors (13,793 chunks → 10,151 embedded) | +13,271,040 | +13 MiB |
| total | 78,086,144 | 74 MiB |

~17% embedding tier on that subset — **do not extrapolate that ratio to the whole tree.** Note the vectors are int8 (388 B/row), so the issue "f32 at a few hundred dims → 2+ GB" sizing was off in two ways.

## Verification

There is no test harness for `bench-kernel.sh` anywhere in the repo, so rather than invent one I mutation-tested the real script: moving the structure measurement *after* the vectors pass drove `db_size_vectors_delta` **1,904,640 → 0** and silently made the headline the with-vectors size — **while still exiting 0**. That silent-wrong-number failure is exactly why the ordering is load-bearing, and it is commented as such. Restoring reproduced the baseline byte-identically. Both `RAG_RAT_KERNEL_VECTORS` branches were exercised (off → measures omitted + "NOT MEASURED"; on → all three).

Gates: `cargo build --no-default-features` → 0 · `cargo test --workspace --no-default-features` → 0 (**2,544 passed**) · `bash -n` → 0 · `shellcheck` → 1 on **one pre-existing SC2086 that is byte-identical on `upstream/main`** (I diffed the finding sets to confirm it is not mine; the word-splitting is intentional). Zero Rust files changed.

## Scope note

Beyond the two files the issue names, I touched `docs/bencher.md` (its measure list said "eight measures" and already omitted `db_size`) and `.github/workflows/bench-release.yml` (the input is what makes the docs "lands from the next dispatch" true). Both are minimal glue — say the word and I will drop either.

One environment caveat that is **not** in the diff: this box python links sqlite 3.34.1, which cannot open the `STRICT` schema (needs ≥3.37), so I injected `pysqlite3` via a `sitecustomize.py` outside the repo. Your `debian:trixie-slim` bench container is new enough that CI needs nothing — but the script does silently require sqlite ≥3.37 in its python, which is undocumented. Worth a line in the docs if you agree.

---
*Implemented with AI assistance (Claude Opus 4.8); the zero-vectors finding and the pre-existing-shellcheck claim were independently re-verified by me against the source before submitting.*

### DIFF
--- a/.github/workflows/bench-release.yml
+++ b/.github/workflows/bench-release.yml
@@ -20,6 +20,10 @@ on:
         description: 'Full-rebuild wave size (RAG_RAT_INDEX_WAVE) — the memory/speed knob (default 2000)'
         required: false
         default: '2000'
+      vectors:
+        description: 'Set to 1 to add the +vectors pass (RAG_RAT_KERNEL_VECTORS) — reports db_size_with_vectors / db_size_vectors_delta on top of the structure-only headline (#78)'
+        required: false
+        default: ''
       mem_trace:
         description: 'Set to 1 to emit per-phase MEMTRACE (RSS + sqlite_memory) to the log for peak diagnosis'
         required: false
@@ -48,8 +52,8 @@ env:
 
 jobs:
   kernel:
-    # Self-hosted runner (62 GB RAM): the whole-kernel index peaks at ~5.5 GiB RSS (the embedding
-    # reconcile; see docs/benchmarks.md) plus the OS page cache over a ~10 GB index DB — runnable on
+    # Self-hosted runner (62 GB RAM): the whole-kernel index peaks at ~5.5 GiB RSS (a post-COMMIT
+    # plateau; see docs/benchmarks.md) plus the OS page cache over a multi-GB index DB — runnable on
     # a hosted box but with a noisy, contended wall-clock, which defeats the point of a headline
     # number. Only the release / workflow_dispatch triggers land here (never PRs), so a public-repo
     # self-hosted runner is safe. See docs/bencher.md.
@@ -68,6 +72,7 @@ jobs:
         env:
           RAG_RAT_KERNEL_SUBDIRS: ${{ github.event.inputs.kernel_subdirs || '.' }}
           RAG_RAT_INDEX_WAVE: ${{ github.event.inputs.wave_size || '2000' }}
+          RAG_RAT_KERNEL_VECTORS: ${{ github.event.inputs.vectors }}
           RAG_RAT_MEM_TRACE: ${{ github.event.inputs.mem_trace }}
           RAG_RAT_SQLITE_SOFT_HEAP_LIMIT_MB: ${{ github.event.inputs.soft_heap_limit_mb }}
           RAG_RAT_WAL_AUTOCHECKPOINT: ${{ github.event.inputs.wal_autocheckpoint }}
@@ -83,6 +88,7 @@ jobs:
             -v rag-rat-target:/repo/target \
             -e RAG_RAT_KERNEL_SUBDIRS="$RAG_RAT_KERNEL_SUBDIRS" \
             -e RAG_RAT_INDEX_WAVE="$RAG_RAT_INDEX_WAVE" \
+            -e RAG_RAT_KERNEL_VECTORS="$RAG_RAT_KERNEL_VECTORS" \
             -e RAG_RAT_MEM_TRACE="$RAG_RAT_MEM_TRACE" \
             -e RAG_RAT_SQLITE_SOFT_HEAP_LIMIT_MB="$RAG_RAT_SQLITE_SOFT_HEAP_LIMIT_MB" \
             -e RAG_RAT_WAL_AUTOCHECKPOINT="$RAG_RAT_WAL_AUTOCHECKPOINT" \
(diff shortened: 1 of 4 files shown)


skakri:
I'll dispatch bench-release with vectors=1 so the kernel-scale +Y actually lands, the docs promise the number comes from next dispatch.

skakri:
The benchmark is dispatched: https://github.com/cq27-dev/rag-rat/actions/runs/29587742391

skakri:
Benchmark timed out.

Reconcile throughput degrades ~15x somewhere between a 10k-chunk index and a 1.26M-chunk one. I'm reproducing kernel-scale locally right now – the timed batch-size windows (64 vs 2048) plus a `perf` sample.

Nitjsefnie:
Two things, neither of which should hold up #724:

**This PR is independent of the +Y run.** The headline `db_size` it reports is structure-only and already correct; `db_size_with_vectors` / `db_size_vectors_delta` are the opt-in pass, and the docs deliberately say the kernel-scale +Y is *unmeasured, lands from a later dispatch* rather than quoting a number. So a timing-out `vectors=1` bench doesn't falsify anything here — if anything it's the case for the split: a single "N GB" figure would have baked in a pass that doesn't even complete at kernel scale. #724 can merge on its own; the +Y cell just stays empty until reconcile throughput is sorted.

**One datapoint for your bracket, if useful.** The `RAG_RAT_KERNEL_SUBDIRS=lib` run that produced the numbers in this PR did complete its +vectors pass cleanly: 673 files → **13,793 chunks, 10,151 embedded**, +13 MiB delta, no timeout. So ~14k chunks is still in the healthy regime — the ~15x degradation is somewhere north of that and south of 1.26M, consistent with a batch-window/allocation cliff rather than a linear slope. The int8-vector path (#112) is the other variable between those two scales.

I won't dig into the reconcile perf from here since you're already on it with `perf` + the batch-size windows — that'll localize it far better than I could from the outside. Happy to rework anything in this PR if the split's shape should change once you find the cause.

(Disclosure: AI-assisted.)

skakri:
You're correct, merging. I'll update benchmark numbers in a follow-up.

PR #732, opened 2026-07-17
fix(tests): route test scratch through a shared self-healing helper (fixes #726)

### LINKED ISSUE #726: test harness leaks /tmp scratch dirs on panic/kill — 16k dirs (~14GB tmpfs) accumulated
The schema/scope/watch test fixtures create per-test dirs under /tmp (`rag-rat-schema-test-*`, `ragrat-wtscope-*`, `rag-rat-init-yes-*`, …) cleaned by Drop impls. A panicking test, a killed nextest run, or a SIGKILL'd process strands them — on the dev box this accumulated 16,388 directories (~14 GB of tmpfs, 89% of /tmp) before a manual sweep.

Durable fix candidates:
- put test scratch under `target/tmp/` (per-checkout, reclaimed by `cargo clean` / the build-gc script) instead of the shared /tmp;
- or a startup sweep in the shared test fixture module: delete same-prefix dirs older than N hours before creating a new one (self-healing, no external cron);
- or route through `tempfile::TempDir` with the same startup sweep (Drop alone cannot cover kill -9).

Interim: `find /tmp -maxdepth 1 -name 'rag-rat-*' -o -name 'ragrat-*' -mmin +60 -exec rm -rf {} +`.

Nitjsefnie:
Fixes #726 — with one deliberate divergence from the issue's first candidate, argued with a reproduction.

## Why NOT `target/tmp/`

Routing scratch under the working tree is unsound for this codebase: `repo_discover` → `gix::discover` searches **upward**, so a fixture root without its own `.git` inside the checkout resolves the **outer rag-rat repo's** identity (the doc-comment on `discover_repo` even names this limitation). Tried empirically: with scratch pointed inside the tree, schema tests slow ~100× and fail with *"this database holds multiple repos; a bare open cannot choose one"* — the DB picks up the outer repo (implementer's first run hit this across ~295 tests; the mechanism was independently re-demonstrated during verification). Many fixtures `git init` throwaway repos and require no ancestor `.git`, so the "reclaimed by cargo clean" property is unreachable here.

## What landed instead — the issue's other two candidates, combined

`rag_rat_base::test_scratch`: all routed scratch lives under one **dedicated namespace** (`$TMPDIR/rag-rat-test-scratch/`) — so manual cleanup is one `rm -rf` and the sweep's blast radius is only our own dirs — with a **once-per-process startup sweep** deleting namespace dirs older than 2h (self-healing after SIGKILL/panic; the age bound means a live parallel worker's fresh dirs are never touched; all fs calls tolerate concurrent-sweep ENOENT). Drop-based cleanup stays as the fast path.

Routed through it: the three fixture families the issue names — the CLI integration `unique_dir` (`rag-rat-*`), `git_context`'s `ragrat-wtscope-*`, and schema-bootstrap's `rag-rat-schema-test-*`. **Scope note:** ~30 further inline `temp_dir()` sites (config tests, oracle/db/mcp) still write to bare /tmp — the helper is in place for incremental adoption; happy to do the full mechanical sweep in this PR or a follow-up, your call.

## Verification (adversarially re-run in a separate pass, not self-reported)

- `cargo nextest run --workspace --no-default-features --locked --profile ci`: **3027 passed / 0 failed** (baseline 3025 — the +2 are the new tests).
- Mutations both bite on independent re-runs: neutered sweep → `sweep_removes_stale_dirs_but_spares_fresh_ones` fails; creation pointed at bare `temp_dir()` → `scratch_dirs_live_under_the_dedicated_namespace` fails.
- `cargo +nightly fmt --all --check` clean (note: your `rustfmt.toml` uses nightly-only options, so *stable* rustfmt shows a spurious whole-repo diff on main and branch alike).
- clippy `--no-default-features -D warnings`: zero new findings; the one error present (`collapsible_match`, `oplog/account/storage.rs:1395`) fires identically on clean main — pre-existing, untouched here.

---
Generated by Claude Fable 5 (brief, review), Claude Opus 4.8 (1M context) (implementation, adversarial verification)

### DIFF
--- a/crates/rag-rat-base/src/test_scratch.rs
+++ b/crates/rag-rat-base/src/test_scratch.rs
@@ -16,8 +16,9 @@
 //!
 //! So this module fixes the leak the other way the issue proposed — a self-healing sweep:
 //!   1. all scratch lives under ONE dedicated, namespaced sub-root of the system temp
-//!      ([`scratch_root`], `<temp>/rag-rat-test-scratch-<euid>`), so a stranded dir is trivially
-//!      found and the sweep below can never touch anything but this helper's own dirs;
+//!      ([`scratch_root`], `<temp>/rag-rat-test-scratch-<tag>` — the euid on Unix, the sanitized
+//!      username on Windows), so a stranded dir is trivially found and the sweep below can never
+//!      touch anything but this helper's own dirs;
 //!   2. the FIRST scratch request in each process sweeps sibling scratch dirs older than
 //!      [`SWEEP_MAX_AGE`] out of that root — self-healing after a `kill -9`, no external cron.
 //!
@@ -34,11 +35,14 @@
 //! [`SWEEP_MAX_AGE`] — attacker-directed deletion of arbitrary directories (#726 review, skakri).
 //!
 //! Two defences, both below:
-//!   * **per-euid namespace** — the root carries the effective uid (`…-<euid>`), so two users
-//!     sharing the temp get structurally distinct roots and a name collision is impossible;
+//!   * **per-user namespace** — the root carries a per-user tag (the effective uid on Unix, the
+//!     sanitized `%USERNAME%` on Windows: `…-<tag>`), so two users sharing the temp get
+//!     structurally distinct roots and a name collision is impossible;
 //!   * **verify before sweep** — [`ensure_secure_root`] re-inspects the created namespace with
-//!     symlink-aware metadata and panics unless it is a real directory (not a symlink) owned by our
-//!     euid, with mode `0700`. A poisoned namespace fails the suite loudly; it is never swept.
+//!     symlink-aware metadata and panics unless it is a real directory (not a symlink, and no
+//!     reparse point on Windows). On Unix it additionally requires ownership by our euid, with mode
+//!     `0700`; ownership verification is Unix-only (see the fn). A poisoned namespace fails the
+//!     suite loudly; it is never swept.
 //!
 //! Not `#[cfg(test)]`: that gate does not propagate across crates, and fixtures in sibling crates
 //! (`rag-rat-core`, `rag-rat-cli`, …) consume this helper. It is not part of the semver-stable API
@@ -50,10 +54,11 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{Duration, SystemTime};
 
 /// Prefix of the dedicated system-temp sub-directory under which ALL test scratch lives. The real
-/// namespace appends this process's effective uid ([`scratch_root`] →
-/// `rag-rat-test-scratch-<euid>`) so users sharing the temp never collide. A dedicated namespace
-/// keeps the sweep's blast radius to this helper's own dirs (never the shared temp root) and makes
-/// manual cleanup a one-liner: `rm -rf $TMPDIR/rag-rat-test-scratch-$(id -u)`.
+/// namespace appends a per-user tag ([`scratch_root`] → `rag-rat-test-scratch-<tag>`: the
+/// effective uid on Unix, the sanitized `%USERNAME%` on Windows) so users sharing the temp never
+/// collide. A dedicated namespace keeps the sweep's blast radius to this helper's own dirs (never
+/// the shared temp root) and makes manual cleanup a one-liner:
+/// `rm -rf $TMPDIR/rag-rat-test-scratch-$(id -u)`.
 pub const SCRATCH_NAMESPACE_PREFIX: &str = "rag-rat-test-scratch";
 
 /// Sweep scratch dirs older than this on first use per process. Two hours is far longer than any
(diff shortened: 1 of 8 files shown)


skakri:
@codex review

crates/rag-rat-base/src/test_scratch.rs
```diff
@@ -0,0 +1,158 @@
...
+/// A unique scratch directory PATH under [`scratch_root`], keyed by `tag`, the PID, and a
+/// process-wide counter. The path is NOT created (callers create it, matching the fixtures this
+/// replaces); any pre-existing dir at the exact path is removed first.
+///
+/// On the first call per process this also sweeps stale sibling scratch dirs out of the root (see
+/// module docs) — the self-healing backstop for `Drop` cleanup that a `kill -9` skips.
+pub fn scratch_dir(tag: &str) -> PathBuf {
+    let root = scratch_root();
+    let _ = std::fs::create_dir_all(&root);
+    SWEEP.call_once(|| sweep_stale(&root));
```
skakri, line 313: On a shared system temp directory, another local user can pre-create `/tmp/rag-rat-test-scratch` as a symlink to any directory writable by the test-running user.

`create_dir_all` accepts that symlink, after which the startup sweep traverses the target and removes directories older than two hours.

Create the namespace securely and verify with symlink-aware metadata that it is a real directory owned by this process/user before sweeping it.

skakri (changes requested):

Nitjsefnie:
Good catch — fixed in `48dbb75` (incremental commit on top of the reviewed `923c8e1`, so the delta is easy to read).

Two layers close it:

1. **Per-euid namespace.** The scratch root is now `<temp>/rag-rat-test-scratch-<euid>` (`libc::geteuid()`; `libc` was already a workspace dep). Another local user pre-creating *their* path no longer intersects ours, and cleanup stays one `rm -rf`.
2. **Secure create + symlink-aware verify before every sweep** (`ensure_secure_root`): `DirBuilder` creates it `0700`, then `symlink_metadata` (which never follows) asserts the path is **not a symlink**, **is a directory**, and is **owned by our euid** — any failure `panic!`s and fails the suite loudly, naming the path. As defense in depth, `sweep_stale` also `symlink_metadata`s each entry and skips anything that isn't a real directory, so it can never follow a planted link into `remove_dir_all`.

Honest scope on the TOCTOU (documented in the code): this defeats the *pre-created* symlink — the realistic "plant it and wait" attack — because we inspect the exact path we then sweep. It does not close a swap raced *between* the check and the sweep; that would need dirfd-relative `openat`/`fstatat`/`unlinkat` pinned to the verified inode, which is out of scope for test scratch (the per-euid namespace already makes a cross-user collision structurally irrelevant). Happy to go further if you'd rather.

Verification: two new tests bite — one plants the euid-suffixed namespace as a symlink to a victim dir with an aged subdir and asserts the helper panics **and the victim survives**; one chowns the namespace to `nobody` and asserts refusal (runs for real under the root CI here, not skipped). Full suite `3029 passed / 0 failed` (baseline 3027 + the 2); `+nightly fmt` clean; clippy clean on `rag-rat-base`. Mutation-checked: removing the verification fails both attack tests; the un-suffixed name fails the namespace test.

skakri:
Provide a Windows-compatible scratch helper – on `windows-latest`, rag-rat-base now compiles this public module unconditionally, but `libc::geteuid` and the later `std::os::unix::fs` imports are unavailable. Gate the Unix implementation and provide a Windows equivalent.

Nitjsefnie:
Done in `8bc7526` (incremental on top of `48dbb75`).

**The cfg split** — gated at the smallest units, shared skeleton stays shared:
- `#[cfg(unix)]`: the euid namespace tag, the ownership assert, and the 0700 create/tighten — byte-identical code, messages, and check order to what you reviewed, just relocated behind the seam.
- `#[cfg(windows)]`: namespace tag from sanitized `%USERNAME%` (`[A-Za-z0-9_-]`, fallback `unknown`), and `verify_namespace_identity` rejects **any** reparse point via `file_attributes() & FILE_ATTRIBUTE_REPARSE_POINT` (std-only, no new deps), panic-loud in the same style. Ownership verification is documented Unix-only — per-user namespace + reparse rejection is the Windows story.
- The shared `!is_symlink()` assert already covers directory junctions on current toolchains — but only via an undocumented std implementation detail (`FileType::new` requires the `0x20000000` name-surrogate bit, and `IO_REPARSE_TAG_MOUNT_POINT` carries it — verified in the 1.95.0 std source, cited in a code comment). Since that's version-specific, the explicit attribute check is there as the version-independent backstop; it also catches non-name-surrogate redirectors (cloud placeholders).
- The two symlink/chown security tests are `#[cfg(unix)]` (they can't compile elsewhere); the sweep and its test were already portable and are untouched.

**Verification, honestly scoped:** Linux behavior is pinned by the existing suite (3029/3029, counts unchanged; the plant-symlink and foreign-owner tests still exercised for real). The Windows side is `cargo check` + `clippy --all-targets -D warnings` clean for `x86_64-pc-windows-msvc`, and the gating is mutation-verified — an ungated `libc::geteuid` reproduces exactly your E0425 on that target, the committed tree doesn't. What this box cannot do is *execute* the Windows path — your `windows-latest` leg will be its first runtime exercise; if anything trips there I'll turn it around quickly.

crates/rag-rat-base/src/test_scratch.rs
```diff
@@ -0,0 +1,434 @@
...
+///
+/// The root is created and its identity verified via [`ensure_secure_root`] before any sweep, so a
+/// namespace an attacker pre-planted as a symlink (or that another user owns) panics the suite
+/// rather than being swept.
+pub fn scratch_dir(tag: &str) -> PathBuf {
+    let root = scratch_root();
+    ensure_secure_root(&root);
+    SWEEP.call_once(|| sweep_stale(&root));
+
+    let name = format!("{tag}-{}-{}", std::process::id(), SEQ.fetch_add(1, Ordering::Relaxed));
```
skakri, line 315: Constrain the scratch tag before joining it.

When tag contains an absolute path or path separators, `format!` preserves them and `root.join(name)` can escape the dedicated namespace; the subsequent `remove_dir_all` may then delete a directory outside the scratch root.

This helper is public and does not document a trusted-input restriction, so sanitize/reject non-component tags or verify that the resulting path remains directly beneath root before deleting it.

Reject any tag that is not exactly one normal path component before constructing or deleting the scratch directory –

draft:
```
  use std::path::{Component, Path};

  fn validate_tag(tag: &str) -> anyhow::Result<()> {
      let mut components = Path::new(tag).components();

      match (components.next(), components.next()) {
          (Some(Component::Normal(_)), None) => Ok(()),
          _ => anyhow::bail!("scratch tag must be a single path component"),
      }
  }
```

Nitjsefnie:
Good catch — fixed in `2ea627d` (incremental on `8bc7526`).

**What landed:** your `validate_tag` check verbatim (exactly one `Component::Normal`, your message text), called as the *first* statement of `scratch_dir` — before the root is constructed, before any `format!`/`join`/`remove_dir_all`. Rejects separators-as-components, absolute paths, `..`, `.`, empty, and Windows prefixes. `scratch_dir`'s doc now states the restriction. Audit of the other two paths into a join/delete: `namespace_user_tag` is internal (euid int / char-whitelisted `%USERNAME%`), and `sweep_stale` deletes only `read_dir`-derived entries — a `DirEntry` file name is by construction a single component of the root, so your "verify it stays beneath root" alternative is covered structurally there.

**One adaptation, your call to reverse:** your draft returns `anyhow::Result`, but `scratch_dir` is infallible (`PathBuf`) and the module's checks are all panic-loud, so I shipped the panic form carrying your exact message. `anyhow` is already a dependency, so flipping to `scratch_dir -> anyhow::Result<PathBuf>` is one signature change away — it just ripples into the rag-rat-core/cli callers. Happy to make that change if you prefer the fallible API.

**Tests (6 new, suite 3029 → 3035, 0 failed):** rejection cases for `nested/evil`, `/tmp/evil`, `..`, `""` (each asserting your message in the panic payload), a positive case, and an escape-proof mirror of the symlink-plant test: pre-created victim dir at the exact path an escaping `../victim-…` tag would delete, asserting the victim survives. Bypassing `validate_tag` makes that test fail by *actually deleting* the victim outside the namespace — same failure chain you described — and restoring it goes green.

**One precision note so the docs don't overclaim:** trailing-separator forms (`"evil/"`, `"evil/."`) normalize to a single `Normal` component and are therefore *accepted* — contained, not escaping (every accepted form joins to a path under the namespace root; `..`/root/prefix anywhere in the string surfaces as its own component and is rejected). That's inherent to the `components()` approach in the draft. If you'd rather reject the raw-string forms too, a `contains(['/', '\\'])` pre-check on top is a two-liner — say the word.

skakri:
Thanks, merged!

PR #768, opened 2026-07-18
Fix #697: bind reconcile embed-stub accept loops to request count, not wall clock

### LINKED ISSUE #697: Flaky test: remote_reconcile_chunks_large_selection_below_sqlite_bind_limit races under load (~50%)
## Symptom
`index::ai::reconcile::embed_loop::freshness_version_tests::remote_reconcile_chunks_large_selection_below_sqlite_bind_limit` fails nondeterministically. Observed ~50% failure rate when run in isolation on a busy machine, and it reliably fails inside a full `cargo nextest run -p rag-rat-core` when heavy CPU-bound tests (e.g. `clones::refine::split::coherence_split_dense_clique_scales`) run concurrently. On a quiet machine it passes 8/8 in isolation.

## Why it's a race
The test spawns a concurrent HTTP stub (`spawn_reconcile_embed_stub`, `Duration::ZERO`) and runs a reconcile over 1005 chunks with `concurrency = 32`, then asserts `report.embeddings_written == 1005`. Under CPU saturation or socket pressure (many stub-server tests in succession) some in-flight requests don't complete before the join, so the count comes up short. It's timing/resource-sensitive, not a logic bug in the code under test.

## Impact
CI's nextest `ci` profile has **no retries** (`.config/nextest.toml`), so a bad scheduling on the runner reddens the job for an unrelated reason. It also interferes with local verification of unrelated changes.

## Options
- Make the assertion wait deterministically for all spawned requests to drain (event-based join) instead of relying on the reconcile returning with everything already written.
- Or reduce `concurrency` / bound the stub's accept loop so the test is not sensitive to host load.
- Or, as a stopgap, add `retries` for this test in the nextest `ci` profile.

Discovered while verifying unrelated graph/symbol perf work (#682/#692/#685); the change under test never touches this subsystem.

Nitjsefnie:
Fixes #697 with the deterministic direction you prescribed — but the mechanism turned out to be different from the issue's "in-flight requests don't complete before the join", and the fix follows the corrected diagnosis, so laying that out first:

## What's actually racing

The reconcile path is fully synchronous — blocking ureq in `embed_batch`, the multi-range fan-out joins its workers inside `thread::scope` before returning, writes land before the report assembles, and the test's `handle.join()` runs after the reconcile returns. Even the pre-fix stub joined accepted workers before exiting — so an accepted request always completed before the join, and "in-flight at the join" can't produce a failure. What can: **the stub dying early**. Its accept loop was a non-blocking poll under wall-clock caps (5s since spawn / 500ms idle), and the named test makes exactly **one** connection (`remote_reconcile_batch_size = max(64, 4096×32)` → one 1005-job window; `sub_batch_ranges` at ~42 chars/text → one range → one request). Under host load, the pre-connect work (skip-summary over 1005 compressed chunks, id-batched selection, cache lookups) pushes that single connect past the caps; the stub exits, the connect is refused, `write_remote_scoped_or_failed` classifies it `AbortImmediately`, and all 1005 chunks fail. The signature agrees: every observed failure is `left: 0, right: 1005` — never a partial count, which the in-flight mechanism would produce.

## The fix

Both stubs (`spawn_reconcile_embed_stub` and its sibling `spawn_selective_failure_embed_stub`, which has the same caps and whose sequential retry traffic could trip the 500ms idle cap between requests) now use the shape your own `spawn_parallel_counting_stub_with_waits` established in openai.rs: a **blocking listener accepting exactly `max_conns` connections** — event-bounded, no wall clock. Load can only make the stub wait longer, never exit early. Trade-off inherited from that precedent: if a stub user ever made fewer connections than `max_conns`, the join would hang until nextest's ci-profile slow-timeout kills it loudly — all four users' counts are exact by construction (1, 4, 1, 8; derived from the batching arithmetic, independently re-verified).

## Evidence (20-run tables, independently re-verified)

| scenario | result |
|---|---|
| before, quiet ×20 | 20/20 pass (matches the issue: quiet machines pass) |
| before, 5-core load wave ×20 | **5/20 pass** — all 15 failures `left: 0` |
| after, quiet ×20 | 20/20 |
| after, same load wave ×20 | **20/20** |
| mutation: wall-clock cap re-added (scratch, 500ms — tighter than the original, hence harsher than the 15/20) ×20 under load | 0/20, same assertion |

Full suite 3173/3173 before and after; the `freshness_version_tests` module (all four stub users) 28/28; nightly fmt clean; clippy clean core-scoped on 1.95 and full-workspace `--all-features` on 1.96 (the pre-existing `collapsible_match` at rag-rat-oplog/…/storage.rs:1452 under 1.95 is on main, untouched).

---
Generated by Claude Fable 5 (brief, review), Kimi K3 (implementation)

### DIFF
--- a/crates/rag-rat-core/src/index/ai/reconcile/embed_loop.rs
+++ b/crates/rag-rat-core/src/index/ai/reconcile/embed_loop.rs
@@ -511,7 +511,7 @@ mod freshness_version_tests {
     use std::sync::atomic::{AtomicUsize, Ordering};
     use std::sync::{Arc, Mutex};
     use std::thread;
-    use std::time::{Duration, Instant};
+    use std::time::Duration;
 
     use rag_rat_base::config::RemoteEmbeddingConfig;
     use rag_rat_base::embedding_models::{FASTEMBED_MODEL_ID, HASH_MODEL_ID, spec};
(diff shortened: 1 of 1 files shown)


skakri:
@codex review

PR #771, opened 2026-07-18
Fix #689: pin backend images by digest; bump deliberately

### LINKED ISSUE #689: cookbook: pin the vLLM image — unpinned vllm/vllm-openai:latest ships upstream regressions straight into box provisioning
`cookbook/recipes/backends.mts` builds the vLLM backend from `vllm/vllm-openai:latest`:

```ts
image: () => "vllm/vllm-openai:latest",
```

Every ephemeral chat/embed box therefore tracks upstream `latest` — a broken or
behavior-changed vLLM release propagates to all provisioning (dream verdict/compaction and
remote embedding alike) with no change on our side, and boot failures surface as opaque
`SandboxWaitUntilReady` errors rather than anything attributable to the image bump.

Suggestion:

- Pin a known-good version tag (or digest) in `backends.mts`, and bump it deliberately.
- Consider the same audit for the other backend images (infinity, ollama) while at it.
- A `log` event naming the resolved image at provision time would make future image-related
  boot failures attributable from the event stream.

Nitjsefnie:
Read your assignment-without-a-tag as "baseline the present" — every pin below freezes exactly what its tag serves today (registry-API `docker-content-digest`, cross-referenced against Docker Hub's tag list for the human-readable version), and the code comment at each names the version + date + "bump deliberately, #689". All three backends folded in per the issue's "while at it":

| image | pinned to |
|---|---|
| vLLM | `v0.25.1@sha256:e4f88a83…` (latest == v0.25.1, 2026-07-13) |
| ollama | `0.32.1@sha256:6345fbc1…` (latest == 0.32.1, 2026-07-16) |
| infinity (GPU) | `0.0.77@sha256:11e8b392…` |
| infinity (CPU) | `latest-cpu@sha256:161ef2dd…` — **see caveat** |

**The caveat worth your eyes:** `infinity:latest-cpu`'s current digest matches *no* published semver tag — including `0.0.77-cpu`, which is a different digest — so the CPU image looks stale relative to `latest`/`0.0.77` upstream. The pin still stops silent drift (that's the goal), but there's no truthful version label to attach; called out in the code comment and README rather than inventing one.

**The third ask** (provision-time log event): done in the existing `log("info", ...)` idiom from `contract.ts` — one line in each provider recipe (`modal.mts`, `runpod.mts`) naming the resolved image right where the box is created, so image-related boot failures are attributable from the event stream. (runpod's duplicate `spec.image(input)` call collapsed into the captured value while there — no behavior change.)

Cookbook `tsc --noEmit` clean; no Rust touched, so no suite run — cookbook/docs-only. Two other `:latest` mentions deliberately left (the local-dev compose snippets in the skills references and a doc-comment in `types.rs`) — outside the provisioning path this issue is about.

---
Generated by Claude Fable 5 (brief, review), Claude Sonnet 4.6 (implementation)

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -32,11 +32,15 @@ All three backends speak the **OpenAI-compatible embeddings API** (`POST <embedP
 input:[…], encoding_format:"float"}` → `{data:[{embedding,index}]}`), so the embed call and the
 readiness probe are backend-independent. Only the **route** differs:
 
-| Backend | Image | Port | Embeddings route | GPU | Model load |
+| Backend | Image (pinned, `#689`) | Port | Embeddings route | GPU | Model load |
 |---|---|---|---|---|---|
-| `ollama` | `ollama/ollama:latest` | 11434 | `/v1/embeddings` | optional | `ollama pull` after boot |
-| `infinity` | `michaelf34/infinity:latest[-cpu]` | 7997 | `/embeddings` | optional | auto-download on boot |
-| `vllm` | `vllm/vllm-openai:latest` | 8000 | `/v1/embeddings` | **required** | auto-download on boot |
+| `ollama` | `ollama/ollama:0.32.1@sha256:6345fbc18bd73a1e16404be681dbc6fd291a027cab43ed541abe78c4c81051b0` | 11434 | `/v1/embeddings` | optional | `ollama pull` after boot |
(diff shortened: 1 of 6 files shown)


skakri:
Please bump version to `0.7.0` and regenerate `package.lock`.

Nitjsefnie:
Done in `ed13d22` (incremental on `d53a527`): cookbook bumped to `0.7.0`, `package-lock.json` regenerated via `npm install --package-lock-only` — the lockfile diff is exactly the two version stanzas, nothing else churned. `npm run typecheck` clean on the result.

Issue #801, opened 2026-07-20
distill: persist source-part identity on evidence rows (V080)


skakri:
V077's `papertrail_distill_evidence` stores `(source_kind, source_id, byte_start, byte_end, quote, provenance)` but not `source_part` (title/body/comment) or the source item identity. Item title and body share the same `source_id` (the item key), so two citations from the same item with identical or overlapping spans are indistinguishable in the persisted record even though the V079 snapshot substrate (`drain.rs` `SourceSnapshot`) retains full identity at drain time.

Fix: additive migration adding `source_part TEXT` (CHECK in title/body/comment, matching V079) — and consider `source_item_kind`/`source_item_key` — to `papertrail_distill_evidence`, populated from the drain's `SourceSnapshot`. Records are derived/regenerable, so a pipeline-version bump + re-drain backfills rather than a SQL backfill. Found during #704 runtime implementation (#799). Part of #113.

skakri:
Heads-up: V080 is now taken by #802 (enriched-context snapshots: `papertrail_distill_fix_diffs` + `papertrail_distill_xrefs`). The evidence `source_part` column described here should target **V081** when implemented.

Issue #800, opened 2026-07-20
distill: enriched prompt context — fix diff, cross-referenced items, all coalesced partners


skakri:
The #704 runtime (#799) ships with xrefs/diff empty and a deterministic single-partner policy (lowest `partner_ordinal` for the singular `PromptInput.partner` slot). The #704 prompt contract calls for full enriched context:

- **Fix diff**, capped by files with symbol candidates (not a flat char prefix) — `PromptInput.diff` exists and renders but is always `None`.
- **Cross-referenced item titles + opening paragraphs** — `PromptInput.xrefs` is always empty.
- **All coalesced partners** — extraction snapshots every partner deterministically (all participate in `distill_input_hash`), but the prompt shows only one. Needs a multi-partner `PromptInput` shape (and schema/prompt-version bump) or another deterministic policy.

Requires an immutable snapshot substrate for diff/xref content (same doctrine as V079 source snapshots: never read mutable mirror/git state at drain time — it must be folded into the extraction identity hash). Deferred from the first runtime slice. Part of #113.

skakri:
Implemented in #802.

skakri:
Shipped in #802 (merged): fix diffs (V080 papertrail_distill_fix_diffs), outbound-ref xrefs, and all coalesced partners now populate the prompt context. Closing.

Issue #704, opened 2026-07-16
distill: LLM distillation pass — batch queue, guided-output ladder, selection-based prompts, model config


skakri:
The model half of the distillation feature (#113): the LLM pass that fills
`papertrail_distill_records` (#703), its prompts, its serving config, and the batch scheduler.

## Serving

- `[llm.distill]` config riding the existing ephemeral-box contract (same shape as
  `[llm.dream.remote]`). Documented default from the measured model comparison:
  **Qwen/Qwen3-30B-A3B-Instruct-2507-FP8 on L40S** — it beat the 4B on every content metric
  (rejected-alternatives on 36% vs 16% of threads; root-cause coverage 54% vs 46%; qualified
  anchors 65% vs 50%) while ~3× faster (14.2 vs 32.5 s/thread) and ~20% cheaper per thread
  (MoE, 3B active). 4B/L4 stays the constrained fallback.
- Provisioning budget is model-size aware: 30B-class needs ~25 min for the weight pull; the
  default recipe budget times out. (Related infra: #689 image pin, #690 model caching.)
- **Batch cadence with a resumable queue — never a git-hook rider.** A GPU provision takes
  minutes; sync must not block on it. Newly-closed threads enqueue at mirror sync; the distill
  pass drains the queue on its own schedule; every completed record is durable immediately
  (a died box loses nothing).

## Prompt contract (all evaluation-derived)

- Record fields as guided JSON; evidence = unit-id SELECTION (verbatim quotes by
  construction); anchors = candidate-index SELECTION from the changed files' symbols
  (`(logical_symbol_id, file, name, kind)` list — measured: selection lifts qualified anchor
  resolution, and index-id selection removes name normalization entirely).
- Plain-prose rule: no markdown tables/lists/headings/fences in field values; backticked
  identifiers only.
- "The decision is what LANDED, not what a reviewer proposed" (measured failure class), and
  claimed-vs-measured discipline for outcome numbers (projections must not be asserted as
  results).
- Enriched context, budgeted: full commit message bodies, the fix diff (capped by files with
  symbol candidates, not a flat char prefix), the coalesced partner thread, cross-referenced
  item titles + opening paragraphs.

## Output handling

Structured-output ladder with per-rung counters in the run stats (a backend regression must be
visible): backend guided decoding (per-backend capability: vLLM json_schema, Ollama format) →
strict serde (`deny_unknown_fields`) → one unguided retry parsed strictly → a single tolerant
[...]

PR #802, opened 2026-07-20
feat(distill): enriched prompt context — fix diffs, outbound-ref xrefs, all coalesced partners (#800)

skakri:
Closes the deferred enrichment slice of #704's prompt contract (#800). All three context blocks the first runtime (#799) shipped empty are now populated — snapshotted at **extraction** time, per the V079 doctrine, so the drain still never reads mutable git/mirror state.

## What ships

- **Fix diff, capped by files with symbol candidates** — new V080 table `papertrail_distill_fix_diffs`: per fixing commit, a gix first-parent tree diff renders unified patches (real `UnifiedDiff`/`ConsumeBinaryHunk`, 3-line context) for exactly the paths that yielded a SYMBOL anchor candidate. Best-effort like `merge_first_parent_paths` (bad sha/shallow/binary → no rows, never a failed pass); 8k per-file cap; merges work (the usual fixing SHA).
- **Cross-referenced item titles + opening paragraphs** — new V080 table `papertrail_distill_xrefs`: outbound `papertrail_refs` keyed by the snapshot's OWN source identities (same `item_source_text`/`comment_source_text` formats the sync mines), targets resolved in the extraction transaction with parser namespace inheritance (bare `#N` prefers the source item's kind, then deterministic kind order) and frozen as title + first body block. Unmirrored targets drop; comment/item dup sources dedupe on resolved identity.
- **All coalesced partners** — `PromptInput.partner` → `partners: Vec`, rendered in durable `partner_ordinal` order against ONE shared 14k partner budget (sequential remaining-bytes accounting, so the hard-cap invariant holds regardless of coalesce fan-out). `rules.md` pluralized.

## Identity/invalidation

Both snapshots fold into `distill_input_hash` (domain tag `v3`) with `PIPELINE_VERSION` 2→3; a referenced item's **title edit regenerates the record** exactly like a primary-source edit (test-pinned). No SQL backfill — old records regenerate under the new pipeline. `PROMPT_VERSION` 2→3 (partners Vec + hydrated blocks).

## Verification

- `cargo nextest run -p rag-rat-core`: **1456 passed** (new: V080 tip/DDL/torn-replay test, adoption re-point for both tables, fix-diff snapshot + no-repo control, xref snapshot/dedupe/unmirrored-drop, xref-title-edit regeneration, drain render test for DIFF:/REFERENCED ITEMS, multi-partner render order).
- `cargo clippy --workspace --all-targets -- -D warnings`: clean; `cargo +nightly fmt`: clean.

## Notes

- `neutralize()` quotes real `--- a/…` diff header lines (they start with the `--- ` structural marker), so rendered DIFF blocks show `> --- a/…`. Deliberate: safe, deterministic, hunks unaffected. Flagged for a future marker-precision tweak if we care.
- #801's evidence `source_part` migration must renumber to **V081** — V080 is taken here.

### DIFF
--- a/crates/rag-rat-core/src/distill/drain.rs
+++ b/crates/rag-rat-core/src/distill/drain.rs
@@ -1370,6 +1370,65 @@ mod tests {
     }
 
     #[test]
+    fn fix_diff_rows_concatenate_in_commit_path_order_with_newline_separators() {
+        // `load_fix_diff` re-sorts the persisted per-file patches by (commit_sha, path) and joins
+        // them, inserting a separating newline after any row that does not already end in one.
+        let conn = fixture();
+        seed(&conn, "first", 10);
+        // Inserted out of (commit_sha, path) order; the SECOND row (commit 'bbb') has no trailing
+        // newline, exercising the normalization branch.
+        conn.execute(
+            "INSERT INTO papertrail_distill_fix_diffs
+                 (tracker, project, item_kind, item_key, commit_sha, path, patch, repo_id)
+             VALUES ('github','org/repo','issue','first','bbb','src/z.rs','PATCH-Z-no-newline',
+                     'repo')",
+            [],
+        )
+        .unwrap();
+        conn.execute(
+            "INSERT INTO papertrail_distill_fix_diffs
+                 (tracker, project, item_kind, item_key, commit_sha, path, patch, repo_id)
+             VALUES ('github','org/repo','issue','first','aaa','src/a.rs','PATCH-A' || char(10),
+                     'repo')",
+            [],
+        )
+        .unwrap();
+
+        let jobs = load_prepared_jobs(&conn, "repo", 1, &PromptBudget::default()).unwrap();
+        let prompt = &jobs[0].rendered_prompt;
+        let a_pos = prompt.find("PATCH-A").expect("patch A renders");
+        let z_pos = prompt.find("PATCH-Z-no-newline").expect("patch Z renders");
+        assert!(a_pos < z_pos, "rows render in (commit_sha, path) order: aaa before bbb: {prompt}");
+    }
+
+    #[test]
+    fn a_null_xref_target_kind_renders_an_empty_kind_label() {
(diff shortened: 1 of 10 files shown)


skakri:
@codex review

skakri:
## Adversarial review round (2 agents, converging verdicts: FIX-FIRST) — all addressed in `6d0df4d7`

**Majors (both reviewers converged):**
1. **Identity coupled to git availability** — hashing rendered patch bytes meant a bare-index/shallow run flipped every record's hash → destroyed good snapshots + re-paid the model, and forced a full diff re-render of every record on every pass inside the IMMEDIATE write txn. **Fix:** fix diffs are a pure function of already-hashed inputs (fix SHAs + anchors), so they're no longer hashed; rendered lazily on identity change, plus a self-heal that fills missing rows when a repo handle appears. Flip-flop stability test-pinned.
2. **DoS: full-blob diff before the cap, inside the write txn** — the 8k cap truncated *after* rendering; a 50MB minified symbol-candidate file would diff in memory under BEGIN IMMEDIATE. **Fix:** 1MiB/side `try_header` pre-filter.

**Also fixed:** shallow-boundary missing parent now skips the commit (never a bogus full-tree 'everything added' patch — same fix in the pre-existing `merge_first_parent_paths`); non-blob entries (gitlinks/symlinks) skipped; control-char paths skipped; xref opening capped at 2k; xref lines render `[kind] #key (ref_kind)` so every hashed field is model-visible (kills hashed-but-unrendered spurious regen + PR-shown-as-issue confusion); snapshot cap pinned == prompt budget.

**Accepted as-is (documented):** `--- a/…` neutralization quoting; xref/partner line-shape spoofing (inside untrusted region, unciteable); unicode line separators in neutralize (pre-existing); attr-stack question (tree-diff attr source — gix uses tree/ODB attrs, no worktree dependence).

**Verification:** 1460 core tests pass (4 new adversarial-driven), workspace clippy `-D warnings` clean, nightly fmt clean.

PR #956, opened 2026-07-25
perf(watch): skip full discovery on safe clean scopes

### LINKED ISSUE #955: perf(watch): use Changed preparation for base maintenance passes
## Problem

Watcher maintenance currently enters the base index through `index_discover_reporting`, forcing `collect_index_files` and the full discovery/hash plan on every pass, including a clean unchanged checkout.

## Change

Add a reporting entry point for `IndexMode::Changed` and use it from `watch::run_pass` for the base maintenance stage.

This is safe because `index_incremental_with_progress` already promotes Changed to Discover when the active base scope is incomplete:

- the current generation/HEAD/worktree/target-fingerprint discovery marker is absent or stale; or
- the active scope file count is below the repository generation count.

That promotion retains full-tree discovery and committed-row carry for HEAD moves, target changes, bootstrap/incomplete scopes, and zero-file scopes that have not been explicitly discovered.

## Invariants

- Keep `ensure_model_manifest`, graph-index healing, generated-flag healing, stale-overlay healing, package refresh, and the watcher tail on their existing paths.
- Do not infer completeness from row presence alone; use `active_base_scope_discovered` plus the existing count check.
- Do not change startup catch-up, GC, reconcile, memory validation, or linked-overlay scope semantics.
- Preserve `IncrementalPassReport` content-change and clone-delta hints.

## Verification

- clean, unchanged maintenance pass uses Changed preparation and does not call full-tree discovery
- HEAD advance promotes to Discover and carries/restamps unchanged committed rows
- target-fingerprint change promotes to Discover and adds/prunes target files
- incomplete/bootstrap and explicitly discovered zero-file scopes retain current behavior
- dirty/untracked/deleted files remain covered by `git_changed_paths`
- graph/generated heals and memory validation still run on the same conditions

Follow-up to #831.

skakri:
## Summary

- enter watcher base maintenance through Changed preparation only for clean Git checkouts with a complete active scope and no worktree-scoped rows
- retain authoritative discovery for non-Git roots, dirty paths, ignore-rule changes, symlink replacements, and stale untracked or modified rows
- preserve the existing automatic discovery promotion for bootstrap, HEAD moves, and target changes
- retain content-change and clone-delta reporting for the watcher tail

Fixes #955

## Verification

- `cargo +nightly fmt --all --check`
- `cargo check -q -p rag-rat-core --tests`
- `cargo nextest run -p rag-rat-core` — 1593 passed, 2 skipped
- `cargo clippy -q -p rag-rat-core --all-targets -- -D warnings`
- `git diff --check`

Regression coverage includes non-Git maintenance, vanished untracked files, ignore-rule changes, and changed symlinks.

### DIFF
--- a/crates/rag-rat-core/src/index/incremental.rs
+++ b/crates/rag-rat-core/src/index/incremental.rs
@@ -81,8 +81,23 @@ impl IndexDatabase {
     where
         F: FnMut(IndexProgress),
     {
-        Self::index_incremental_with_progress(config, IndexMode::Changed, None, &mut progress)
-            .map(|(db, _)| db)
+        Self::index_incremental_with_progress(
+            config,
+            IndexMode::Changed,
+            None,
+            false,
+            &mut progress,
+        )
+        .map(|(db, _)| db)
+    }
+
+    /// Watcher-safe reporting entry point. A complete, clean Git scope with no worktree-scoped
+    /// rows uses the cheap `Changed` preparation; every ambiguous case retains authoritative
+    /// discovery semantics (non-Git roots, dirty paths, and stale untracked/modified rows).
+    pub(crate) fn index_watch_reporting(
+        config: &Config,
+    ) -> anyhow::Result<(Self, IncrementalPassReport)> {
+        Self::index_incremental_with_progress(config, IndexMode::Changed, None, true, &mut |_| {})
     }
 
     /// Reconcile exactly the supplied candidate `paths` (#659) — the edit-driven-reindex substrate.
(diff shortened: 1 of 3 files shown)


skakri:
Withdrawing this implementation. Scope completeness plus clean Git status is not a sufficient proof that the filesystem and index agree: tombstones, suppressed Git index entries, parser failures without file rows, and other semantic states can remain hidden.

The replacement direction is tracked in #958: carry event provenance into maintenance, use explicit-path reconciliation for safe event-driven mutations, and retain authoritative discovery for startup, periodic, classification-changing, and ambiguous passes.

PR #959, opened 2026-07-25
fix(dream): ground divergence verdicts in reviewed evidence

### LINKED ISSUE #954: dream-verify: restore divergence precision with reviewed-finding replay and claim-grounded evidence
## Problem

`dream --verify` still over-classifies incidental identifier absence as memory divergence after the targeted #678 / #686 fixes.

A production review on 2026-07-25 at `c5303848` started with 36 open `memory_divergence` findings produced by `Qwen/Qwen3-4B-Instruct-2507` under `verify-pack-v3`:

- 3 were accepted as real, intentional `note_ahead` findings.
- 33 were dismissed after checking the memory and current source.

That is 3/36 precision (8.3%) for this reviewed batch. The stored population currently contains 75 `diverged` and 340 `current` model verdicts, but those totals are not ground truth; the reviewed batch is the actionable signal.

The committed `verify_pack_test` gate does not represent this workload. Its manifest has 14 cases: 12 current and 2 diverged. It covers short, hand-selected notes and one doctored code-ahead case, but not the long historical/design memories that dominate the live false positives.

This follows #678 rather than reopening it. #678 fixed memory-id cross-references, text-present wording, and explicit removed/renamed/absent prose. PR #686 explicitly left broader namespace, cross-language, and evidence-precision work out of scope.

### 1. Index coverage is reported as whole-tree absence

The pack says `NOT FOUND anywhere in the source tree`, but the resolver can only prove absence inside indexed source/chunk text. Tokens living primarily in non-indexed configuration, workflow, package, or cookbook files were treated as authoritative absences:

- `NODE_AUTH_TOKEN`
- `minimatch`
- `readinessProbe`
- `__imp_tolower`
- `workflow_call`
- cargo-deny keys such as `unmaintained`, `unsound`, `vulnerability`, and `yanked`

The wording overstates the evidence and encourages the model to convict.

### 2. Identifier granularity or namespace does not match the symbol index

Memories cite fields, enum variants, qualified members, test-name fragments, or architecture vocabulary that is present semantically but not represented by an exact `symbols.name` row:

- `FoldState::enrollment_roles`
- `ReferenceShape::UnqualifiedOnly`
- `ConfigSnapshot::trusted_path`
- `remote_node_id`
- `sync_invites`
- `below_sqlite_bind_limit`
- `allows_already_indexed_to_go_empty`
- `never_echoes`

`extract_identifiers` admits every backticked span plus every long snake_case token, but the rendered row does not retain extraction provenance, namespace, or note-local context.

### 3. Historical or incidental names are mistaken for load-bearing current claims

Long memories contain measurements, old helper names, renamed paths, test fixtures, and descriptions of earlier implementation stages. The model cited these incidental absences instead of the memory's current invariant:

- `clear_reconcile_meta`
- `write_bytes`
- `pre_owner`
- `body_hash`
- `index/papertrail/autosync.rs`
- `src/pkg/lib.rs`

The prompt already says absence alone is not divergence, but the small model does not reliably recover the role of one identifier from a long note.

### 4. The citation guard proves provenance, not contradiction

`verdict_is_cited` correctly checks that each evidence line came from the pack. It does not check that the model grounded the other half of the contradiction in the note.

One reviewed false positive cited:

```text
`content_digest::ensure_content_digest(conn)` -> not a defined symbol; appears verbatim as source text
```

The prompt explicitly defines that resolution as presence, yet the `diverged` verdict passed because the citation was real.

### 5. Genuine note-ahead findings must remain detectable

The accepted findings are important recall traps, not noise:

- `discover_peers`
- `own_account`
- `gouqi`

Any precision fix must continue to surface a note that deliberately describes work absent from the checkout.

## Root cause hypothesis

This is now an evidence-representation and evaluation problem more than a prompt-wording problem:

1. `extract_identifiers` loses where and how an identifier was mentioned.
2. `resolve_identifier` has to collapse many namespaces into `Symbol | File | TextPresent | Absent | Unresolvable`.
3. `render_pack` gives the model an absence row without the note-local claim that would make the absence contradictory or incidental.
4. `verdict_is_cited` validates only the repository-evidence citation.
5. The eval corpus is too small and too unlike production to catch the failure.

### A. Build a reviewed-finding replay set first

Capture representative accepted and dismissed findings from the production review as a committed, sanitized verifier corpus. Include at least:

- non-indexed config/workflow/package tokens;
- fields, variants, qualified members, and test-name fragments;
- historical/renamed/incidental identifiers in long memories;
- text-present citations that must not support divergence by themselves;
- genuine `note_ahead` findings that must stay diverged.

Record the current `verify-pack-v3` baseline before changing resolver or prompt behavior.

### B. Make evidence confidence and provenance explicit

Explore a structured identifier row carrying enough information to avoid overstating the evidence:

- extraction source: backtick vs incidental snake_case;
- note-local sentence or bounded context;
- resolution namespace/category;
- whether absence is authoritative for the indexed language/file coverage;
- whether the row is presence, absence, or indeterminate.

At minimum, stop rendering coverage-limited misses as “anywhere in the source tree.” Prefer `Unresolvable`/indeterminate when the index cannot prove absence for that namespace.

### C. Ground both sides of a divergence

Evaluate requiring a structured `CLAIM:` citation copied from the note in addition to repository `EVIDENCE:`. Guard both deterministically:

- the claim must be a verbatim note line/span;
- repository evidence must be a citable pack row;
- `diverged` needs at least one actual contradiction pair;
- a `TextPresent` row alone cannot support divergence unless the claim specifically requires a defined symbol/file.

This should make long-note incidental-token failures rejectable without trying to encode tense and historical semantics entirely in regexes.

### D. Keep prompt/version/eval lifecycle coherent

Any rendered-pack or prompt change must:

- bump `VERDICT_PROMPT_VERSION`;
- regenerate `verify-packs.json` through the shipped dump path;
- re-derive expected labels;
- update the mirrored eval prompt;
- preserve churn-key and stale-verdict invalidation invariants.

## Acceptance criteria

- [ ] Add a reviewed replay corpus covering every false-positive class above and the three genuine note-ahead cases.
- [ ] Publish the `verify-pack-v3` baseline on that corpus before implementation.
- [ ] Classify absence as authoritative vs indeterminate using actual index coverage; do not claim whole-tree absence when coverage cannot establish it.
- [ ] Add deterministic guards/tests for contradiction grounding, including rejection of a `diverged` verdict supported only by a text-present row.
- [ ] Preserve all genuine note-ahead replay cases as `diverged`.
- [ ] Reduce false positives on the reviewed 36-case replay from 33 to at most 3 without losing the accepted cases.
- [ ] Regenerate the existing verification corpus, bump prompt version when required, and keep the production/eval prompt byte-equivalent.
- [ ] Add unit tests around identifier extraction/resolution provenance and verdict acceptance, plus the offline model gate.

## Likely surfaces

- `crates/rag-rat-query/src/memory/evidence.rs`
  - `extract_identifiers`
  - `resolve_identifier`
  - `IdentifierResolution` / `ResolutionKind`
- `crates/rag-rat-dream/src/verdict.rs`
  - `render_pack`
  - `parse_verdict`
  - `verdict_is_cited`
- `crates/rag-rat-dream/src/prompts/verdict_head.md`
- `evals/memory-compaction/`

Related: #421, #428, #678, #686, #695.

skakri:
## Summary

Ground dream divergence verdicts in reviewed evidence so a compact verifier model stops hallucinating stale memories:

- strengthen the verifier prompt and parser with a required, note-grounded `CLAIM`: a `diverged` verdict must copy one load-bearing claim **verbatim and in full** from the note (punctuation/case-insensitive), so a copied prefix cannot ground invented prose appended after it
- reject divergence resting only on TextPresent evidence, and require divergence claims to cite evidence identifiers that occur in the copied claim
- bound absence authority by indexed coverage: a terminal identifier miss is authoritative only when the note is unbound or **every** bound source path resolves to an indexed file — partial coverage stays indeterminate, and qualified `::` member references are never treated as conclusively absent
- add a fixed 36-case replay from reviewed production findings, a scoring harness that mirrors the production acceptance guard exactly (per-evidence, per-line existential predicate), regression tests, and a model-candidate evaluation
- point the `real_29` corpus record's prompt binding label at the resolved `precompute/build.rs` path so the verification gate exercises the same prompt as production
- compare larger dense and MoE candidates on the same replay; retain Qwen3-4B because it preserves the best precision/recall tradeoff

The reviewed replay improves from 33 false positives to 0 while retaining 2 of the 3 genuine note-ahead findings.

## Verification

- `cargo +nightly fmt --all -- --check`
- `cargo nextest run -p rag-rat-dream` (78 passed)
- `cargo nextest run -p rag-rat-query` (193 passed)
- `cargo clippy -p rag-rat-dream -p rag-rat-query --all-targets`
- compact verifier gate: Qwen3-4B 11/14 (78.6%)
- reviewed replay: Qwen3-4B 0/33 false positives, 2/3 true positives (100% precision, 66.7% recall); Qwen3-8B 0/33 fp, 1/3 tp; Qwen3-Coder-Next-FP8 and gemma-3-12b worse

Closes #954.


skakri:
@codex review

Issue #581, opened 2026-07-10
Flaky: adoption_hints tests fail under parallel test execution (coverage job)


skakri:
### Symptom

`index::adoption_hints::tests::*` git-fixture tests intermittently fail when run in **parallel** (the `coverage` CI job runs `cargo test` single-process/multi-threaded under llvm-cov). Observed on the coverage job for an unrelated PR: `same_identity_join_warns_even_after_a_read_only_open_of_the_clone` panicked at the test `git` helper with `git [...] failed` (a git subcommand returned non-zero). A re-run passed.

### Reproduces locally

```
# parallel: multiple adoption_hints tests fail (varies run to run)
cargo test -p rag-rat-core --lib --no-default-features adoption_hints

# serial: all 11 pass, every time
cargo test -p rag-rat-core --lib --no-default-features adoption_hints -- --test-threads=1
```

So it's a **parallel-safety bug in the tests themselves**, not product logic.

### Likely cause

The module's `git` helper (`adoption_hints.rs:290`) runs `git` in a temp repo but does **not** isolate git from the ambient environment — it sets `GIT_AUTHOR_*`/`GIT_COMMITTER_*` but not `HOME`, `GIT_CONFIG_GLOBAL`, or `GIT_CONFIG_SYSTEM`. Under parallelism the concurrent git invocations read/contend on the ambient `~/.gitconfig` (and `git init` depends on the ambient `init.defaultBranch`). A hostile ambient config (e.g. a global `core.hooksPath`) makes it worse.

### Suggested direction

Isolate the test `git` helper from the environment:
- `GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_SYSTEM=/dev/null`, `HOME=<temp>`
- `git init -b main` (or `-c init.defaultBranch=main`) so it doesn't depend on ambient config

This mirrors the hardening other git-fixture helpers already use and would make the module parallel-safe.

### Impact

Intermittent red `coverage` job on unrelated PRs; masks real failures and costs re-runs.

PR #975, opened 2026-07-26
fix(tests): isolate fixture git invocations from the ambient environment

### LINKED ISSUE #970: Test git helpers outside adoption_hints still inherit the ambient git environment
**What happens**

Several test helpers shell out to `git` without isolating the environment, so tests inherit whatever git configuration the developer or CI runner happens to have. That makes them pass or fail depending on the machine rather than the code — the same class of flake as #581.

**Where**

Still un-isolated:

- `crates/rag-rat-core/src/index/oracle_surfacing_tests.rs`
- `crates/rag-rat-core/src/index/git_context.rs`
- `crates/rag-rat-core/src/index/repo_identity.rs`
- `crates/rag-rat-core/src/index/trackers.rs`
- `crates/rag-rat-core/src/mcp/tests.rs`
- `crates/rag-rat-cli/tests/common/mod.rs`

For reference, `crates/rag-rat-cli/src/commands/remove.rs` already does this correctly, and `adoption_hints.rs` was fixed in #969.

**Why / mechanism**

The helpers run `git init` / `git add` / `git commit` with the ambient environment, so anything in the user's or system's git config applies inside the test fixture — `core.hooksPath`, `commit.gpgsign`, `init.defaultBranch`, `user.*`, templates, and so on. Bare `git init` in particular takes its branch name from the ambient `init.defaultBranch`, so a fixture's branch differs between machines.

**How to reproduce**

Point `HOME` at a directory whose `.gitconfig` sets `core.hooksPath` to a hooks directory containing an `exit 1` `pre-commit` hook, then run the affected suites. Fixture creation fails inside the helper. This is the same setup that reproduced #581 deterministically — before that fix it gave 6 of 11 failures in `adoption_hints`, all `git ["commit", "-q", "-m", "init"] failed`.

**Expected**

The helpers pin the environment the way `remove.rs` and (now) `adoption_hints.rs` do:

- `GIT_CONFIG_GLOBAL=/dev/null` and `GIT_CONFIG_SYSTEM=/dev/null`
- `HOME` set to the test's own temp root
- `git init -b main` explicitly, rather than relying on the ambient default

**Note**

#581's description says the hardening "mirrors" the existing helpers, but as far as I can tell `remove.rs` was the only hardened one at the time — hence this follow-up covering the rest.

If nobody picks this up, I will take it in a few days.

skakri:
## Summary
- new `rag_rat_base::test_git` seam (`command` / `run` / `output`) that pins the whole class at one place: `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM=/dev/null`, `HOME` pinned to the fixture dir, pinned author/committer identity, and command-scope `init.defaultBranch=main` + `core.autocrlf=false`
- every test-code `git` invocation delegates to it — the six files listed in #970 (`cli/tests/common`, `oracle_surfacing_tests`, `git_context`, `repo_identity`, `papertrail/trackers`, `mcp/tools/tests`) plus the same-class helpers the file audit surfaced: `schema_bootstrap_tests` (previously pinned only `core.autocrlf`), `config/tests`, `distill/extract`, `papertrail_autosync`, `git_history/tests`, `watch/tests`, `ignore_rules`, `cli` unit tests (`index_ops`, `init/mod`, wizard steps), `index_guards`, `init_yes`, `multi_repo_e2e`
- the `init --yes` integration tests isolate the production subprocess from the ambient config: hook install is a real user-facing path that honors the environment by design, so the harness pins `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` to keep the test machine-independent
- `test_git`’s own tests include a guard-the-guard: a hostile global `core.hooksPath` with a failing pre-commit hook really aborts a commit without the helper, so the isolation can never silently degrade

Benches and production paths (`build.rs`, eval replay) intentionally keep the ambient environment; `--version` probes stay ambient by design.

## Verification
- `cargo check --workspace --tests --benches`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo +nightly fmt --all --check`
- `cargo nextest run --workspace` (3912 passed) under an isolated `TMPDIR` with `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` pointed at a config that sets a failing `core.hooksPath`, `core.autocrlf=true`, `init.defaultBranch=dev`, and a foreign `user.*` — the #581/#970 repro; the scratch namespace contained no fixture directories after the run
- `git diff --check`

Closes #970.

### DIFF
--- a/crates/rag-rat-base/src/config/tests.rs
+++ b/crates/rag-rat-base/src/config/tests.rs
@@ -53,7 +53,7 @@ fn config_load_resolves_main_and_linked_worktrees_to_one_database() {
     // The actual guarantee (review item 1): Config::load from the main worktree and from a
     // linked worktree of the same repo produce the *same* database path — not two DBs.
     let git = |dir: &Path, args: &[&str]| {
-        std::process::Command::new("git").arg("-C").arg(dir).args(args).output().unwrap()
+        crate::test_git::run(dir, args);
     };
     let tmp = scratch("cfgload");
     let main = tmp.join("main");
@@ -330,8 +330,7 @@ fn repo_id_override_is_parsed_and_does_not_change_the_database_path() {
 /// identity).
 fn git_commit_all(dir: &Path) {
     let git = |args: &[&str]| {
-        let out = std::process::Command::new("git").arg("-C").arg(dir).args(args).output().unwrap();
-        assert!(out.status.success(), "git {args:?}: {}", String::from_utf8_lossy(&out.stderr));
+        crate::test_git::run(dir, args);
     };
     git(&["init", "-q"]);
     git(&["config", "user.email", "t@e"]);
@@ -373,8 +372,7 @@ fn config_load_without_a_database_key_resolves_to_the_global_database() {
 #[test]
 fn config_load_in_a_linked_worktree_is_governed_wholesale_by_the_main_config() {
     let git = |dir: &Path, args: &[&str]| {
-        let out = std::process::Command::new("git").arg("-C").arg(dir).args(args).output().unwrap();
-        assert!(out.status.success(), "git {args:?}: {}", String::from_utf8_lossy(&out.stderr));
+        crate::test_git::run(dir, args);
     };
     let tmp = scratch("wholecfg");
     let main = tmp.join("main");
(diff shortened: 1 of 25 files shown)


skakri:
@codex review

PR #976, opened 2026-07-28
fix(index): resolve Rust method calls by receiver type

dip-labs:
## Summary

Improve Rust call-graph precision when multiple types define methods with the same name. The index now preserves method ownership and uses conservative, scope-aware receiver-type evidence to resolve calls such as `value.render()` to the correct `impl` method instead of treating every same-named method as an equivalent candidate.

## Problem

Rust method names are not globally unique. A repository may contain many unrelated implementations such as:

```rust
impl JsonRenderer { fn render(&self) {} }
impl HtmlRenderer { fn render(&self) {} }
```

Name-only edge resolution loses the owner type and can produce ambiguous or incorrect references/callers. This makes graph navigation noisy for real repositories and prevents reliable receiver-aware code intelligence without a full compiler session.

## Changes

- Preserve Rust method owners in logical symbol identities and graph records.
- Extract conservative receiver-type evidence from Rust syntax, including relevant bindings and method-call contexts.
- Carry receiver hints through unresolved edge records and fingerprinting.
- Resolve same-named Rust methods using receiver type while retaining safe fallback behavior when evidence is unavailable.
- Keep inference scope-safe so bindings from unrelated functions or nested scopes cannot contaminate resolution.
- Update graph persistence/schema migration handling for the new receiver metadata.
- Extend query-memory resolution and validation to retain and verify the stronger symbol identity contract.

## Safety properties

- Resolution is conservative: an inferred receiver narrows candidates only when the syntax provides usable evidence.
- Ambiguous or unsupported cases retain fallback behavior rather than inventing a target.
- Receiver evidence participates in edge identity/fingerprints, preventing distinct typed calls from collapsing into one edge.
- Existing non-Rust extractors pass no receiver metadata and keep their prior behavior.

## Tests

Added and strengthened coverage for:

- two types exposing the same method name;
- strict owner isolation in symbol lookup;
- receiver-aware edge resolution and fingerprints;
- local binding inference and scope isolation;
- graph persistence and schema migration;
- multi-repository scope and memory validation;
- fallback behavior when receiver type is unknown.

The patch was first validated in `dip-labs/rag-rat#1`. Its full GitHub Actions matrix passed, including build, test, rustfmt, clippy, clippy with all features, coverage, eval, oracle matrix/report, and benchmark checks.

## Scope

This PR intentionally does not add a language server or compiler dependency. It improves the existing parser/index graph with conservative receiver-aware resolution while preserving current fallback semantics.


skakri:
@codex review

skakri (changes requested):
Please see raised issues by automated review then I'll do a final manual pass (this is a pretty big change). Thanks!

crates/rag-rat-core/src/index/graph_index.rs
```diff
@@ -764,14 +793,15 @@ impl IndexDatabase {
...
-                   symbols.end_line, symbols.file_id
+                   qn.value, COALESCE(symbols.scope_path, ''), symbols.kind, symbols.signature,
+                   symbols.start_line, symbols.end_line, symbols.file_id
             FROM main.symbols AS symbols
             JOIN main.files ON main.files.id = symbols.file_id
             LEFT JOIN main.name_strings qn ON qn.id = symbols.qualified_name_id
             WHERE main.files.repo_id = ?1 AND main.files.generation = ?2{extra_where}
             ORDER BY symbols.language, main.files.path, symbols.name, qn.value,
-                     symbols.kind, symbols.signature, symbols.start_byte, symbols.end_byte
+                     COALESCE(symbols.scope_path, ''), symbols.kind, symbols.signature,
```
skakri, line 813: Please order by the normalized scope key; the SQL ordering must use the same normalized value as `LogicalSymbolKey` (or grouping must retain already-seen normalized keys). This is the source of the `UNIQUE constraint failed: logical_symbols.id`.

skakri (changes requested):
See comment.

crates/rag-rat-core/src/index/languages/rust/edges.rs
```diff
@@ -245,3 +247,778 @@ pub(super) fn rust_impl_edges(
...
+    let mut stack = vec![root];
+    while let Some(current) = stack.pop() {
+        let mut cursor = current.walk();
+        for child in current.named_children(&mut cursor) {
+            match child.kind() {
+                "impl_item" => {
+                    let impl_type_matches = child
+                        .child_by_field_name("type")
+                        .map(|type_node| degeneric_path(&node_text(type_node, text)))
+                        .is_some_and(|impl_type| qn_tail(impl_type.trim()) == owner_tail);
```
skakri: When one file contains modules with identically named constructor owners, this tail-only comparison  can select the wrong impl. For example, `a::Factory::new()` may be classified using `b::Factory::new()`  -> WorkerB because both impl types reduce to Factory and the first traversal match is returned.

Fix it by comparing a canonical module-qualified owner, not the final type segment.

A helper such as this is the clean seam:
```rust
  fn module_qualified_type_path(
      context: Node<'_>,
      raw_type: &str,
      text: &str,
  ) -> Option<String>
```
It should:

- Collect enclosing mod_item names, outermost first.
- Remove generics.
- Resolve crate:: from the file root.
- Resolve self:: from the current module.
- Resolve leading super:: by popping module components.
- Treat an otherwise relative path as relative to the current module.
- Decline unsupported forms such as UFCS and associated-type projections.

Then replace `qn_tail(impl_type.trim()) == owner_tail` with an exact comparison of canonical qualified paths.

Also avoid returning the first tail match. If the helper cannot distinguish candidates, gather matching candidates and return CtorReturn::Opaque when more than one remains. A missing same-file definition can remain Unknown, preserving the external-constructor convention fallback; an ambiguous same-file definition must not fall back.

Add regressions for:

- a::Factory and b::Factory returning different types.
- An unqualified Factory::new() inside mod a.
- A root call to a::Factory::new().
- self::Factory, crate::a::Factory, and super::Factory.
- Ambiguous/unresolvable ownership producing no receiver hint.

The conservative ambiguity fallback is important: a missing hint loses one possible edge, while a wrong hint creates a confidently incorrect call-graph target.

skakri (changes requested):
Interrogated Codex for a bit.

```rust
  mod a {
      impl Factory {
          fn new() -> WorkerA { WorkerA }
      }

      fn make() {
          let worker = Factory::new();
          worker.run();
      }
  }

  mod b {
      impl Factory {
          fn new() -> WorkerB { WorkerB }
      }
  }
```
The relevant identities should be:

call owner:  a::Factory
first impl:  a::Factory
second impl: b::Factory

So only WorkerA may become the receiver hint.

skakri:
@codex review

skakri:
I suggest introducing one explicit Rust type-identity classification rather than adding independent guards, e.g.:
```rust
  enum ReceiverTypeIdentity {
      LocalQualified(String),
      LocalUnqualified(String),
      ExternalQualified(String),
      Ambiguous,
  }
```

The shared rule should be:

1. Canonicalize qualification and generics while retaining module/crate identity.
2. Use that same canonical key for ordering, grouping, lookup, and comparison.
3. Only use a tail fallback after proving the reference is local and the tail has exactly one viable target.
4. Decline resolution on external or ambiguous identities.

- constructor lookup must accept only an exact LocalQualified match, or a unique local unqualified match.
- receiver resolution must never tail-match ExternalQualified or Ambiguous.
- tail matching becomes an explicitly conservative fallback instead of the default normalization strategy.

skakri:
While untangling the symbol rebuild and related logic, I found a few issues that should be addressed now. I’ll commit those changes in this PR (better to land them together), then I'll fire the CI gates -> merge.

skakri:
This really opened a can of worms. PR has outgrown its title, spanning extraction, symbol identity, resolution precedence, etc. Splitting off related work and merging.

skakri:
Split off to #1023, you seeded the identity work now squashed into it, I co-authored your changes. Closing this pr.

Issue #996, opened 2026-07-28
Live oracle: project-resolution changes do not invalidate live verdicts


skakri:
A live verdict's covered-skip key is `(tool, tool_version, source_path, file_sha, commit, worktree)`.
Nothing in it reflects the *project configuration* that decided how the callee resolved, and the
watcher's worklist only ever contains files in the backend's own language. So a change that alters
module resolution without touching a caller's source bytes leaves the old verdict marked current,
and it is never re-resolved.

Concretely, for `ts-lsp`: editing `tsconfig.json` `paths` aliases or `references` can repoint an
import at a different file. The caller's `.ts` bytes are unchanged, so its `file_sha` is unchanged,
so the covered-skip gate treats the existing verdict as still current. `tsconfig.json` itself never
enters the worklist (`LiveBackend::claims_path` admits only the backend's language), so nothing
triggers a re-resolve.

This is not TypeScript-specific — `ra-lsp` has the same exposure to `Cargo.toml` edits that change
features, targets, or path dependencies. It predates the TypeScript backend; #536 widened the blast
radius rather than introducing it.

Impact is bounded but real. The batch pass writes under its own tool id and wins on overlap, so any
edge the batch also covers masks a stale live verdict on read. A checkout running live without a
batch baseline keeps the stale one.

Fixing it properly needs a decision on each of:

- Which paths count as project-resolution inputs per backend (`tsconfig.json`, `package.json`,
  lockfiles; `Cargo.toml`, `Cargo.lock`, `.cargo/config.toml`), and whether that list lives on
  `LiveBackend` next to `language`/`readiness`.
- What such a change should do — clear the affected coverage, or enqueue the project's callers.
  Clearing is cheaper; enqueueing keeps the budget accounting honest.
- How to scope the blast radius, since a root config change can cover the whole checkout and a
  naive re-enqueue would flood the request budget for many passes.

Deliberately not bolted onto #536: it touches the shared live path for every backend and wants its
own design pass and worktree-scoped tests.

skakri:
The C/C++ backend adds the sharpest instance of this, and it is worth recording here rather than
opening a duplicate.

A `compile_commands.json` is regenerated in place by every CMake configure, and the flags it carries
are what select which preprocessor branch a translation unit compiles. So the same call can resolve
to a different definition after a rebuild, with no caller byte changed — the covered-skip key holds,
the database is not in the backend's language so it never enters the worklist, and the old verdict
stays marked current.

[...]

Issue #536, opened 2026-07-09
SCIP phase 6 slice 4: additional live LSP backends (TS / clangd / Kotlin)


skakri:
Slice 4 of #74 (mirrors #71/#72 for the batch side, but for the live path).

`typescript-language-server` / `clangd` / Kotlin LSP reuse the same client substrate from slice 1 — each is a manifest entry (binary, args, readiness signal, encoding quirks), not new protocol code. Each gets its own distinct `OracleTool` live variant + `tool_version` scheme, consistent with slice 2's tool-identity decision.

Deferred until the Rust live path (slices 2–3) is proven.

Part of #74 / #61.

skakri:
TypeScript shipped in #992 (`6f62726b`). `clangd` and Kotlin remain.

What landed, beyond the backend itself: the live path is no longer shaped around one server. A new
backend is now two registry entries — `LiveBackend::for_tool` (language, readiness policy, per-extension
`languageId`) and `ToolManifest::for_tool` (program, `live_args`, prerequisite) — plus the `OracleTool`
variant and its CLI selector. Spawn argv, readiness signal, `languageId`, worklist filter, sentinel
namespace, and watcher state were all previously hardcoded to Rust and are now per-backend.

The finding that shaped it, for whoever picks up the next backend: readiness is a correctness gate,
not a latency optimisation. A warming `typescript-language-server` does not answer `null` — it resolves
an imported callee to the *import statement* in the calling file, a plausible non-null the write path
would persist and never revisit. Do not assume a new server degrades to nulls; measure what it actually
answers before it is ready, and what signal (if any) it emits to say so. `ReadinessPolicy` has two
shapes so far: an explicit quiescence notification (rust-analyzer) and the LSP-standard work-done
progress cycle (TypeScript).

Two things to carry forward:

- The request budget is shared across backends and claimed in rotating order. A third backend needs no
  change there, but should not assume it gets a full allowance.
- `#996` tracks stale verdicts when project configuration changes without changing a caller's bytes.
  It applies to every live backend, Rust included, and is worth resolving before the backend count grows.

skakri:
State update: **clangd has shipped; Kotlin is the only backend left in this slice**, and its registry blocker is now gone.

The shape a new live backend declares has changed since the comment above. It is still a `LiveBackend` entry plus a `ToolManifest` entry, but the project half is now one `ProjectModel` (#1069) rather than three loose fields:

[...]

Issue #74, opened 2026-06-12
Live LSP oracle — rust-analyzer hardening and additional backends remaining


skakri:
The live LSP oracle produces the same `edge_oracle` resolution shape as the batch SCIP pass, but refreshes changed files at watcher cadence. The batch pass remains canonical; live evidence is a freshness patch for files being edited.

## Status

- [x] Native LSP client substrate: process/protocol/position handling and per-file definition resolution (#531).
- [x] Rust watcher integration, atomic `edge_oracle` persistence, definition-to-symbol mapping, and deterministic batch precedence (#534, #972).
- [ ] rust-analyzer production hardening: readiness, bounded requests, crash recovery, idle lifecycle, and required server-request handlers (#535).
- [ ] Additional live backends for TypeScript, clangd, and Kotlin (#536).

## Settled invariants

- Live rust-analyzer evidence uses the distinct persisted `ra-lsp` tool identity so it cannot make the batch oracle appear current.
- `[oracle.live]` is opt-in and independent of batch `auto_run`. Without a batch baseline it can still upgrade edge resolution, but has no batch monikers for clone collapse or memory anchoring.
- Live rows copy the resolved target's batch moniker when available and otherwise use a SCIP-local sentinel; LSP moniker strings are never persisted.
- Verdict rows and their backing run row commit atomically, and batch evidence wins when both tools cover the same edge.

Part of #61.

skakri:
Descoped after an audit vs HEAD.

**Shipped:** slice 1 — the LSP client substrate (#531, `crates/rag-rat-oracle/src/lsp/`).

**Remaining:** slice 2 watcher wiring + `edge_oracle` write + def→symbol mapping (#534), slice 3 rust-analyzer hardening (#535), slice 4 additional live backends (#536). Confirmed no `OracleTool::RaLsp` variant / `ra-lsp` write path exists at HEAD yet.

Title unchanged in spirit — this is the tracking issue; retitled only to name the shipped substrate.

skakri:
Current status: #531 and #534/#972 have shipped. The critical path is now #535: gate definition consumption on rust-analyzer readiness and bound requests so a warming or wedged server cannot silently drop work or hold the repository write lock indefinitely. Additional backends in #536 remain deferred until that Rust lifecycle is hardened.

skakri:
[...]

PR #999, opened 2026-07-28
feat(oracle): add a live clangd LSP backend for C and C++ (#536)

skakri:
Third live backend after Rust (#972) and TypeScript (#992), and the first that serves **two**
languages from one session — `LiveBackend.language` becomes a set, which the watcher's target gate
and the worklist filter now consult.

Part of #74. Kotlin is the remaining slice of #536.

## The trade that defined this backend

clangd resolves a call across translation units only through its background index, and it persists
that index to `.cache/clangd/` **inside the checkout**. Nothing relocates it — `--help-list` has no
index-path option, and `XDG_CACHE_HOME` is ignored (verified: the index still landed in the source
tree). Measured on a C fixture where `main.c` calls `greet()`, declared in `lib.h` and defined in
`lib.c`:

| mode | a cross-TU call resolves to | readiness signal | writes into checkout |
|---|---|---|---|
| background index on (clangd's default) | the **definition** in `lib.c` | `$/progress` "indexing" | `.cache/clangd/index/*.idx` |
| `--background-index=false` | the **declaration** in `lib.h` | none at all | nothing |

The write-free mode is close to worthless: resolving a call to a header prototype disagrees with
what the batch `scip-clang` backend resolves for the same edge, and adds little over the tree-sitter
heuristic. So the write is accepted deliberately, and `--background-index` is pinned rather than
left to clangd's default — a future clangd flipping that default would otherwise silently gut
resolution.

Shipping that means flooring `.cache/clangd` out of indexing: a large, entirely machine-written
index tree does not belong in the discovery walk, and anything source-shaped appearing under it
would otherwise be classified as first-party code — the same reason `.rag-rat` is floored. (Its
`.idx` artifacts carry no target extension, so they never arm the watcher's debounce either way.)
It is floored as a *path* rather than as the bare name `.cache`, because the floor is unconditional
and cannot be whitelisted back — a repo that tracks its own `.cache/` must not lose it.

## Finding the compilation database is this backend's job

clangd searches only an opened file's ancestor directories and a `build/` subdirectory of each, so a
database in `out/` or `cmake-build-debug/` is invisible to it. Measured: with the database in
`build/` clangd works with no flag; in `out/` it emits no progress at all and resolves only to the
header declaration. The backend therefore finds the database and passes `--compile-commands-dir` —
but only when the checkout has exactly one, and that one is usable.

That flag is global. With several databases (an ordinary monorepo), forcing one project's database
onto another project's files gives them wrong include and define flags, which selects a different
preprocessor branch and so a **wrong definition, persisted**. With several, the session points at
none and clangd's own per-file lookup decides, which is correct by construction. A file whose
database clangd cannot find on its own is then **skipped** rather than answered with fallback flags:
a fallback-flags answer resolves a cross-translation-unit call to the callee's header declaration,
which is a wrong verdict rather than a missing one, and the covered-skip budget would never revisit
it.

That refusal is correct and would otherwise be silent — the skips write no rows and defer nothing,
which is exactly what the per-pass log keys on. A pass that issued no requests while skipping
candidates is reported once per session and reset as soon as the backend resolves anything, the same
discipline used for a blocked prerequisite and for a server that never becomes ready. The skipped
paths are also parked, deliberately outside the backlog — a non-empty backlog is what schedules the
next pass, so putting them there would spin the watcher on work every pass can only re-skip — and
they ride along with the next worklist that has real work in it. Without that they are simply
dropped: a pass admits only changed paths plus the backlog, so an operator who consolidates their
databases as the warning asks sees the sources they just made resolvable stay without live evidence
until someone edits them again. A pass that is happening anyway re-answers the question for free,
because an unconfigurable path is skipped before the request budget is touched. A
`compile_commands.json` edit still does not trigger a pass by itself — it contributes no path in
this backend's languages — which is tracked in #996. Separately, the watcher now drops an aborted
session on a typed field rather than on an `"Aborted:"` prefix in operator-facing status text.

The marker search is deliberately more permissive than the warm-up-document search: a build
directory may be hidden (`.build/`), while a warm-up document must never come from a dot-directory.
Excluding a stray vendored database matters more than finding every one — counting it would flip a
working single-database checkout into the multi-database mode and drop the flag that makes it
resolvable. One database reachable through a symlink alias (`out/` alongside `current-build -> out`,
which the search follows on purpose) is counted once, by canonical identity of the marker file, for
the same reason.

Usability is decided by parsing the **whole** database, not its first entry: clangd rejects the
entire file when any single entry is malformed. Verified against clangd 19.1.2 — a database whose
first entry carries `-DPROBE_OK=1` and whose second omits the compiler invocation logs `Missing key:
"command" or "arguments"`, and the define is gone from the flags it falls back to. Trusting the
first entry alone would report such a database usable and pin the session to it. The file is
streamed rather than buffered, so nothing allocates in proportion to its size; the cost is about
0.3s in a release build for a 39 MB / 120k-entry database, held to once per marker file per layout
by the memo below.

Each entry is judged by node SHAPE rather than JSON type, because clangd reads the file with
clang's YAML/JSON reader. Measured, the rule is: every field must be a scalar node, of any kind,
except `arguments`, which must be a sequence. So `"directory": 7`, `"command": null`, and
`"command": true` all load — the node is read as text — while `"command": []` and `"directory": {}`
give `Expected string as value`, and an `"arguments"` that is null, a number, or an object gives
`Expected sequence as value`. Each refusal discards the whole file.

Getting that wrong in either direction is expensive, so the test pins the ACCEPTED shapes as well
as the rejected ones. Too strict and a database the server loads is reported unusable, silently
costing the checkout all live evidence; too loose and a database the server discards looks usable,
which is how a fallback-flags answer becomes trusted evidence. Key presence is tracked separately
from value shape for the same reason: `"arguments": null` is a present field of the wrong shape,
and folding it into "absent" would let the entry pass on its `command` alone.

Both halves of that search run while the maintenance pass holds the repository write lock, so both
are bounded. The walk follows directory symlinks on purpose (a symlinked `build/` is an ordinary
layout), which makes a cycle possible: a depth bound alone caps recursion depth but not the number
of paths, and two links pointing at the same ancestor branch at every level. Only entries whose type
says symlink are canonicalized — that check is answered from the readdir result — and a link whose
target the search already entered is not followed again. A link's target must also stay under the
canonical checkout root: `sdk -> /opt/sdk` or `external -> ..` would otherwise send the walk through
an unrelated tree to the depth limit, and count a `compile_commands.json` found out there as this
checkout's, changing both the pinning decision and the prerequisite verdict. Ordinary subdirectories
cannot escape, so they are not checked and the walk's cost profile is unchanged. Database usability
verdicts are memoized on the layout that produced them, so a worklist no longer re-reads and
re-parses the same file once per file.

The whole walk is skipped outright when the server is not installed: availability is probed
before the layout is resolved, so a checkout with no clangd does not pay for a full traversal under
the write lock on every spawn attempt, and is told about the missing program rather than sent after
a compilation database it cannot use.

## What this generalises

A `tsconfig.json` declares the sources beneath it; a `compile_commands.json` is a build artifact that
need not sit above anything it describes. The project marker introduced for TypeScript now carries
which of those two shapes it has, because conflating them reports an ordinary out-of-tree CMake
project as `Blocked`. Both backends gate on the same question — is there a project here whose load
this server would report? — while each hint names its own marker and its own measured symptom.

## Verification

Measured against clangd 19.1.2 throughout. Beyond the unit and registry tests:

- `real_clangd_resolves_a_call_into_another_translation_unit` (ignored by default, needs `clangd` on
  `PATH`) puts the database in `out/` — where clangd cannot find it unaided — and asserts the call
  resolves to the definition in the other translation unit, not the header declaration. It fails if
  either the background index or the discovered `--compile-commands-dir` regresses; both were
  confirmed by disabling them in turn.
- Position encoding needed no work: clangd counts UTF-16, which is the client's existing fallback
  (with six astral characters before the callee, only the UTF-16 offset resolved; UTF-8 and UTF-32
  both returned `null`).
- The two layout-dependent branches of the pass — the per-file skip and the age-out re-resolve that
  ends a session whose pinned database moved — are covered against real fixture checkouts through a
  seam that supplies a resolved layout and its age.
- The `.cache/clangd` indexing floor is asserted on the real indexing path, over a main checkout and
  a linked worktree sharing one database, against persisted file rows. Its fixtures carry an
  indexable extension rather than clangd's `.idx` artifacts, which no include glob matches and which
  would therefore make the assertion hold with the floor deleted; a sibling `.cache/cmake-build/`
  tree is asserted present so an empty result cannot be explained by dot-directories never being
  walked. Confirmed discriminating by emptying `FLOOR_PATHS` — the rebuild and overlay halves each
  fail independently.
- Workspace `cargo nextest run` (4154 tests) and `cargo test` on the touched crates, all four CI
  clippy legs, and nightly `cargo fmt --check`. The oracle crate's suite is also run with no
  language servers on `PATH`, so a test cannot quietly assert one code path locally and another on
  a machine without the servers installed.


skakri:
@codex review

Issue #935, opened 2026-07-24
sync: enforce roster and role authority for table-sync ingest


skakri:
## Remaining gap

The transport-independent table-sync ingest path (`accept_row_entry`) verifies signatures, stream identity, and chain continuity, but it has no account-roster or device-role admission. There are still no production transport callers, so this is not currently live.

Before the per-scope table-sync wire milestone in #892 ships:

- derive the authenticated peer principal and effective account role before exposing inventory;
- refuse table-row pushes from off-roster and read-only devices before storage;
- enforce the same roster/scope authority at the table-sync ingest boundary so transport is not the only backstop;
- keep unknown tables and scopes forward-compatible without allowing an unauthorized principal to populate retained streams;
- cover direct-ingest and transport tests for read-only, writer/member, owner, removed, and wrong-account devices.

The `/3` content path already has a durable `RoleForbidsAuthoring` fold backstop. Table-sync needs the equivalent authority coupling before its engine becomes reachable from the wire.

## Resolved context

The original `OwnerDemote` concern no longer applies. The account fold retains the immutable enrollment role separately from owner incarnations, rejects promotion of a `ReadOnly` enrollment, and demotion only removes the matching owner incarnation. It therefore cannot turn a read-only enrollment into a writer/member role.

References #892, #930, #406.

Issue #892, opened 2026-07-23
Table→log sync: replicate derived/metadata tables (anchors, overlay, distill) per project


skakri:
## Goal

Replicate rag-rat derived and metadata tables across an account device roster, per project (`repo_id`), on signed streams. A synced memory should carry the metadata needed to surface as a drive-by memory, not only its `/3` content.

Initial data classes:

1. **Anchors** (`repo_memory_bindings`, selected columns): the gating product milestone for drive-by surfacing.
2. **Compacted memories** (`memory_summaries`): costly, regenerable model output.
3. **Verdicts** (`memory_reality`): advisory, regenerable model output.
4. **Distilled papertrail** (`papertrail_distill` and child tables): costly derived records that may contain private tracker evidence.

The `/3` content log remains authoritative for memory content. Table-sync carries row-level derived and metadata state alongside it.

## Architecture

- **Separate signed streams keyed by `(repo_id, scope)`.** A scope binds authority, retention, and flood budgets. Initial scopes are `anchors/1` (durable roster data), `overlay/1` (regenerable roster data), and `distill/1` (regenerable roster-only private evidence).
- **Whole-row Lamport LWW.** One row write clock plus tombstone determines the winner; there is no per-column merge. The producer emits complete after-images and uses a published-row anti-echo hash so remotely applied rows are not re-authored.
- **Forward compatibility belongs in storage and projection, not trust.** Unknown tables/scopes remain retainable and relayable, but cannot project until supported. A projector-version re-fold must reapply retained entries after upgrades.
- **Manifest intersection is routing advice only.** Versioned scope names and a scoped manifest can avoid sending unsupported data, but never grant authority.
- **Principal admission composes by scope.** Roster device, grant device, and anonymous principals are admitted locally according to the scope policy. Roster and role enforcement is tracked in #935.

## Shipped

- [x] **Content to memory drain**: accepted `/3` content becomes local memory rows — #691 / #902.
- [x] **Transport-independent table to log engine**: registry validation, signed row operations, whole-row LWW application, tombstones, producer, and anti-echo handling — #893 / #896.

## Wire-readiness requirements

These must be closed before registering a production table and exposing table-sync over transport:

- [ ] roster and role authority at transport and ingest boundaries — #935;
- [ ] projector versioning plus replay of retained inert or partially understood entries after upgrade;
- [ ] driver ordering: flush and author pending local rows before applying remote entries;
[...]

PR #1000, opened 2026-07-28
feat(sync): re-adopt table-sync rows and deletions orphaned by a removed writer

### LINKED ISSUE #997: sync: re-adopt orphaned table-sync rows whose LWW author is no longer roster-effective
## Context

Table-sync ingest now gates on **current** roster effectiveness (#935): an entry is accepted only if its signing device is a roster-effective writer (`account_roster_history.closed_at IS NULL`, role `member`/`owner`). This is the correct security backstop — a removed or read-only device can never populate a retained stream — but it has one benign-looking cost that is real for one data class.

## The gap

`apply` records an ingested remote row as published, so `produce_and_author` never re-emits it under the local device key. Combined with current-roster ingest gating, a row whose **only** author is a now-removed device:

- persists on replicas that ingested it while the author was effective, but
- never reaches a **newly-enrolled** device (that peer's ingest of the removed author's entry is refused), and
- nothing re-authors it under a current writer, so it does not converge organically.

For **anchors** and **overlay** scopes this is harmless: each device regenerates that state from its own local indexing. For **distill** it is not — distilled records are expensive model output that may have been authored by a single box that later left the roster; that content can then be permanently unavailable to devices enrolled after the removal.

## Proposed fix (no wire/format change)

A current writer re-publishes rows it already holds whose whole-row-LWW owner is no longer roster-effective: re-author the row under the current device key (a new signed entry at a fresh Lamport), so it re-enters the signed log and propagates to every peer, including newly-enrolled ones. This neutralizes the only real cost of the current-roster model and removes any need for a citation-based (epoch-aware) authority model on the table-sync path.

- scope it to rows whose stored author fingerprint is not roster-effective;
- re-author under whole-row LWW so it competes at the current Lamport (no silent overwrite of a newer legitimate row);
- likely a maintenance-pass step alongside `produce_and_author`.

Deferred from #935 (which delivers the ingest gate). References #935, #892.

skakri:
Table-sync ingest accepts an entry only from a device that is currently a roster-effective writer (#935). That gate is the right security backstop, but it strands derived table state whose whole-row last-writer-wins author was later removed from the roster — a newly-enrolled replica drops the removed writer's entries, and because ingested rows are not re-authored by active devices, no current writer ever re-publishes them.

## Two orphan kinds

- **Live rows.** A row whose only author is a since-removed device persists on replicas that ingested it before the removal, but never reaches a device enrolled afterward.
- **Deletions.** A delete authored by a since-removed device lives only in the tombstone table (the applier clears the row clock on delete). A fresh replica refuses that writer's `Remove` but can accept an *older* `Upsert` from a current writer — so the deleted row **resurrects** and the replica diverges from peers that saw the delete.

## Change

At the start of every producer pass, re-adopt both kinds under a current writer's key. Scan the per-row LWW clocks and tombstones; for each whose winning device is no longer a roster-effective writer, re-author the state under the local device at a fresh `next_stream_lamport` — a full-after-image `Upsert` for a live row, a `Remove` for a tombstone.

The re-authored entry sits strictly above the orphan's clock, so it wins last-writer-wins everywhere the state exists and establishes it on a fresh replica. An orphan has no newer legitimate author by definition, so nothing newer is overwritten. A tombstone whose row also has a live clock is skipped, so a re-authored delete never removes a row a concurrent upsert has since resurrected.

Re-adoption is folded into the producer rather than left as a separate step, so it is an intrinsic part of bringing peers up to date and runs wherever the producer is driven. It is a no-op unless the local device is itself a roster-effective writer (a removed device's re-authorship would be refused by every peer), and once state is re-authored locally it is no longer an orphan — so the pass is idempotent and bounded (one entry per orphan, once). No wire or format change.

## Tests

Engine-level coverage for both orphan kinds: a removed writer's live row reaching a fresh replica; a removed writer's deletion re-adopted so a fresh replica does not resurrect the row; selectivity (only the orphan is re-adopted, not a current writer's row); and the guard that a non-writer local device re-adopts nothing. The existing producer tests are unchanged.

Closes #997.

### DIFF
--- a/crates/rag-rat-oplog/src/table_sync/apply.rs
+++ b/crates/rag-rat-oplog/src/table_sync/apply.rs
@@ -483,6 +483,70 @@ fn clear_row_clock(
     Ok(())
 }
 
+/// Every live row's current whole-row-LWW winner for `table`/`repo_id` as `(row_pk, device hex)` —
+/// the live-row half of the re-adoption scan (#997). The `device_fingerprint` is stored as the
+/// lowercase hex the applier wrote (`OpMeta::device.to_string()`); the caller parses it back with
+/// `FromStr`.
+pub(crate) fn row_clock_winners(
+    tx: &Transaction<'_>,
+    repo_id: &str,
+    table: &str,
+) -> anyhow::Result<Vec<(String, String)>> {
+    winners_of("sync_row_clocks", tx, repo_id, table)
+}
+
+/// Every tombstone's current whole-row-LWW winner for `table`/`repo_id` as `(row_pk, device hex)` —
+/// the deletion half of the re-adoption scan (#997). A delete authored by a since-removed writer
+/// lives ONLY here (`apply_remove` cleared the row clock), so scanning row clocks alone would miss
+/// it and a fresh replica could resurrect the row from an older upsert by a current writer.
+pub(crate) fn tombstone_winners(
+    tx: &Transaction<'_>,
+    repo_id: &str,
+    table: &str,
+) -> anyhow::Result<Vec<(String, String)>> {
+    winners_of("sync_row_tombstones", tx, repo_id, table)
+}
+
+fn winners_of(
+    lww_table: &str,
+    tx: &Transaction<'_>,
+    repo_id: &str,
+    table: &str,
+) -> anyhow::Result<Vec<(String, String)>> {
+    // `lww_table` is one of two `&'static` names chosen by the caller, never untrusted input.
+    let mut stmt = tx.prepare(&format!(
+        "SELECT row_pk, device_fingerprint FROM {lww_table}
+         WHERE repo_id = ?1 AND table_name = ?2"
+    ))?;
+    let rows = stmt
+        .query_map(rusqlite::params![repo_id, table], |row| {
+            Ok((row.get::<_, String>(0)?, row.get::<_, String>(1)?))
+        })?
+        .collect::<rusqlite::Result<Vec<_>>>()?;
+    Ok(rows)
+}
+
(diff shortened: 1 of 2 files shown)


skakri:
Closing: this standalone every-pass re-adoption is the wrong seam. It infers an event-history question — which signed state was orphaned by a roster change, and was it authorized when authored — from the merged LWW clock/tombstone tables crossed with the *current* roster, a projection that has already discarded the information the decision needs:

- A locally re-created primary key (a raw insert after an ingested delete) has no `sync_row_clocks` entry, so the tombstone pass re-authors the delete and destroys the local row — and broadcasts a signed delete for it — before the producer can publish the recreation.
- "Winner is now off-roster" cannot distinguish legitimate pre-removal state from a write a replica accepted during the window before it folded the author's removal, so re-adoption can re-sign a revoked write under a current key and spread it to fully up-to-date peers. The row clock carries no authoring-time provenance, so this is undecidable at this seam.
- With tombstones retained indefinitely, a full scan on every producer pass is O(all rows ever deleted), not O(current rows).

Re-adoption is still needed, but it belongs in the sync driver (#892), which owns account-log-first ingest ordering (bounds the acceptance window) and DeviceRemove handling (the correct trigger). The redo will be roster-event/worklist-driven, scope candidates by walking the removed device's `table_sync_entries` chains, treat the physical table as the liveness authority (never delete a pk whose row exists; run after the producer), make the pre-removal-write grandfathering policy explicit and auditable, and file a roster-epoch-on-wire authority model as the deterministic closure for the revoked-write case. Re-scoped in #997 against #892. Reusable pieces (the author+self-apply helper, the op builders, the convergence tests) carry over on the branch.

PR #1046, opened 2026-07-30
fix(sync): tolerate an unreadable row at store open, and purge the table-sync entry log with its repo (#1017, #1004)

### LINKED ISSUE #1004: sync: align rag-rat rm across the table-sync entry log and its stream directory
Follow-up to #1001. Not reachable today — `SYNCABLE_TABLES` is empty, so neither table ever holds a row — but the two halves of the repo purge now disagree, and re-registering a removed repo can resurrect its sync history.

## The mismatch

`purge_repo_rows` sweeps every table carrying `repo_id`. The `table_sync_streams` directory added in #1001 carries one, so it is swept. `table_sync_entries` is keyed by an opaque `stream_id` with no `repo_id`, and the stream-keyed sync substrate is deliberately excluded from repo purge (the same posture as the account and content logs).

So after `rag-rat rm` a repo's signed table history survives without the context needed to place it. The refold skips those entries — correct, and now the documented reason for that branch — but the scope stream id is a pure function of `(repo_id, account_id, scope_id)`, so **re-registering the same repo recreates the identical mapping**, and a later projector bump can replay the removed repo's pending operations back into it.

## Options

1. **Purge the entries too.** Capture the repo's stream ids from the directory before the sweep, then delete `table_sync_entries` for them. This is the "delete by derived scope streams" resolution already sketched for the deferred rm-purge gap; the directory is what makes it possible, since the account id no longer has to be re-derived. Matches what `rm` means to a user, and closes that gap.
2. **Retain both.** Exclude the directory from the class sweep so the entries keep their context, consistent with the retained posture for the rest of the sync substrate. Leaves removed-repo history intact and replayable, which is the weaker answer for a command whose job is de-indexing.

Option 1 is preferred. Either way the two must agree, and the purge tripwire needs the new table seeded so it asserts the chosen behavior rather than silently drifting.

Part of #892. References #767 for the purge inventory and its tripwire.

### LINKED ISSUE #1017: fix(sync): a non-0/1 Bool cell should not make the index unopenable
`read_synced_cells` / `read_all_rows` hard-error when a `Bool` column holds an integer other than 0 or 1. Since the table-sync refold runs at store open, that error now propagates out of `IndexDatabase::open` and re-fails on every subsequent open, with `index --full` taking the same path — so there is no recovery.

Two neighbouring cases in the same replay are already handled without failing the open: undecodable stored bytes are recorded as a terminal quarantine, and a malformed pk arity defers. This one should join them.

The registry lint explicitly cannot require the `CHECK (col IN (0, 1))` that would prevent the state (no pragma exposes it), so the runtime path has to tolerate it.

Not reachable today — no `Bool` column is registered — but it should be fixed before the first one is.

skakri:
Two table-sync defects that are unreachable while `SYNCABLE_TABLES` is empty, and that both get
harder to fix once the first table registers.

## An unreadable row no longer fails the open (#1017)

`read_typed` refuses a `Bool` column holding an integer other than 0 or 1 — coercing it to `true`
would replicate a value that differs from the stored row. It refused by *erroring*, and the
table-sync refold runs at store open, so that error propagated out of `IndexDatabase::open` and
re-failed on every subsequent open, with `index --full` taking the same path. There was no recovery.

The refusal is now a value, and each reader decides what it means:

- `read_synced_cells` reports `SyncedRow::Unreadable` distinctly from `Absent`. Collapsing the two
  would be its own bug: the refold's guard reads an absent row as a local delete awaiting authorship
  and refuses to replay over it, which for a row that is merely unreadable would block that entry
  for good.
- `row_has_unsent_local_change` answers "no unsent change" for a row it cannot read, because it can
  prove nothing about it. The producer cannot author such a row either, and this verdict's two
  readers must not both defer — the row would be unauthorable *and* would permanently block its own
  pending entries. Replaying is the direction with a floor: the entry still has to win on the clock,
  and a winner rewrites every synced column, which is what repairs the cell.
- the producer scan splits the two unreadable cases, because treating them alike deletes data. A row
  whose synced column is unreadable is still live and addressable, so its identity must stay in the
  live set — drop it and the published identity it leaves behind reads as a local delete, and one
  device's malformed cell authors a `Remove` that removes the row from every peer. A row whose *pk*
  is unreadable has no identity to keep alive, and the pk it was published under genuinely no longer
  names a live row.

`Malformed` has exactly one producer, and that follows from the schema rules rather than from
inspection: the registry lint requires a STRICT table, so the `Text`, `I64` and `Blob` mappings are
total, and `Bool` is the only value type whose domain SQLite will not enforce without a
`CHECK (col IN (0, 1))` that no pragma exposes for the lint to require.

## `rag-rat rm` no longer leaves resurrectable sync history (#1004)

`purge_repo_rows` sweeps every table carrying a `repo_id`, which includes the `table_sync_streams`
directory. `table_sync_entries` is keyed by an opaque `stream_id` with no `repo_id`, and the
stream-keyed sync substrate is otherwise deliberately retained across a repo purge — so after
`rag-rat rm` a repo's signed table history survived without the directory row needed to place it.

Retention is only safe where the key cannot be regenerated, and this key can: `scope_stream_id` is a
pure function of `(repo_id, account_id, scope_id)`, so re-registering the same repo recreates the
identical stream mapping and a later projector bump replays the removed repo's parked operations
back into it. The entry log therefore joins the purge, as a transitive child of the directory —
captured up front, because once the class sweep removes the directory row a one-way hash leaves no
way back from an entry to its repo.

The poison-sibling tripwire now seeds a stream and an entry for both the victim and the sibling, so
the guarantee is observable rather than passing on an empty table: the victim's entries must be
gone, and the sibling's — which no `repo_id` predicate could have protected — must survive.

## Verification

Every behavior was mutation-checked: the code under test was reverted and the test confirmed to
fail, and to fail for the stated reason.

| Mutation | Killed by |
|---|---|
| `read_typed` errors on a malformed `Bool` (the prior behavior) | 8 tests, incl. the store-open refold |
| `read_typed` coerces any non-zero integer to `true` | 7 tests |
| the guard defers on an unreadable row | `the_unsent_edit_guard_does_not_defer_on_a_row_it_cannot_read` |
| an unpublishable row is skipped without keeping its identity live | `a_row_that_cannot_be_read_is_neither_authored_nor_removed` |
| an unreadable pk is filled with a placeholder instead of unnaming the row | `a_row_whose_pk_cannot_be_read_releases_the_identity_it_was_published_under` |
| the entry log is retained (the prior behavior) | the poison-sibling tripwire |
| the stream ids are captured after the class sweep | the poison-sibling tripwire |
| the stream capture is unscoped | the sibling-integrity assertion |

Gates: `cargo +nightly fmt --all --check`, clippy on both feature legs (`--no-default-features` and
`--all-features`, `--all-targets`, `-D warnings`), `cargo nextest run --workspace
--no-default-features` (4353 passed), and `cargo test` on the touched crates — the coverage job's
shared-process runner.

Fixes #1017
Fixes #1004

### DIFF
--- a/crates/rag-rat-core/src/index/remove.rs
+++ b/crates/rag-rat-core/src/index/remove.rs
@@ -732,29 +732,10 @@ mod tests {
         let b_stream = seed_table_sync_stream(conn, POISON_REPO_ID, 0xbb);
         let a_stream = seed_table_sync_stream(conn, &repo_a, 0xaa);
 
-        // The multi-checkout shape: a linked worktree registers a SECOND root under the SAME
-        // `repo_id`, which is the only thing about it the store records — the stream directory is
-        // keyed by `(repo_id, account_id, scope_id)` with no checkout dimension, so both checkouts
-        // share one stream row set and a purge from either must take it exactly once. Seeding the
-        // second root pins that: if the purge ever grew a per-checkout predicate, a repo reachable
-        // from two roots would leave half its rows behind.
-        // Plain INSERT, not `INSERT OR IGNORE`: OR IGNORE would swallow a missing NOT NULL column
-        // and leave the scenario un-seeded while the test still read as passing.
-        for root in ["/tmp/zz_poison_main_checkout", "/tmp/zz_poison_linked_worktree"] {
-            conn.execute(
-                "INSERT INTO repo_roots(repo_id, root, registered_at_ms) VALUES (?1, ?2, 0)",
-                params![POISON_REPO_ID, root],
-            )
-            .unwrap();
-        }
-        let b_roots: i64 = conn
-            .query_row(
-                "SELECT COUNT(*) FROM repo_roots WHERE repo_id = ?1",
-                params![POISON_REPO_ID],
-                |row| row.get(0),
-            )
-            .unwrap();
-        assert!(b_roots >= 2, "the victim must be reachable from more than one checkout");
+        // The multi-checkout dimension of this purge is covered where it can be exercised for real:
+        // `schema_bootstrap_tests::worktree_purge` builds a main checkout and a linked worktree
+        // sharing one database, resolves the removal through the linked one, and asserts a sibling
+        // repo's stream-keyed log survives. This tripwire stays about class-level completeness.
 
         // The full class-level table set, captured from the LIVE schema. A subset here would
         // silently narrow the guarantee, so pin a floor on its breadth.
--- a/crates/rag-rat-core/src/index/remove.rs
+++ b/crates/rag-rat-core/src/index/remove.rs
@@ -732,6 +732,30 @@ mod tests {
         let b_stream = seed_table_sync_stream(conn, POISON_REPO_ID, 0xbb);
         let a_stream = seed_table_sync_stream(conn, &repo_a, 0xaa);
 
+        // The multi-checkout shape: a linked worktree registers a SECOND root under the SAME
+        // `repo_id`, which is the only thing about it the store records — the stream directory is
+        // keyed by `(repo_id, account_id, scope_id)` with no checkout dimension, so both checkouts
+        // share one stream row set and a purge from either must take it exactly once. Seeding the
+        // second root pins that: if the purge ever grew a per-checkout predicate, a repo reachable
+        // from two roots would leave half its rows behind.
+        // Plain INSERT, not `INSERT OR IGNORE`: OR IGNORE would swallow a missing NOT NULL column
+        // and leave the scenario un-seeded while the test still read as passing.
+        for root in ["/tmp/zz_poison_main_checkout", "/tmp/zz_poison_linked_worktree"] {
+            conn.execute(
+                "INSERT INTO repo_roots(repo_id, root, registered_at_ms) VALUES (?1, ?2, 0)",
+                params![POISON_REPO_ID, root],
+            )
+            .unwrap();
+        }
+        let b_roots: i64 = conn
+            .query_row(
+                "SELECT COUNT(*) FROM repo_roots WHERE repo_id = ?1",
+                params![POISON_REPO_ID],
+                |row| row.get(0),
+            )
+            .unwrap();
+        assert!(b_roots >= 2, "the victim must be reachable from more than one checkout");
+
         // The full class-level table set, captured from the LIVE schema. A subset here would
         // silently narrow the guarantee, so pin a floor on its breadth.
         let repo_id_tables = repo_scoped_table_names(conn).unwrap();
--- a/crates/rag-rat-core/src/index/remove.rs
+++ b/crates/rag-rat-core/src/index/remove.rs
@@ -555,6 +555,34 @@ mod tests {
         .unwrap();
     }
 
+    /// Seed one table-sync stream directory row for `repo_id` plus one entry on that stream, and
+    /// return the stream id. The entry log carries no `repo_id` and is reached ONLY through the
+    /// directory (#1004), so without this the stream-keyed half of the purge is unobservable — the
+    /// class-level assertion below would pass on an empty table and prove nothing.
+    fn seed_table_sync_stream(conn: &rusqlite::Connection, repo_id: &str, seed: u8) -> Vec<u8> {
+        let stream_id = vec![seed; 32];
+        conn.execute(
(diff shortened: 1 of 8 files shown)


skakri:
Folded the review findings; pushed as a follow-up commit.

**A `Remove` must not replay over an unreadable row.** Correct and the sharpest of the five — the
"a winner rewrites every synced column, which repairs the cell" argument holds for an `Upsert` and
not for a `Remove`, which deletes the row outright. The guard is now asked about the op as well as
the row (`replay_would_destroy_unsent_work`): an upsert still replays, because both readers of this
verdict deferring would leave the row unauthorable *and* permanently blocking its own entries; a
remove defers, because it offers no repair and the surviving row is the recoverable state. Both
edges are pinned — a test fails if a remove is let through, and a different one fails if both kinds
defer.

**Invalid UTF-8 in a `Text` column.** Confirmed by probe: `CAST(X'80' AS TEXT)` stores with
`typeof() = 'text'` and `get::<String>` returns `Err(Utf8Error)`, so my "STRICT makes these mappings
total" comment was simply false and the store-open brick survived for that case. Fixed at the root
rather than by adding a UTF-8 arm: `read_typed` now matches on `ValueRef` and is total over
(declared type, storage class), so a mismatched storage class is carried too. That "should be
unreachable on a STRICT table" reasoning is what produced the original bug, so it is no longer load
bearing anywhere.

**Counting before the directory migration.** Correct, and the guard was wrong as a class, not just
for this table: `count_repo_rows` checked the transitive child's existence but not its parent's,
which was sufficient only while every parent predated its children. `table_sync_entries` (V087)
under `table_sync_streams` (V093) is the first inversion. Both halves are guarded now, via a
`parent_table()` mapping alongside `parent_id_select`, and the regression test reproduces the shape
by dropping the directory — it fails with the exact `no such table: table_sync_streams` when the
guard is reverted.

**The chain epoch.** Correct, and it is the most consequential of the five, but it cannot be settled
here. A local epoch does not work: `repos.removal_generation` is per-device, so two devices with
different removal histories would derive different stream ids for the same scope and never sync at
all. The epoch has to be part of the scope identity peers agree on, which makes it the per-scope
signed stream / manifest milestone's problem — filed as #1049, with the constraint noted where the
purge explains itself. Landing the purge anyway is deliberate: the resurrection path it closes needs
no peer at all, while the fork it exposes needs a transport that does not exist yet. Happy to hold
#1004 instead if that trade reads wrong.

**Linked-worktree coverage.** The tripwire now seeds a second registered root for the victim, which
is the shape a linked worktree leaves in the store. Not a full two-checkout fixture, because the
stream directory is keyed by `(repo_id, account_id, scope_id)` with no checkout dimension — both
checkouts share one row set and the purge is keyed by `repo_id` alone, so there is no per-checkout
path to regress. The second root is what would catch such a predicate if one were ever added, and
the existing test already drives the purge through a bare connection precisely so the scope view
cannot mask it.

Gates re-run: nightly fmt, clippy on both feature legs, `cargo nextest run --workspace` (4358
passed), and `cargo test` on the three touched crates. Each fix is mutation-checked.

skakri:
Fair — the `repo_roots` stand-in could pass without a worktree ever existing, which is exactly the
gap the rule is written against. Replaced with the real fixture in
`schema_bootstrap_tests::worktree_purge`: a main checkout, a linked worktree of the same repo, both
indexed into ONE database alongside an unrelated sibling repo, with the removal resolved through the
LINKED checkout.

It pins the property the purge depends on instead of assuming it — both checkouts register under one
`repo_id`, which is what makes a sweep keyed by `repo_id` alone reach every checkout — and then both
halves the rule asks for: active-checkout scope (the shared repo's stream directory and the
stream-keyed entry log it places are both gone, driven from the linked worktree) and sibling
preservation (the other repo's survive). The sibling half is the load-bearing one, because
`table_sync_entries` carries no `repo_id`, so nothing but the captured stream ids could have spared
it.

Mutation-checked independently of the poison-sibling tripwire — retaining the entry log and
capturing an unscoped stream set are each caught by this test on its own — and the two-root stand-in
is dropped so the tripwire stays about class-level completeness and the checkout dimension lives in
one place.

Gates re-run: nightly fmt, clippy on both feature legs, `cargo nextest run --workspace` (4359
passed), and `cargo test` on the touched crates.

The remaining open item is #1049 (a purged stream needs an epoch before a re-registration can be
safe against peers) — not fixable here, since the epoch has to be part of the scope identity peers
agree on.

Issue #1058, opened 2026-07-30
sync: retain and promote gapped table-sync chain entries instead of dropping them


skakri:
Part of #892 (wire-readiness requirement 4).

## The gap

`store::accept_row_entry` classifies a verified entry against its `(stream_id, device_fingerprint)`
chain. A `ChainFit::Gap` — the entry links to a predecessor this device does not hold — returns
`AcceptOutcome::MissingPredecessor` and the entry is **discarded**.

Out-of-order delivery is the normal condition on a transport, not an edge case. A chain delivered
tail-first therefore never converges: every entry after the gap is dropped, and only redelivery in
exact causal order recovers it. Today this is unreachable (`SYNCABLE_TABLES` is empty and there is
no transport), but it must be closed before table-sync is exposed over the wire, and it is
storage-shaped — retrofitting it after the wire protocol is designed is strictly more expensive.

## Approach

Retain a gapped entry in its own table, keyed by entry hash, and promote it when its predecessor
arrives.

- **A separate table, not a status column.** Six queries read `table_sync_entries` as "the accepted
  chain" — the authoring Lamport clock, the lamport-advance bound, `chain_tail`, `entry_exists`,
  `winning_entry_op`, and the refold's pending set. A separate table gives all six the right answer
  by construction.
- **Gapped entries do not advance the Lamport clock.** Otherwise one far-ahead retained entry drags
  the local authoring clock, which is what the advance bound exists to prevent. This is the axis
  distinction from *pending* entries, which are stored and do count.
- **Promotion re-runs the full accept + apply path**, so a promoted entry cannot skip a gate a fresh
  entry passes — including the unsent-local-work deferral (#1005) and quarantine recording.
- **A fork revealed by promotion is resolved.** When two retained entries share a predecessor, the
  promoted one takes the successor slot and the sibling is then provably a conflict; the promote
  loop drains siblings before advancing so the loser cannot be stranded.
- **Bounded, not reclaimed.** A per-chain cap evicts the furthest-ahead entry; no TTL ships here.
  Reclamation belongs with retention/compaction, together with #1020's constraint that a row's
  publication record must be refreshed before its winner is dropped.

Only a roster-effective writer can occupy the table at all — the authority gate runs before chain
classification — so the flood surface is roster members, whose remedy is device removal.

Fork *evidence* (proving equivocation to a peer) stays with the transport milestone.

References #892, #1005, #1004, #1020.

Issue #1020, opened 2026-07-29
sync: expose unresolved row-winner causes across replay and local production


skakri:
## Remaining gap

`StaleRow::Unknown` combines missing clocks, absent or undecodable winner entries, wrong operation/table/key identities and unprojectable values. Replay already persists the coarse `DeferredUnresolvedWinner` reason, but the underlying cause is not distinguished. Scan-only unreadable local rows also need an observable explanation without requiring an incoming entry.

Successful re-authoring refreshes publication state, so the original claim that every such row repeats winner lookup forever is not universal. The remaining requirement is useful, bounded diagnostics with defined lifetime and repair behavior.

## Scope

- Define stable unresolved-winner subcauses and cover incoming replay, producer/self-apply and scan-only states.
- Decide whether durable causes use stable pending tokens or a separate field; retain compatibility with existing coarse pending reasons.
- Make causes queryable without warning on every pass. Ensure diagnostic writes survive the relevant transaction/rollback path and clear on actual repair.
- Keep merge ordering, unsent-edit protection and retention semantics unchanged.

## Retention constraint

#1291 superseded the old suggestion to refresh publication hashes and then drop a live winner. Live winners must remain deliverable to fresh peers; current retention pins or carries them. Missing winners from older compacted stores are a legacy-repair case, not permission to weaken current retention.

## Acceptance

Each unresolved cause; persistence across restart; producer success and `Superseded` rollback; unreadable local row without incoming work; scope-moved identity mismatch; repaired state clearing; and no accidental deletion or loss of unsent edits. A slice exposing only the existing coarse reason does not complete durable fine-grained/scan-only diagnostics.

No retention redesign or tombstone GC is required. Use the current typed pending-reason and row-clock seams.

skakri:
Not addressed by #1053 (which is #1005 only). Recording what tracing this settled, because three designs were attempted and each failed differently — the constraints are worth having before a fourth.

**The producer cannot record it.** `produce_row_ops`, `author_row_entry` and the self-apply share one transaction, so a reason written there is either cleared by `record_published` in the same transaction or rolled back by the `Superseded` bail. There is no committed state to inspect afterwards.

[...]

PR #1120, opened 2026-08-02
feat(sync): re-adopt rows orphaned by a removed writer

### LINKED ISSUE #997: sync: re-adopt orphaned table-sync rows whose LWW author is no longer roster-effective

skakri:
Closes #997.

## Summary
- record each effective roster removal as durable stream-scoped re-adoption work during the authority projection rewrite
- drain that work from the producer transaction by walking only the removed writer’s accepted table entries
- re-author surviving whole-row state under the current writer, with physical-table liveness and audit provenance

## Verification
- cargo +nightly fmt --all -- --check
- cargo clippy -p rag-rat-oplog -p rag-rat-db --all-targets -- -D warnings
- cargo test -p rag-rat-oplog readoption --lib
- cargo test -p rag-rat-oplog --lib
- cargo test -p rag-rat-db --lib

### DIFF
--- a/crates/rag-rat-core/src/index/remove.rs
+++ b/crates/rag-rat-core/src/index/remove.rs
@@ -590,6 +590,31 @@ mod tests {
             ]],
         )
         .unwrap();
+        // The re-adoption tables are swept through the same directory (#997): a pending repair and
+        // its provenance both name the stream, so each needs a seed row for the orphan check to
+        // observe.
+        conn.execute(
+            "INSERT INTO table_sync_readoption_work(account_id, device_fingerprint, stream_id, \
+             roster_ref, removed_at_epoch, enqueued_at_ms) VALUES (?1, ?2, ?3, ?4, 1, 0)",
+            params![vec![seed; 32], vec![seed; 32], stream_id, vec![seed ^ 0x11; 32]],
+        )
+        .unwrap();
+        conn.execute(
+            "INSERT INTO table_sync_readoption_audit(account_id, removed_fingerprint, \
+             adopter_fingerprint, stream_id, repo_id, scope_id, table_name, row_pk, \
+             original_lamport, original_entry_hash, adopted_entry_hash, adopted_at_ms) VALUES \
+             (?1, ?2, ?3, ?4, ?5, 'demo/1', 't', 'aa', 1, ?6, ?7, 0)",
+            params![
+                vec![seed; 32],
+                vec![seed; 32],
+                vec![seed ^ 0x22; 32],
+                stream_id,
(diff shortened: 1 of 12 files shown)


skakri:
Alignment with #892 and its comments:

- The Jul 28 redesign is the shipped shape: durable worklist on effective DeviceRemove, candidates walked from the removed device's table_sync_entries chains, physical table as liveness authority. The merged-state scan from PR #1000 is not reused.
- Lag-window case: writes accepted before the removal fold are grandfathered (re-adopted) with audit provenance; the roster-epoch-on-wire authority model remains the deterministic closure and stays the follow-up. No wire change here.
- Retention: anchors/1 is fully retained, so #1020's refresh-before-winner-drop constraint and the #1058 gapped-table TTL do not bind yet — this PR drops no winners and does not prune gapped entries.
- Only repo-scoped tables (repo_column as text PK member) participate; repo_column: None remains unsupported, per the account-global caveat.

Issue #1122, opened 2026-08-02
Late repository merge strands live memory verification rows under the retired id


Nitjsefnie:
## Problem

The late LocalOnly-to-Portable merge moves memories to an already-registered portable repository but leaves their live verification siblings under the retiring ID:

- `memory_reality`
- `memory_note_summaries`
- `memory_model_failures`

In-place adoption already handles these tables. The late-merge disposition does not; the registry's `LATE_MERGE_MEMORY_VERIFICATION_UNRESOLVED` coverage set records the gap explicitly. No foreign-key cascade or per-active-repository cleanup can recover rows after their repository ID is retired. The result is invisible model output and permanent orphan rows, requiring regeneration for the moved memories.

This is established by source inspection; no fresh runtime reproduction is claimed here.

## Updated scope

#1320 replaced the live `memory_summaries` table with `memory_note_summaries`. The retired table's abandonment is deliberate retained-entry compatibility behavior and must not be casually included in a cleanup. Some earlier documentation inconsistencies have been corrected; the live-row disposition remains unresolved.

Choose one explicit policy:

- discard the three live siblings as derived data, preserving target rows and regenerating as needed; or
- move them with the memories, with explicit target-collision, content-hash and prompt-version rules.

Discard-as-derived matches the existing derived-periphery treatment and is the smaller proposal; the issue does not assume that policy has already been implemented or approved as a preservation requirement.

## Acceptance

- Atomic disposition in the late-merge transaction, with accurate registry coverage and revision/invalidation behavior.
- Actual shallow-clone deepening into an occupied portable ID; all three live tables and conflicting target rows.
- In-place adoption unchanged and retired-summary compatibility preserved.
- Rollback/retry and correct sync publication/stream-retirement bookkeeping.
- An explicit policy for already-stranded rows, not only future merges; repair only where retirement/orphanhood can be established.
- Main/linked checkouts sharing a DB and unrelated sibling rows preserved.

Use the current migration ladder and registry layout; schema refactoring alone does not settle this issue. Independent of #1295 tombstone-row GC.

Issue #706, opened 2026-07-16
split rag-rat-core into a layered workspace (base/db/index + papertrail, llm, clones, oracle, oplog, query, dream, watch)


skakri:
rag-rat-core has grown to ~200k lines across ~15 top-level modules with clean measured seams;
a workspace split buys per-crate incremental builds (today the 36k-line integration-test
module recompiles on any core touch), real CI path-gating, and lets new features (the
distillation milestone) land in their own crate from birth.

Measured module dependency structure (import counts at module level) shows:

- **Nearly standalone**: `oplog` (24k lines → schema+query only), `index/papertrail` (13.6k),
  `index/clones` (12.9k), `index/oracle` (13.4k), `index/ai` (12.9k) — each depends only
  downward on schema/config/index-core.
- **Foundation**: `index/schema` + `config` + `locks`/`repo_identity`/`embedding_models` are
  imported by everything → become the base crates.
- **The only cycles**: `query ↔ dream` (query/query_api import dream for review surfaces and
  verdict hydration; dream imports query for memory reads) and a smaller `query ↔ oplog`.
  Breaking these (trait seam or surface relocation) is the only non-mechanical work in the
  whole split.

Target DAG: base → db(schema) → index; satellites papertrail / clones / oracle / llm(ai) /
oplog in parallel; then query(+query_api+search), dream (post-inversion), watch; mcp/cli
unchanged as consumers. Integration tests move to a dev-only tests crate.

Sequencing is conflict-aware, not dependency-pure:

- [x] 1. foundation cut: base + db (mechanical, one PR, full call-site migration — no
      re-export shims per repo policy)
- [x] 2. papertrail crate (cleanest satellite; unblocks the distillation substrate so #702+
      land in the new crate directly)
- [x] 3. llm crate (index/ai: providers + cookbook + embed loop — needed by dream and #704)
- [x] 4. clones crate
- [x] 5. oracle crate
- [x] 6. query crate + break the query↔dream and query↔oplog cycles (the design work)
- [x] 7. dream crate
- [x] 8. oplog crate — deliberately LAST: the account/oplog phase-C stream (#662-665) is in
      flight; extracting under it invites merge hell
- [ ] 9. dissolve the core remainder (watch/maintenance) + integration-tests crate

Mechanics to settle in the first PR: release configuration for the new crates (all must be
published since the published cli/mcp depend on them; recommend a single workspace version),
[...]

PR #1123, opened 2026-08-02
test(schema): fail when a repo_id-scoped table declares no adoption or merge disposition

### LINKED ISSUE #571: test(schema): tripwire enumerating repo_id-scoped tables against the adoption coverage lists
## Problem

`DIRECT_SCOPED_ADOPTION_TABLES` and `LATE_MERGE_DERIVED_PERIPHERY_TABLES` (`crates/rag-rat-core/src/index/schema/registry.rs`) are **hand-audited** to cover every `repo_id`-carrying table — the coverage note dates the last full audit to V044 and instructs that any new `repo_id`-scoped table be added to one of the lists (or the authored-move set). Nothing enforces this automatically.

V056 (`git_change_couplings`) shipped omitted from both lists and the full suite stayed green. The omission would have stranded coupling rows behind a moved `repo_id` on a LocalOnly→Portable in-place upgrade (the `repo_meta` stamp moves, the rows do not → `ensure_coupling_fresh` sees a matching stamp and returns an empty section) and left invisible garbage on the late-merge path. It was caught in review, not by a test.

## Proposal

Add a schema test that turns the hand-audit into a tripwire: enumerate every table carrying a `repo_id` column (via `PRAGMA table_info` over the freshly-bootstrapped schema) and assert each is in exactly one of:

- `DIRECT_SCOPED_ADOPTION_TABLES`,
- `LATE_MERGE_DERIVED_PERIPHERY_TABLES`, or
- an explicit, documented allowlist of authored-move / intentionally-excluded tables (the memory tables, etc.).

A new `repo_id`-scoped table then fails this test until its adoption disposition is declared.

## Acceptance

- The test fails if a `repo_id`-scoped table is absent from all three sets.
- The allowlist is documented inline with the reason each entry is excluded.

Nitjsefnie:
## Summary

Adds a schema test that enumerates every `repo_id`-scoped table from the live schema and fails if one is not declared in a disposition list, so the next omission is caught mechanically rather than by eye — the way the V056 `git_change_couplings` gap was.

## Related Issues and Pull Requests

Fixes #571

## Changes

- **`crates/rag-rat-db/src/schema/registry.rs`** — one appended `#[cfg(test)] mod repo_id_scope_coverage`, plus a corrected doc comment on `LATE_MERGE_DERIVED_PERIPHERY_TABLES`. **No production code path changes**; the only edit outside the test module is comment text.
  - That comment described the list as the A5 periphery list "MINUS the three memory tables". The real delta is **7**: four are authored-and-moved (`repo_memories`, `repo_memory_bindings`, `repo_memory_fts`, `repo_node_edges`, each explicitly `UPDATE`d to the target id in `merge_local_incumbent_into_registered`), and the other three are the dream-v2 verification siblings (`memory_reality`, `memory_summaries`, `memory_model_failures`), which that path neither moves nor deletes. The comment now names all seven with their disposition.
  - The `40` floor no longer appears twice: `MIN_REPO_SCOPED_TABLES` is interpolated into both the comparison and its failure message, so the two cannot drift apart.

Three points where the implementation departs from the issue text, each for a reason in the code:

- **The issue's path is stale.** It names `crates/rag-rat-core/src/index/schema/registry.rs`; the schema layer moved to `crates/rag-rat-db/` in `6ddd64e` (#706 phase 1). The consts are also bare `const`, not `pub`, so the test has to live in-crate — which matches the existing pattern (`schema/mod.rs` already carries two inline test modules).
- **It asserts per re-point path, not once over a union.** There are three lists (`DIRECT_SCOPED_ADOPTION_TABLES` 26, `A5_PERIPHERY_DIRECT_SCOPED_TABLES` 16, `LATE_MERGE_DERIVED_PERIPHERY_TABLES` 9) and two independent mechanisms — in-place adoption and the late-upgrade merge. A single union assertion absorbs the whole A5 list without distinguishing the paths, which would let a table that is re-pointed on upgrade but dropped on merge pass silently. That is exactly the shape of #1122 below.
- **The enumeration reuses production code rather than restating a list.** It calls `schema::repo_scoped_table_names` — the same `pragma_table_info`-based sweep the purge's class-level backstop uses. A second hand-written roster would be forgotten in both places on the same commit and keep passing; deriving it from the schema means a new migration that adds a `repo_id` column changes the test's input set with no test edit.

## Testing

`cargo test -p rag-rat-db --no-default-features repo_id_scope_coverage` → `3 passed; 44 filtered out`. (Named via substring, not `--exact` — a bare name with `--exact` runs zero tests here while still printing `test result: ok`.)

**Proof the assertions bite**, each reverted afterwards:

- Removing `git_change_couplings` from `DIRECT_SCOPED_ADOPTION_TABLES` — the issue's own case — fails 2 of 3, naming `["git_change_couplings"]`.
- Misspelling `reconcile_attempts` as `reconcile_attempt` in `A5_PERIPHERY_DIRECT_SCOPED_TABLES` fails 2 of 3. This is the silent class: both re-point loops guard on table/column presence, so a renamed or misspelled entry is a no-op no coverage assertion would otherwise see.

**Proof they cannot pass vacuously.** Dropping the `apply()` call so the bootstrap enumerates nothing initially left two of the three green — "every enumerated table is declared" is vacuously true over an empty set. The shared helper now asserts a floor of 40 on the sweep (live count: **50** of 125 total tables), mirroring the existing purge tripwire at `remove.rs:761-766`. With that in place the same mutation fails all three tests at the floor rather than one test incidentally.

Gates: `cargo +nightly fmt --check` 0; `cargo clippy --workspace --all-targets --no-default-features --locked -- -D warnings` 0; `cargo nextest run --workspace --no-default-features --locked --profile ci` 0 (4896 passed, 7 skipped); `cargo nextest run -p rag-rat-db --no-default-features --locked` 0 (47 passed).

**Not run, and not to be read as passing:** `clippy (all features)` and the default-feature `build`, because the default feature set pulls ONNX which fails to link on this machine; and `coverage`, which exceeded the local time budget mid-compile and was killed — CI covers all three.

## Bugs Discovered

- #1122 Memory verification rows are stranded under the retiring repo id on a late-upgrade merge

## Follow-ups / Known Limitations

- **Two dispositions are parked rather than settled**, in sets named `LATE_MERGE_MEMORY_VERIFICATION_UNRESOLVED` and `TABLE_SYNC_UNRESOLVED`, each with an inline note. Parking keeps the test green on today's `main` without pretending the question is answered; both shrink to nothing once you rule.
  - The memory verification tables are #1122 above.
  - `table_sync_streams`, `sync_published_rows`, `sync_row_clocks` and `sync_row_tombstones` carry `repo_id`, are reached by no foreign key, and are re-pointed by neither path. I did **not** file that as a bug, because their `repo_id` sits beside a `stream_id` derived by one-way sha256 from `(repo_id, account_id, incarnation_ref, scope_id)` — re-pointing `repo_id` alone would desynchronise the key from its own derivation, so "not re-pointed" may well be correct by design. It is an undeclared disposition, and the answer is yours.
- The V044 audit date in the surrounding doc is many migrations stale — 50 of 125 tables are now repo-scoped, which is the surface this test is meant to hold.

## Footer

Generated by Claude Opus 5 (brief, implementation, review, verification), Claude Sonnet 5 (implementation)

### DIFF
--- a/crates/rag-rat-db/src/schema/registry.rs
+++ b/crates/rag-rat-db/src/schema/registry.rs
@@ -1386,3 +1386,252 @@ pub fn is_root_already_indexed_conn(
     }
     Ok(false)
 }
+
+/// TRIPWIRE (#571): the adoption coverage lists in this module are HAND-AUDITED against every
+/// `repo_id`-carrying table, and the audit note dates the last full pass to V044. V056
+/// (`git_change_couplings`) shipped absent from both lists with the full suite green: no test
+/// could see the omission. These tests turn the audit into a gate.
+///
+/// The enumeration comes from the FRESHLY BOOTSTRAPPED SCHEMA ITSELF —
+/// [`super::repo_scoped_table_names`], the same `sqlite_master` × `PRAGMA table_info` sweep the
+/// purge runs on — never from a roster restated here. A restated roster would pin nothing: the
+/// defect this guards against IS an omission from a hand-written list, and a second hand-written
+/// list reproduces it exactly (the new table would be forgotten in both places at once).
+///
+/// A new `repo_id`-scoped table therefore fails these tests until its disposition is declared on
+/// BOTH re-point paths — the in-place LocalOnly→Portable adoption in [`register_repo`] and the
+/// late-merge retirement in [`merge_local_incumbent_into_registered`].
+#[cfg(test)]
+mod repo_id_scope_coverage {
+    use std::collections::BTreeMap;
+
+    use super::*;
+
+    /// `(table, why)` — a `repo_id` table whose adoption disposition is settled somewhere OTHER
+    /// than the two loops: an explicit statement in [`register_repo`], or an FK that carries it.
+    const ADOPTION_HANDLED_ELSEWHERE: &[(&str, &str)] = &[
+        (
+            "repos",
+            "the registry table itself — `repo_id` is its primary key, not a scope column; \
+             adoption INSERTs the new row and DELETEs the source one",
+        ),
+        (
+            "repo_meta",
+            "explicit UPDATE in the adoption transaction, every key except the shallow boundary \
+             (deliberately left behind so the source row's cascade drops it)",
+        ),
+        (
+            "repo_roots",
+            "explicit UPDATE, ordered BEFORE the source `repos` row is deleted — its FK would \
+             otherwise cascade the roots away",
+        ),
+        (
+            "git_file_changes",
+            "deliberately NOT updated: its `(repo_id, commit_hash)` FK to `git_commits` is `ON \
+             UPDATE CASCADE`, so re-pointing `git_commits` carries it, and an explicit UPDATE \
+             would trip the FK (child before parent)",
+        ),
+    ];
+
+    /// `(table, why)` — a `repo_id` table whose disposition is settled somewhere OTHER than the
+    /// two DELETE loops of [`merge_local_incumbent_into_registered`]: an AUTHORED move onto the
+    /// target id, or an FK that drops it with a deleted parent.
+    const LATE_MERGE_HANDLED_ELSEWHERE: &[(&str, &str)] = &[
+        ("repos", "the retiring owner's row is DELETEd outright"),
+        (
+            "repo_meta",
+            "falls with the owner's `repos` row via `ON DELETE CASCADE`, the now-meaningless \
+             shallow boundary included",
+        ),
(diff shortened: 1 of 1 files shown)


skakri:
Let's fold in the 2 and 3 in this PR as it already touches this logic: stale comment and 40 twice.

Nitjsefnie:
Both folded in shortly — the doc comment rewritten to the real delta, and the floor collapsed to one named constant so the comparison and the message can't drift apart.

PR #1170, opened 2026-08-14
fix(dispatch): read only the leading identifier when testing for PascalCase

### LINKED ISSUE #1124: is_pascal_case treats a SCREAMING_CONST with a method chain as PascalCase, so classify_call returns Wrapper
## Description

`is_pascal_case` cannot distinguish a `SCREAMING_CONST` followed by a method chain from a genuine `PascalCase` variant, so `classify_call` returns `Wrapper` for a call whose head is a constant.

Concrete case, at `crates/rag-rat-mcp/src/server/service.rs:43`:

```rust
crate::tools::TOOL_NAMES.iter().map(|name| { … })
```

The head segment here is `TOOL_NAMES.iter().map(…)`. `is_pascal_case` sees a leading `T` plus the lowercase characters contributed by the appended method names and reports PascalCase, so the call classifies as `Wrapper` — "a transparent variant constructor, trace through it" — which it is not. `TOOL_NAMES` is a constant and the call is a plain delegation.

The misreading is in the case predicate, not in path splitting: a segment that begins with an uppercase letter and later contains lowercase satisfies the current test regardless of whether the uppercase run is a `PascalCase` word or a `SCREAMING_SNAKE` constant with a method chain glued onto it.

## Expected Behavior

`is_pascal_case` distinguishes `PascalCase` from `SCREAMING_SNAKE_CASE`, so a constant head does not classify as a variant constructor. `crate::tools::TOOL_NAMES.iter().map(…)` should classify as `Delegate`.

## Reproduction Steps

1. On `main`, classify the call expression at `crates/rag-rat-mcp/src/server/service.rs:43`.
2. Observe the head segment `TOOL_NAMES.iter().map(…)` is treated as PascalCase and the call classifies as `Wrapper`.

## Environment / Context

Present on `main` independently of any path-splitting behaviour — the predicate misreads the segment whichever splitter produced it.

## Discovered During

Sampling classification changes while working #1094. Twenty role changes were examined by hand; nineteen traced to the path splitter, and this one did not — it is the case predicate, and it is unaffected by that work.

## Suggested Fix

Unverified: reject a segment containing an underscore, or one whose leading uppercase run is longer than one character before any lowercase appears, before accepting it as PascalCase. Either would separate `TOOL_NAMES…` from `Wrapper`, but which boundary you want depends on whether single-word screaming constants (`NOW`, `SPEC`) should also be excluded, and that is your call.

Nitjsefnie:
## Summary

`classify_call` was returning `Wrapper` for a dispatch arm whose tail is a `SCREAMING_CONST` head with a method chain, so the `dispatch_handle` fact for that arm was dropped rather than emitted. `is_pascal_case` now weighs only a segment's leading identifier, which classifies the arm as `Delegate`.

## Related Issues and Pull Requests

Fixes #1124

## Changes

- `crates/rag-rat-core/src/index/languages/rust/dispatch.rs`: `is_pascal_case` reads only the leading identifier of a path segment — everything up to the first character that cannot continue an identifier — leaving the existing "starts uppercase, carries a lowercase" test untouched. The head scan follows Rust's `XID_Continue` rule rather than `char::is_alphanumeric`, so a decomposed identifier such as `A\u{301}ccent` keeps its combining mark instead of being truncated at `A`. Without that, `A\u{301}ccent(handle())` classifies as `Delegate` and emits the constructor itself as the dispatch handler.
- `crates/rag-rat-core/src/index/mod.rs`: `GRAPH_INDEX_VERSION` 16 → 17. `ensure_graph_index_current_inner` re-extracts only rows below the current version, so without the bump an existing database keeps its stale classification and missing `dispatch_handle` edge until the source file changes or the index is rebuilt by hand.
- Unit tests beside the existing dispatch tests covering the predicate, `classify_call`, and `enum_variant_key` — the predicate's other production consumer.
- `crates/rag-rat-core/src/index/schema_bootstrap_tests/graph_heal_robustness.rs`: coverage over persisted `edges_data` rows, including a forced v16 database that re-extracts to v17, and a shared database with a base checkout plus a linked worktree.

The boundary is deliberately narrower than either option the issue floated, and it is your call to overrule. Rejecting any segment containing `_` would leave `NOW.elapsed()` misread and would newly break `Wrapped(payload).to_string()`. Rejecting a leading uppercase run longer than one character would demote `HTTPResponse` from constructor to handler. Reading only the head fixes `TOOL_NAMES.iter().map(…)` and `NOW.elapsed()` while leaving `HTTPResponse` and `Wrapped(payload).to_string()` exactly where they are. The underscore boundary is a one-line change if you prefer it.

`XID_Continue` and `char::is_alphanumeric` are incomparable rather than nested, so the head scan is not simply a narrowing. `XID_Continue` accepts characters `is_alphanumeric` rejects (U+0301 COMBINING ACUTE ACCENT, U+203F UNDERTIE) and rejects characters it accepts (U+00B2 SUPERSCRIPT TWO, which is not valid in a Rust identifier). Both directions are pinned by tests, along with a character in both sets (U+05B0 HEBREW POINT SHEVA) as an unchanged anchor.

## Testing

`cargo test -p rag-rat-core` for the dispatch, `enum_variant_key` and graph-heal tests, plus fork CI under this project's `--no-default-features --locked` configuration: rustfmt, clippy and test all green.

The predicate tests fail on assertions — not compile errors — when the head scan is reverted to `char::is_alphanumeric`, across every boundary class above. The graph-heal test fails when `GRAPH_INDEX_VERSION` is returned to 16, so the upgrade path is covered rather than assumed. The real multi-line site at `crates/rag-rat-mcp/src/server/service.rs:43` classifies `Delegate` with the change and `Wrapper` without it.

## Bugs Discovered

- #1171 cargo-deny fails on main since 2026-08-12 on RUSTSEC-2026-0253 in lru 0.18.1

## Follow-ups / Known Limitations

Existing v16 databases re-extract on open and advance eligible file rows to version 17. In a database shared by several checkouts, only rows whose indexed bytes match the active checkout heal on that open; a sibling worktree row whose bytes differ stays at 16 until the index is opened from a checkout containing those bytes.

## Dependencies

Adds `unicode-ident` as a direct production dependency for Rust `XID_Continue` classification, at the version already present in the workspace lockfile.

## Footer

Generated by Claude Opus 5 (implementation, review, testing), GPT-5.6 Luna (implementation), GPT-5.6 Sol (review)
Generated by Kimi K3 (implementation, testing)

### DIFF
--- a/crates/rag-rat-core/src/index/languages/rust/dispatch.rs
+++ b/crates/rag-rat-core/src/index/languages/rust/dispatch.rs
@@ -879,12 +879,55 @@ mod classify_call_tests {
     /// traces through the constant instead of recording the call as the handler.
     #[test]
     fn a_screaming_const_with_a_method_chain_is_not_a_wrapper() {
+        // A chain on a scoped constant carrying a closure transformation keeps its delegation:
+        // the closure is behavior, not a produced value, and the trailing `map` is a std adapter
+        // method that never resolves to an in-corpus symbol, so no edge is synthesized (#1124).
         assert!(matches!(
             role_of("crate::tools::TOOL_NAMES.iter().map(|name| describe(name))"),
             CallRole::Delegate
         ));
         // A single-word constant has no underscore to give it away and reads the same way.
         assert!(matches!(role_of("NOW.elapsed()"), CallRole::Delegate));
+    }
+
+    /// A method whose receiver is ANOTHER CALL'S RESULT is a chained adapter tail
+    /// (`LIMIT.min(cap).max(..)`, `list.len().saturating_sub(..)`): it adapts a value, it is
+    /// never the handler, and it is not a transparent wrapper either — its argument is adapter
+    /// input, not the response. Only a BARE callee (a free function, a path, or a method on a
+    /// plain binding/`self` receiver) may fall through to `Delegate`; otherwise the trailing
+    /// method name is recorded as the handler and bare-name fallback binds it to an unrelated
+    /// same-named repository symbol — a false persisted edge (#1124 maintainer feedback).
+    #[test]
+    fn a_chained_adapter_tail_is_neither_a_delegate_nor_a_wrapper() {
+        assert!(matches!(role_of("LIMIT.min(cap).max(1)"), CallRole::Skip));
+        assert!(matches!(role_of("BACKENDS.len().saturating_sub(1)"), CallRole::Skip));
+        assert!(matches!(role_of("items(count).len().max(1)"), CallRole::Skip));
+        assert!(matches!(role_of("make_worker().run(input)"), CallRole::Skip));
+        // Bare callees keep their delegation: a free function, and a method on a plain
+        // binding/`self`/field receiver (`worker.run` IS the handler — #208 review round 11).
+        assert!(matches!(role_of("run(input)"), CallRole::Delegate));
+        assert!(matches!(role_of("worker.run(input)"), CallRole::Delegate));
+        assert!(matches!(role_of("self.embed(input)"), CallRole::Delegate));
+        assert!(matches!(role_of("self.worker.run(input)"), CallRole::Delegate));
+    }
+
+    /// An associated-constant method chain invoked with PLAIN DATA is the dispatch itself
+    /// (`Handler::DEFAULT.run(input)` — the constant is the chained method's receiver). A chain
+    /// that WRAPS a produced value is a builder (`Resp::DEFAULT.with_body(handle(cap))`): the
+    /// nested call is the real handler, so the builder is a transparent wrapper of its payload
+    /// and the builder method itself is never recorded (#1124 maintainer feedback).
+    #[test]
+    fn an_associated_constant_builder_wraps_a_produced_payload() {
+        assert!(matches!(role_of("Resp::DEFAULT.with_body(handle(cap))"), CallRole::Wrapper));
+        assert!(matches!(
+            role_of("<Resp as Build>::DEFAULT.with_body(handle(cap))"),
+            CallRole::Wrapper
+        ));
+        assert!(matches!(role_of("Handler::DEFAULT.run(input)"), CallRole::Delegate));
+        assert!(matches!(role_of("<Handler as Runner>::DEFAULT.run(input)"), CallRole::Delegate));
+        // An intermediate builder call inside the chain still delegates when the final call is
+        // invoked with plain data.
+        assert!(matches!(role_of("Handler::DEFAULT.build().run(input)"), CallRole::Delegate));
     }
 
     /// A method chain glued onto an ASSOCIATED CONSTANT delegates to the chained method (#1124
--- a/crates/rag-rat-core/src/index/languages/rust/dispatch.rs
+++ b/crates/rag-rat-core/src/index/languages/rust/dispatch.rs
@@ -932,42 +932,50 @@ mod classify_call_tests {
         assert!(matches!(role_of("<Resp as Default>::default()"), CallRole::Skip));
     }
 
-    /// A `SCREAMING_CONST` head carrying a method chain is a plain delegation, not a transparent
-    /// variant constructor (#1124). The chain's method names supply the lowercase the case test
-    /// looks for, so reading the whole segment bills `TOOL_NAMES.iter().map(..)` as a wrapper and
-    /// traces through the constant instead of recording the call as the handler.
+    /// A `SCREAMING_CONST` head carrying a method chain is not a transparent variant constructor
+    /// (#1124). The chain's method names supply the lowercase the case test looks for, so reading
+    /// the whole segment would bill `TOOL_NAMES.iter().map(..)` as a wrapper and trace through the
+    /// constant instead of classifying the call on its own shape.
     #[test]
     fn a_screaming_const_with_a_method_chain_is_not_a_wrapper() {
-        // A chain on a scoped constant carrying a closure transformation keeps its delegation:
-        // the closure is behavior, not a produced value, and the trailing `map` is a std adapter
-        // method that never resolves to an in-corpus symbol, so no edge is synthesized (#1124).
+        // A chain on a SCOPED associated constant delegates to its chained method: `TOOL_NAMES` is
+        // the receiver named by its owning module, so the call is the dispatch itself (#1124).
         assert!(matches!(
             role_of("crate::tools::TOOL_NAMES.iter().map(|name| describe(name))"),
             CallRole::Delegate
         ));
-        // A single-word constant has no underscore to give it away and reads the same way.
-        assert!(matches!(role_of("NOW.elapsed()"), CallRole::Delegate));
+        // A BARE constant receiver names no owner, so the trailing method is an adapter over the
+        // constant's value rather than a handler — neither a wrapper nor a delegate.
+        assert!(matches!(role_of("NOW.elapsed()"), CallRole::Skip));
     }
 
-    /// A method whose receiver is ANOTHER CALL'S RESULT is a chained adapter tail
-    /// (`LIMIT.min(cap).max(..)`, `list.len().saturating_sub(..)`): it adapts a value, it is
-    /// never the handler, and it is not a transparent wrapper either — its argument is adapter
-    /// input, not the response. Only a BARE callee (a free function, a path, or a method on a
-    /// plain binding/`self` receiver) may fall through to `Delegate`; otherwise the trailing
-    /// method name is recorded as the handler and bare-name fallback binds it to an unrelated
-    /// same-named repository symbol — a false persisted edge (#1124 maintainer feedback).
+    /// A method glued onto ANOTHER CALL'S RESULT (`LIMIT.min(cap).max(..)`,
+    /// `list.len().saturating_sub(..)`) or onto a BARE SCREAMING constant (`BASE.to_string()`) is
+    /// an adapter tail: it adapts a value, it is never the handler, and it is not a transparent
+    /// wrapper either — its argument is adapter input, not the response. Only a BARE callee (a
+    /// free function, a path, or a method on a plain binding/`self` receiver) may fall through to
+    /// `Delegate`; otherwise the trailing method name is recorded as the handler and bare-name
+    /// fallback binds it to an unrelated same-named repository symbol — a false persisted edge
+    /// (#1124 maintainer feedback).
     #[test]
-    fn a_chained_adapter_tail_is_neither_a_delegate_nor_a_wrapper() {
+    fn an_adapter_tail_is_neither_a_delegate_nor_a_wrapper() {
         assert!(matches!(role_of("LIMIT.min(cap).max(1)"), CallRole::Skip));
         assert!(matches!(role_of("BACKENDS.len().saturating_sub(1)"), CallRole::Skip));
         assert!(matches!(role_of("items(count).len().max(1)"), CallRole::Skip));
         assert!(matches!(role_of("make_worker().run(input)"), CallRole::Skip));
+        // A single method on a bare SCREAMING constant is the same adapter shape without the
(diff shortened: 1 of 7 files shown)


skakri:
@codex review unicode identifiers and linked-worktrees (may affect indexing edges)

Nitjsefnie:
Both angles you named turned out to matter, so this pushes a fix for each.

The unicode one was a real regression: `char::is_alphanumeric` truncates a decomposed identifier at its combining mark, so `Áccent(handle())` classified as `Delegate` and emitted the constructor itself as the dispatch handler. The head scan now follows `XID_Continue`. Worth noting the two predicates are incomparable rather than nested — `XID_Continue` also rejects things `is_alphanumeric` accepts, like `²` — so both directions are pinned by tests, and the PR body's earlier claim that the predicate was "strictly narrower" was wrong and is corrected.

The worktree one exposed that the extractor change never reached an existing database at all: `ensure_graph_index_current_inner` only re-extracts rows below `GRAPH_INDEX_VERSION`, which was still 16. That is now 17, with the upgrade path and a shared-database-plus-linked-worktree case covered over persisted `edges_data` rows.

Happy to split the version bump into its own commit if you would rather review it separately.

crates/rag-rat-core/src/index/languages/rust/dispatch.rs
```diff
@@ -5,14 +5,25 @@
...
+///
+/// Only that leading identifier is weighed because a segment can carry a method chain glued onto
+/// its head (`TOOL_NAMES.iter().map`, `NOW.elapsed`). The appended method names supply the very
+/// lowercase the test looks for, so reading the whole segment takes a constant for a variant
+/// constructor and bills the call as a transparent wrapper instead of the delegation it is (#1124).
 fn is_pascal_case(name: &str) -> bool {
-    name.chars().next().is_some_and(char::is_uppercase) && name.chars().any(char::is_lowercase)
+    let mut head =
+        name.chars().take_while(|&character| is_xid_continue(character) || character == '_');
+    head.next().is_some_and(char::is_uppercase) && head.any(char::is_lowercase)
```
skakri: There's a remaining classifier bug: even a fresh index misclassifies `Handler::DEFAULT.run(input)` because the PascalCase Handler receiver fallback returns `Skip`. Fix it narrowly; don’t treat it as a duplicate of the graph-version issue.

In classify_call, recognize a SCREAMING_CONST method-chain tail before the PascalCase receiver fallback.

e.g.
```rust

  if is_screaming_const_method_chain(tail) {
      return CallRole::Delegate;
  }

  match segments.as_slice() {
      [.., receiver, _] if is_pascal_case(receiver) => CallRole::Skip,
      _ => CallRole::Delegate,
  }
```

crates/rag-rat-core/src/index/languages/rust/dispatch.rs
skakri: `<Handler as Runner>::DEFAULT.run(input)` hits the early raw.starts_with('<') return. Adding the predicate only before the PascalCase receiver fallback will not fix this. Detect constant method chains before both qualifier fallbacks and add unit plus end-to-end coverage.

crates/rag-rat-core/src/index/schema_bootstrap_tests/graph_heal_robustness.rs
skakri: stamps graph_version = 17, now equal to `GRAPH_INDEX_VERSION`. Use current version + 1: currently 18, and assert it remains unchanged.

skakri (changes requested):

crates/rag-rat-core/src/index/languages/rust/dispatch.rs
```diff
@@ -9,10 +9,18 @@ use tree_sitter::Node;
...
+///
+/// Only that leading identifier is weighed because a segment can carry a method chain glued onto
+/// its head (`TOOL_NAMES.iter().map`, `NOW.elapsed`). The appended method names supply the very
+/// lowercase the test looks for, so reading the whole segment takes a constant for a variant
+/// constructor and bills the call as a transparent wrapper instead of the delegation it is (#1124).
 fn is_pascal_case(name: &str) -> bool {
-    name.chars().next().is_some_and(char::is_uppercase) && name.chars().any(char::is_lowercase)
+    let mut head =
+        name.chars().take_while(|&character| character.is_alphanumeric() || character == '_');
+    head.next().is_some_and(char::is_uppercase) && head.any(char::is_lowercase)
```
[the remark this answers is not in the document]
Nitjsefnie: Addressed: the graph index version is now 17, with a v16 upgrade regression that removes the corrected fact and verifies re-extraction. The exact pushed head passes the focused upgrade coverage.

crates/rag-rat-core/src/index/languages/rust/dispatch.rs
```diff
@@ -768,4 +776,33 @@ mod classify_call_tests {
         assert!(matches!(role_of("render_page(body)"), CallRole::Delegate));
         assert!(matches!(role_of("<Resp as Default>::default()"), CallRole::Skip));
     }
+
+    /// A `SCREAMING_CONST` head carrying a method chain is a plain delegation, not a transparent
+    /// variant constructor (#1124). The chain's method names supply the lowercase the case test
+    /// looks for, so reading the whole segment bills `TOOL_NAMES.iter().map(..)` as a wrapper and
+    /// traces through the constant instead of recording the call as the handler.
+    #[test]
+    fn a_screaming_const_with_a_method_chain_is_not_a_wrapper() {
```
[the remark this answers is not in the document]
Nitjsefnie: Added persisted-edge coverage for active and sibling linked checkouts. It verifies the active checkout is corrected while the sibling rows, resolutions, and version remain isolated until that checkout is opened.

crates/rag-rat-core/src/index/languages/rust/dispatch.rs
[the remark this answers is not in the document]
Nitjsefnie: Updated leading-identifier handling to use Rust XID continuation semantics, including decomposed and connector characters. The exact pushed head passes the Unicode boundary coverage.

crates/rag-rat-core/src/index/languages/rust/dispatch.rs
Nitjsefnie: Fixed narrowly with structural SCREAMING_CONST method-chain recognition before the PascalCase receiver fallback. Unit and persisted-dispatch coverage pass on the exact pushed head.

crates/rag-rat-core/src/index/languages/rust/dispatch.rs
Nitjsefnie: Moved structural constant-chain detection before both qualifier fallbacks and covered plain, UFCS, nested, and interrupted chains end to end. The exact pushed head passes those cases.

crates/rag-rat-core/src/index/schema_bootstrap_tests/graph_heal_robustness.rs
Nitjsefnie: Updated the fixture to stamp current graph version plus one and assert the future graph version, scope version, targets, and resolutions remain unchanged. The exact pushed head passes that regression.

Nitjsefnie:
Rebased and addressed all six review threads. The two remaining red Supply chain jobs are RUSTSEC-2026-0258 for h2 0.4.15; I reproduced the identical cargo-audit and cargo-deny failures on current main at e9c5a41794208fe52c1cbb78b5b2a567855097a9, and this PR does not alter h2 or the security policy.

skakri:
Two real regressions, both reproduced:

HIGH: rust/dispatch.rs:23: false dispatches edges from std adapter chains.
CONST.min(cap).max(1) now fails is_pascal_case, and with one segment the Skip guard can't fire — so it falls through to Delegate and records the last chained method name as the handler. Bare-name target_name_fallback then binds it to an unrelated repo symbol.
- PR: dispatch_handle(to=max) + synthesized edge to src/lib.rs::max. main: nothing.
- This shape exists today: rag-rat-llm/src/throughput_tune.rs:189, rag-rat-cli/src/init/wizard/steps/embedding.rs:856, rag-rat-core/src/index/consolidate.rs:957. The GRAPH_INDEX_VERSION 17 re-extract will materialize them on every existing index.
- Fix: gate the fall-through on the tail being a bare callee — a chained tail (contains ./() should be Skip, not Delegate.

MEDIUM: rust/dispatch.rs:66: is_associated_const_method_chain loses the real handler.
Resp::DEFAULT.with_body(handle(cap)) was Wrapper (argument traced → handle recorded); it's now unconditionally Delegate → records with_body, never descends. The predicate can't tell Handler::DEFAULT.run(input) from a builder wrapper, and resolves the ambiguity against the module's own false-edge contract.

LOW:
- schema_bootstrap_tests/dispatch.rs:1462 — the new test asserts fact rows only (no to_symbol_id, no dispatches edge), so neither regression is guarded. Needs an end-to-end edge assertion plus a negative fixture with a same-named in-repo helper.
- rust/dispatch.rs:25,40,53 — || character == '_' is dead. _ is Pc, already in XID_Continue — the same fact the PR's own A\u{203F}bc test relies on. It contradicts the doc comment above it.
- PR description calls unicode-ident test-only; it's a production dep on the extraction hot path.

skakri:
Supply chain issues resolved on main, rebase when convenient.

Nitjsefnie:
There is one structural ambiguity in the requested classifier fix: `Resp::DEFAULT.with_body(handle(cap))` and `Handler::DEFAULT.run(normalize(input))` are identical to the extractor modulo identifier spelling. Should this nested-call shape be **Skip** (emit nothing, preserving the no-false-edge contract but losing `handle`) or **Wrapper** (keep `handle`, but accept a false producer edge when the chained method is the real handler)?

Nitjsefnie:
Moved to draft so this cannot land while the two regressions you reproduced are still open — the std adapter-chain false edge at `dispatch.rs:23` and the handler loss in `is_associated_const_method_chain`. I will flip it back to ready once both are fixed and the tests assert the edges end to end rather than the fact rows alone.

skakri:
> There is one structural ambiguity in the requested classifier fix: `Resp::DEFAULT.with_body(handle(cap))` and `Handler::DEFAULT.run(normalize(input))` are identical to the extractor modulo identifier spelling. Should this nested-call shape be **Skip** (emit nothing, preserving the no-false-edge contract but losing `handle`) or **Wrapper** (keep `handle`, but accept a false producer edge when the chained method is the real handler)?

I think neither. The dilemma is manufactured by the payload-shape heuristic itself. Make an associated-const method chain `Delegate` unconditionally, and delete call_carries_nested_call / subtree_has_nested_call.

skakri:
`is_associated_const_method_chain` runs before the `is_chained_adapter_tail` guard and accepts a `crate::…::SCREAMING_CONST` root, so the qualified spelling bypasses the `fix: LIMIT.min(cap).max(1)` -> Skip, but `crate::config::LIMIT.min(cap).max(1)` -> Delegate (records max). The e2e regression test only covers the unqualified form.

skakri (changes requested):
See comments

Nitjsefnie:
Fixed at `dea6baa`. The ordering was the bug: `is_associated_const_method_chain` claimed the expression before the chained-adapter guard could Skip it, so the guard now runs first and the e2e regression covers bare, `crate::`-qualified, `self::`/`super::`-relative and aliased roots rather than only the unqualified spelling. Our own graph-heal fixtures turned out to be a second instance of the same gap and are rekeyed. Full crate suite 2088/2088, clippy and nightly fmt clean.

One thing that needs your call rather than ours: `GRAPH_INDEX_VERSION` is already `17` on `main` from #1239, and the "rows owed" predicates gate on strict `<`, so an index a `main`-built binary stamped `17` before this merges will not re-extract and keeps the stale classification. Released users are unaffected — v0.23.0 ships `16` and re-extracts once either way. Bumping to `18` here costs one extra re-extract for unreleased-`main` indexes and a number change in three fixture assertions; leaving it at `17` costs nothing but helps nobody. Want me to bump it in this PR?

PR #1262, opened 2026-08-26
fix(test): stop tests reserving a port they cannot hold

### LINKED ISSUE #1251: Tests that free an ephemeral port and then require it back race any concurrent binder
Several tests bind `127.0.0.1:0`, record the kernel-assigned port, drop the listener, and then require that *exact* port to behave a particular way. Between the drop and the reuse the port is free for the whole machine — including sibling tests in the same binary doing the identical dance, since ephemeral ports come from one shared range. Under parallel test load the port gets taken and the test fails on an unrelated assertion.

## Sites

Two opposite intents, one mechanism:

**"Give the port back to me"** — fails if anything else takes it first:

- `crates/rag-rat-mcp/src/lens_server.rs:1341` — `port_scan_falls_back_after_a_busy_candidate` frees `available_port`, then asserts `bind_first_free` selects it.
- `crates/rag-rat-mcp/src/lens_server.rs:1396` — frees a probe port so `serve_standalone` binds that one; the comment notes the chosen address "is not otherwise observable from here".
- `crates/rag-rat-mcp/src/lens_server.rs:1465` — same pattern.

**"The port must refuse connections"** — fails in the other direction if something binds it:

- `crates/rag-rat-core/src/index/ai/reconcile/embed_loop.rs:1718` — `drop(listener); // the port now refuses connections`. If the port is reused, the connection succeeds and the test no longer exercises the unreachable-endpoint path. It would pass or hang rather than fail, so this one degrades silently.

## Evidence

`port_scan_falls_back_after_a_busy_candidate` fails **16 times in 40 runs** of the `rag_rat_mcp` lib test binary standalone under `cargo test`. It passes in isolation and under `cargo nextest run`, which gives each test its own process — so CI's nextest job is green while the coverage job's plain `cargo test` is not. `http::tests::dropping_a_request_cancels_its_database_work` flakes in the same runs.

## Directions

- Where the test only needs *a* port that is not the busy one, assert that rather than a specific number.
- Where the code under test picks its own port, pass `0` and read the chosen port back from what it publishes — the `serve_standalone` tests already read the discovery file for other assertions.
- Where the test needs a port that refuses, the reserve-then-free trick cannot guarantee it; the closed-endpoint case needs a different construction.

Related: #1240 (same runner split, different mechanism), #1242 (wall-clock assertions under load).

skakri:
Five tests bound `127.0.0.1:0`, recorded the kernel-assigned port, dropped the listener, and then required that *exact* port to behave a particular way. Between the drop and the reuse the port is free for the whole machine — including sibling tests in the same binary doing the identical dance, since ephemeral ports come from one shared range.

## Reproduced, then fixed

Running the `rag-rat-mcp` lib binary 20× under plain `cargo test`:

| | failures | where |
|---|---|---|
| before | **1 / 20** | `lens_server.rs:1328`, the equality assertion in `port_scan_falls_back_after_a_busy_candidate` |
| after | **0 / 20** | — |

(The issue measured 16/40 on a busier machine. The rate is load-dependent; the failure line is the same.)

These are invisible to the CI gate. `.config/nextest.toml` sets `retries = 2` and `ci.yml` runs nextest, which also gives each test its own process. The **coverage** job runs plain `cargo test` — one process per binary, threads within, no retries — which is the only place this shows up.

## Each site now asserts the property it means

| site | was | now |
|---|---|---|
| `port_scan_falls_back_after_a_busy_candidate` | "returns *this* port" | "skips the busy candidate" — the second candidate is `0`, which nothing can take first |
| `active_task_publishes_and_cleans_up_on_abort` | `discovery.port == port` | binds `0`; connects to the published port, which establishes it is the bound one without naming a number |
| `standalone_serving_publishes_discovery_only_for_a_loopback_bind` | reserved port | binds `0`, reads the address back |
| `standalone_serving_publishes_the_advertise_url_when_given` | reserved port | binds `0` — it asserts only on the advertised URL and never needed a port |
| `light_pass_with_unreachable_query_endpoint_defers` | reserve-then-free "closed" port | dials port 9, as its own sibling twelve lines below already does |

The last one is the one worth reading twice. It needs a port that **refuses**, and reserve-then-free cannot deliver that in either direction — a reused port *answers*, so the test would exercise the reachable path while still passing. It degraded silently rather than failing.

## One production change

`serve_standalone` now reports the address it bound, through an optional channel:

```rust
pub bound_address: Option<tokio::sync::oneshot::Sender<SocketAddr>>,
```

This closes a gap the code already admitted — the deleted test comment read "its chosen address is not otherwise observable from here". A caller that asks for port 0 had no way to learn what it got, and naming a concrete port instead means picking one nothing else has taken, which is not something a caller can know: a port is only yours while you hold it. The CLI passes `None`.

## Verification

```
cargo nextest run --workspace                                                         → 5232 passed
cargo test -p rag-rat-mcp --lib, 20 runs                                              → 0 failures (1 before)
cargo clippy --workspace --all-targets --no-default-features --locked -- -D warnings  → 0
cargo clippy --workspace --all-targets --all-features --locked -- -D warnings         → 0
cargo +nightly fmt --check                                                            → 0
```

Fixes #1251.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5268,6 +5268,7 @@ name = "rag-rat-oracle"
 version = "0.23.0"
 dependencies = [
  "anyhow",
+ "libc",
  "protobuf",
  "rag-rat-base",
  "rag-rat-clones",
--- a/crates/rag-rat-cli/src/main.rs
+++ b/crates/rag-rat-cli/src/main.rs
@@ -496,6 +496,7 @@ fn serve_http(config: Config, args: &ServeArgs) -> anyhow::Result<()> {
                 auth_token: token,
                 allowed_origins,
                 advertise_url,
+                bound_address: None,
             },
             election_lock,
             shutdown_signal(),
--- a/crates/rag-rat-core/src/index/ai/reconcile/embed_loop.rs
+++ b/crates/rag-rat-core/src/index/ai/reconcile/embed_loop.rs
@@ -1713,13 +1713,12 @@ mod freshness_version_tests {
         // probe embed's connect is refused, so defer (SkipEphemeral), NOT embed-and-fail into
         // `Failed` chunk_embeddings, and without paying an O(repo) candidate scan first.
         let conn = schema_conn();
-        let closed = {
-            let listener = TcpListener::bind("127.0.0.1:0").unwrap();
-            let port = listener.local_addr().unwrap().port();
-            drop(listener); // the port now refuses connections
-            format!("http://127.0.0.1:{port}")
-        };
-        activate_ephemeral_with_query_endpoint(&conn, Some(&closed));
+        // Port 9 (discard) is never served on a dev box or a CI runner, so the connect is refused
+        // by construction. Binding an ephemeral port and dropping it does NOT give a closed port:
+        // it is free for the whole machine from that moment, and a sibling test that takes it
+        // turns this into a test of the reachable path — silently, since it would then pass or
+        // hang rather than fail. The sibling below already dials this address for the same reason.
+        activate_ephemeral_with_query_endpoint(&conn, Some("http://127.0.0.1:9"));
         seed_embedding_chunk(&conn, 1);
         assert!(matches!(ephemeral_light_acquire(&conn), ChunkEmbedder::SkipEphemeral));
     }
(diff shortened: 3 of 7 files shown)


skakri:
Duplicate of #1261, which was opened first and covers the same five sites. #1261's approach is better where they differ: it proves the server reached serving by signalling from inside the shutdown future, rather than adding a bound-address reporting channel to `serve_standalone` as this branch did — same guarantee, no production API change.

Issue #1372, opened 2026-09-17
oplog/sync: load-bearing guards whose removal no test would notice


skakri:
## Behavior

Across the op-log and sync crates, a number of guards can be deleted or inverted with the relevant suite still green against a green baseline. Each was verified by applying the mutation, confirming it compiled, and running the suite; build-break mutants are excluded rather than counted either way.

Separate issues cover the four with a concrete failure scenario. This one covers the remainder, which share one cause: the suite pins each rule where it **refuses**, never where it would **over-grant**.

### `rag-rat-oplog` — control v2 (`crates/rag-rat-oplog/src/account/`)

| site | mutation that survives |
|---|---|
| `fold/v2.rs:333` | absent-verdict default flipped to `Authorized` |
| `fold/v2.rs:297` | any authorized entry may mint an incarnation, not only `is_mint()` |
| `fold/v2.rs:434` | an unknown demote owner no longer parks |
| `fold/v2.rs:435` | the `OwnerDemote` subject/`owner_id` binding removed |
| `fold/v2.rs:447` | the watermark-misbound reject removed |
| `fold/v2.rs:516` | cut ordering reversed |
| `fold/v2.rs:571` | tombstone applied unconditionally |
| `control_v2/executor.rs:174` | `RejectCause::Condemned` rerouted to `Parked(Signer)` |
| `control_v2/executor.rs:202` | `Precondition(reason)` rerouted to `Inadmissible` |
| `control_v2/executor.rs:204` | `ParkCause::CutTarget` rerouted to `Signer` |
| `control_v2/views.rs:187` | consumer-in-evidence check removed |
| `control_v2/views.rs:208` | `i64::try_from(seq)` removed |
| `storage.rs:2985-2995` | the `!grew` short-circuit and headroom `ensure!` in the reservation top-up, both neutered together |

The module header at `fold/v2.rs:33-47` argues each register precondition as load-bearing; the I2 last-owner guard is pinned, the `OwnerDemote` binding and watermark binding are not.

### `rag-rat-oplog` — fold and store

[...]

PR #1385, opened 2026-09-18
test(oplog): pin every conjunct of the v2 candidate admission check

skakri:
Rows of #1372 plus a control, and an instance of a trap this issue now records: **a multi-conjunct `require` shares one error message**, so a test asserting that message pins whichever conjunct its fixture violates and leaves the rest free.

`decode_candidate` (`crates/rag-rat-oplog/src/account/control_v2/views.rs`) refuses with one message across five conjuncts. A previous PR covered `seq`; the rest were still deletable with the whole crate suite green — as was the payload's checkpoint check just below.

## The two sharp ones

**`account_id == pin.account_id`.** Evidence signed for a *different account* verifies fine on its own terms, so nothing downstream re-checks it. Admitting one would draw a foreign account's history into this account's replay plan.

**`op.checkpoint == pin.checkpoint_digest`.** A header can be entirely conformant while the operation inside commits to a different checkpoint — and since this same decode is the membership test for the execution pool, admitting one would pull an operation bound to another checkpoint into this account's refold.

## The sealed guard has nothing behind it

Worth stating precisely, because the obvious reading is wrong. The guard refuses on the **declared** suite, before the payload is read — and that is the whole of the defense. With the `crypto_suite == 0` conjunct disabled the plan *resolves* on a conformant plaintext payload, so `ops::decode` is not a backstop. Genuinely sealed bytes reaching it would be refused only by the accident of ciphertext failing to parse as CBOR. The test comment says this, because a reader who assumes a second line of defense can delete the check.

## The duplicate-candidate refusal

`checkpoint_mismatch_and_duplicate_evidence_are_rejected` already reaches this guard, but each of its four assertions is a bare `Invalid(_)` — red if the guard is **deleted**, green if it is **replaced** by any other refusal. Asserting the refusal's own message distinguishes it from every sibling reporting through the same variant.

## The control test is the interesting part

`plan_replay` decodes the **consumer** through `decode_candidate` *before* any evidence, with the same refusal message. So if the shared template ever stopped being admissible, the refusal tests would keep passing — asserting a message the consumer produced, never examining their evidence.

`the_shared_template_is_itself_admissible` closes that, and supplies the template as *evidence* rather than resolving a bare consumer, so it also covers the evidence-side work no refusal reaches: distinctness against the consumer, the duplicate insert, the citation record. Ordering is **not** reachable from it — the root view is empty, so citations are written and never read; that path stays with the diamond test whose root names its candidates.

It is proved to control rather than assumed: sabotaging the helper's consumer turns it red while `evidence_signed_for_another_account_is_refused` stays green — exactly the drift the control exists to catch.

## A fixture trap, recorded

A non-zero `crypto_suite` **requires** a `key_id`; `sign_account_entry` refuses the header outright without one. Setting the suite alone yields a candidate that never reaches the planner, so the test passes without exercising the guard.

## Naming

The `signed` helper's fourth parameter is renamed `salt` → `distinguisher`. Nothing derives a key from it — it varies the operation's reason string so two otherwise identical fixtures hash differently. The name was the whole of its cryptographic claim.

## Verification

Each guard neutered individually, against a green 1185-test baseline:

| mutation | kills |
|---|---|
| `account_id == pin.account_id` → `true` | `evidence_signed_for_another_account_is_refused` |
| `log_id == 0` → `true` | `evidence_on_another_log_is_refused` |
| `op_version == CONTROL_VERSION` → `true` | `evidence_at_another_op_version_is_refused` |
| `crypto_suite == 0` → `true` | `sealed_evidence_is_refused` |
| delete the `i64::try_from` conjunct | `a_candidate_whose_seq_exceeds_sqlite_integer_is_refused` |
| `op.checkpoint == pin.checkpoint_digest` → `true` | `a_candidate_naming_another_checkpoint_is_refused` |
| drop the `duplicate v2 candidate` refusal | `the_same_candidate_supplied_twice_is_refused`, and the bundled test that also reaches it |

Each conjunct kills exactly its own test and no other. The surrounding `&&` was never flipped to `||` — that disables every sibling at once, so a kill would prove a test noticed *something* rather than that it pins one guard.

`cargo clippy -p rag-rat-oplog --all-targets -- -D warnings`: clean. No production code changes.

## Also on the issue

`views.rs:180` (`MAX_REFERENCES`) is **struck as unreachable** — exceeding it needs ~16.78 MB of entry hashes, past the 16 MiB byte limit that refuses first. `MAX_VIEWS` and `MAX_ENTRIES` remain open and are both reachable and cheap.

`checkpoint_mismatch_and_duplicate_evidence_are_rejected` is itself four wildcard `Invalid(_)` assertions covering the duplicate-view and view-checkpoint-mismatch guards — more surface for the same issue, left for a follow-up rather than grown into this PR.


skakri:
CodeQL flagged five critical `rust/hard-coded-cryptographic-value` alerts on this branch. Diagnosed and fixed; the fix is unverified until the gate re-runs, so here is what it is and why.

## Cause

All five flagged lines are the same call: `signed(&checkpoint, &device, root, 7)`. The rule is keying on the helper's fourth parameter, which was named **`salt`**.

It is not a salt. `signed` uses it exactly once:

```rust
*reason = format!("revoked {salt}");
```

It varies the operation's reason string so two otherwise identical fixtures hash differently. Nothing derives a key from it, and no cryptographic operation reads it. The name was the whole of its cryptographic claim, and a scanner reads that claim literally.

## Fix

Renamed to `distinguisher`, with a doc comment saying what it does and why the name matters. This should also clear the two pre-existing alerts on `main` at the `signed(..., 255)` call sites, which are the same pattern.

No suppression: this repo has no CodeQL config and no suppression precedent, and the rename is a genuine improvement — `salt` described nothing about the value's role.

## Two things I got wrong on the way

I predicted the cause was the three 32-byte literals this PR adds (`[0xaf; 32]`, `[0x11; 32]`, `[0x5c; 32]`), since `hard-coded-cryptographic-value` was already flagged twice in this file. **None of them appears at any flagged line.** Reading the alerts settled what guessing would not have.

I also first read the branch's `0 open alerts` against main's `10` as reassuring. It was not — no analysis had been uploaded for that ref yet, so the zero meant *no data*, not *no findings*.

## Caveat

I cannot run CodeQL locally, so the rename clearing the rule is a prediction, not a result. If the gate still fails after this push, the rule is matching on something other than the parameter name and I will re-diagnose rather than reach for a suppression.

Suite 1184 passed, clippy clean, no production code changes.

